PR #10, opened 2025-01-20
feat: Truncate input schema, limit description to 200 characters

jirispilka:
- Handle server readiness probe
- Add Origin/mcp-server to user-agent
- Truncate inputSchema description to 200 characters. Limit enums to 50 values.

### DIFF
--- a/src/actorDefinition.ts
+++ b/src/actorDefinition.ts
@@ -1,9 +1,9 @@
 import { Ajv } from 'ajv';
 import { ApifyClient } from 'apify-client';
 
-import { MAX_DESCRIPTION_LENGTH, MAX_ENUM_LENGTH } from './const.js';
+import { MAX_DESCRIPTION_LENGTH, MAX_ENUM_LENGTH, MAX_MEMORY_MBYTES } from './const.js';
 import { log } from './logger.js';
-import type { ActorDefinitionWithDesc, Tool, SchemaProperties } from './types.js';
+import type { ActorDefinitionWithDesc, SchemaProperties, Tool } from './types.js';
 
 /**
  * Get actor input schema by actor name.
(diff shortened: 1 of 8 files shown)


src/actorDefinition.ts
```diff
@@ -52,6 +53,38 @@ async function fetchActorDefinition(actorFullName: string): Promise<ActorDefinit
...
+    for (const property of Object.values(properties)) {
+        if (property.description.length > MAX_DESCRIPTION_LENGTH) {
+            property.description = `${property.description.slice(0, MAX_DESCRIPTION_LENGTH)}...`;
+        }
+        if (property.enum) {
+            property.enum = property.enum.slice(0, MAX_ENUM_LENGTH);
+        }
+        if (property.enumTitles) {
+            property.enumTitles = property.enumTitles.slice(0, MAX_ENUM_LENGTH);
+        }
```
MQ37, line 71: does including `enumTitles` make sense? They are only for the Console UI, right?

src/actorDefinition.ts
```diff
@@ -52,6 +53,38 @@ async function fetchActorDefinition(actorFullName: string): Promise<ActorDefinit
...
+}
+
+/**
+ * Filters schema properties to include only the necessary fields.
+ * @param properties
+ */
+function filterSchemaProperties(properties: { [key: string]: SchemaProperties }): { [key: string]: SchemaProperties } {
+    const filteredProperties: { [key: string]: SchemaProperties } = {};
+    for (const [key, property] of Object.entries(properties)) {
+        const { title, description, enum: enumValues, enumTitles, type, default: defaultValue, prefill } = property;
```
MQ37, line 83: same here, is  `enumTitles` really needed?

src/types.ts
```diff
@@ -18,3 +18,13 @@ export interface Tool {
     inputSchema: object;
     ajvValidate: ValidateFunction;
 }
+
+export interface SchemaProperties {
+    title: string;
+    description: string;
+    enum: string[]; // Array of string options for the enum
+    enumTitles: string[]; // Array of string titles for the enum
```
MQ37, line 28: same here

src/actorDefinition.ts
jirispilka: I kept it on purpose for these types of fields
```    enum: [
      '',
      'two',
      'twoAndHalf',
    ],
    enumTitles: [
      'All',
      '⭐️ 2+ star',
      '⭐️ 2.5+ stars',
    ],

```
Without the enumTitles you don't know, what empty string `''` is referring to. But I don't have any strong preference. I think it can work without `enumTitles` too.

I'll keep it for now. Let us see later.

PR #37, opened 2025-03-11
fix: Rename tool name, sent `notifications/tools/list_changed`

jirispilka:
Closes https://github.com/apify/tester-mcp-client/issues/11 https://github.com/apify/tester-mcp-client/issues/12


src/examples/clientSse.ts
```diff
@@ -23,7 +23,7 @@ dotenv.config({ path: path.resolve(dirname, '../../.env') });
 
 const SERVER_URL = 'https://actors-mcp-server.apify.actor/sse';
 // We need to change forward slash / to underscore -- in the tool name as Anthropic does not allow forward slashes in the tool name
-const SELECTED_TOOL = 'apify--rag-web-browser';
+const SELECTED_TOOL = 'apify-slash-rag-web-browser';
```
MQ37: maybe use the `actorNameToToolName('apify/rag-web-browser')`?

src/examples/clientStdio.ts
```diff
@@ -24,7 +24,7 @@ const SERVER_PATH = path.resolve(dirname, '../../dist/index.js');
 const NODE_PATH = execSync(process.platform === 'win32' ? 'where node' : 'which node').toString().trim();
 
 const TOOLS = 'apify/rag-web-browser,lukaskrivka/google-maps-with-contact-details';
-const SELECTED_TOOL = 'apify--rag-web-browser'; // We need to change / to _ in the tool name
+const SELECTED_TOOL = 'apify-slash-rag-web-browser';
```
MQ37: same here

src/actors.ts
```diff
@@ -1,34 +1,39 @@
...
-import { ACTOR_ADDITIONAL_INSTRUCTIONS, defaults, MAX_DESCRIPTION_LENGTH } from './const.js';
+import { ACTOR_ADDITIONAL_INSTRUCTIONS, defaults, MAX_DESCRIPTION_LENGTH, ACTOR_README_MAX_LENGTH } from './const.js';
 import { log } from './logger.js';
 import type { ActorDefinitionPruned, ActorDefinitionWithDesc, SchemaProperties, Tool } from './types.js';
 
 export function actorNameToToolName(actorName: string): string {
-    return actorName.replace('/', '--');
+    return actorName
+        .replace(/\//g, '-slash-')
+        .replace(/\./g, '-dot-');
```
MQ37: what if the tool name is longer that 64 chars after the replace? Based on errors I got from working on 5ire.app MCP supports 64 chars at max. We should maybe use hash of the whole Actor name, slice to the say 60 chars and last 4 replace with the chars from hex digest of the hash.

jirispilka: Yeah, the problem I was trying to avoid is to have this lookup 
```
            const tool = Array.from(this.tools.values()).find((t) => t.name === name || t.actorFullName === name);
```
I faced issue with tool names vs Actor names

But I tested it now, with correct tool description it is working fine

jirispilka: I've decided to ignore name collision as of now. I don't think it will happen

MQ37: sure, we can ignore that for now :+1:

PR #42, opened 2025-03-14
fix: add enum values and examples to schema property descriptions

MQ37:


jirispilka (approved):
Looks good. Thanks

For future, let us document it: https://github.com/apify/actors-mcp-server/issues/43

PR #45, opened 2025-03-16
fix: tool schema array type infer and nested props

MQ37:
mastra.ai agent failed to run WCC because of the nested properties, the object input properties: had to infer their schema based on the editor provided; This helps the agent to call the WCC with the correct options.

### DIFF
--- a/src/actors.ts
+++ b/src/actors.ts
@@ -109,9 +109,9 @@ export function truncateActorReadme(readme: string, limit = ACTOR_README_MAX_LEN
 /**
  * Helps determine the type of items in an array schema property.
  * Priority order: explicit type in items > prefill type > default value type > editor type.
- * 
+ *
  * Based on JSON schema, the array needs a type, and most of the time Actor input schema does not have this, so we need to infer that.
- * 
+ *
  */
 export function inferArrayItemType(property: ISchemaProperties): string | null {
     return property.items?.type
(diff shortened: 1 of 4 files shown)


MQ37:
closes https://github.com/apify/actors-mcp-server/issues/44

src/actors.ts
```diff
@@ -143,21 +150,104 @@ export function addEnumsToDescriptionsWithExamples(properties: { [key: string]:
...
+                    ...filteredProperties[key].items,
+                    title: filteredProperties[key].title ?? 'Item',
+                    description: filteredProperties[key].description ?? 'Item',
+                    type: itemsType,
+                };
             }
         }
     }
     return filteredProperties;
 }
```
jirispilka: Would it be nicer like this?

I haven't tested it, only draft. 
```
export function filterSchemaProperties(properties: { [key: string]: ISchemaProperties }): { [key: string]: ISchemaProperties } {
    const filteredProperties: { [key: string]: ISchemaProperties } = {};
    for (const [key, property] of Object.entries(properties)) {
        filteredProperties[key] = {
            title: property.title,
            enum: property.enum,
            type: property.type,
            default: property.default,
            prefill: property.prefill,
            properties: property.properties,
            items: property.items,
            required: property.required,
        };
        if (property.type === 'array' && !property.items?.type) {
            const itemsType = inferArrayItemType(property);
            if (itemsType) {
                filteredProperties[key].items = {
                    ...filteredProperties[key].items,
                    title: filteredProperties[key].title ?? 'Item',
                    description: filteredProperties[key].description ?? 'Item',
                    type: itemsType,
                };
            }
        }
    }
    return filteredProperties;
}
```

jirispilka (changes requested):
It seems to be a bit complicated .... can you please provide more documentation, why we need it? Maybe including examples?

The doc will become handy for developing standalone MCP Serve (at the level of Apify Api)

src/actors.ts
MQ37: @jirispilka yes, I haven't refactored the original code, but I think once we have a fully functional version a refactor would be great :+1:

MQ37:
> It looks good to me. I hope you've tested it with a set of different Actors

Tested with set of different Actors and seems to be working, the only issue currently with mastra.ai agent is the timeout: when tool call takes longer than 60 s it times out. Will check with the mastra.ai guys.

PR #59, opened 2025-04-01
feat: Decouple Actor from mcp-server, add ability to call Actorized MCP and load tools

MQ37:
Closes https://github.com/apify/actors-mcp-server/issues/55
Closes https://github.com/apify/actors-mcp-server/issues/56

~~The idea is that we expose the MCP server and the Express app that can then be extended with other endpoints - so it can be used in the platform MCP server.~~

The new proposed solution is to expose MCP server and tools/prompt/resources functionalty as additional library and keep the stdio (and the Actor for now - legacy).

```

### CODE REVIEW

**Comment by jirispilka** [comment]
@MQ37 thanks! But it would be better to have separate PRs, it is becoming very big and intractable.

First, we need to decouple Actor and server. You also need to separate README, Actor README is going to be different from the package README

**Comment by MQ37** [comment]
> @MQ37 thanks! But it would be better to have separate PRs, it is becoming very big and intractable.
> 
> First, we need to decouple Actor and server. You also need to separate README, Actor README is going to be different from the package README

I agree with you that this is a large PR but I think these changes go hand in hand with each other and it would be easier to do both of them at once for us - I am also working on integrating this into the `apify-mcp-server` repo.

**Comment by jirispilka** [comment]
I'm afraid, I don't really like it, there is no clear separation of the Actor code and a new mcp-server code :(
That was the requirement of the issue.
Instead, there is a mix of tools.ts, toolkits, tools directory, which I don't quite get.

**Comment by MQ37** [comment]
> I'm afraid, I don't really like it, there is no clear separation of the Actor code and a new mcp-server code :( That was the requirement of the issue. Instead, there is a mix of tools.ts, toolkits, tools directory, which I don't quite get. I would first finish one thing, settle the structure and then introduce a new stuff

It is still a WIP, but the main idea is that all tools will be in the `tools/` directory (same for `resources/` and `prompts/` as suggested in https://github.com/apify/actors-mcp-server/issues/56) and the server entry points and MCP server class are separated (so we can easily use this in the `apify-mcp-server` based on https://github.com/apify/actors-mcp-server/issues/55). So the `main.ts` is the entry point for the Actor and `stdio.ts` is the entry point for the stdio version of the server. There are some places that still need work and some moving around. What other project structure do you suggest then?

**Comment by jirispilka** [comment]
I think the structure is better now .... but this PR is huge because we couple together structure and tools refactoring. 
We will need to live with that and test properly.

There are two things I'm not happy with:
- I don't like that we are passing `apifyMcpServer` and `mcpServer` to a tool
- We need to handle errors better by sending `server.sendLoggingMessage (   )`

So this is still work in progress

File: src/actor/const.ts [hunk 1]
```diff
@@ -0,0 +1,20 @@
...
+ * Constants for the Actor.
+ */
+export const HEADER_READINESS_PROBE = 'x-apify-container-server-readiness-probe';
+
+export const defaults = {
+    actors: [
+        'apify/instagram-scraper',
+        'apify/rag-web-browser',
+        'lukaskrivka/google-maps-with-contact-details',
+    ],
```
MQ37: The default list of loaded Actors/tools is not only Actor related (is used in `stdio.ts`), I think we should move it to `src/const.ts`

src/main.ts
```diff
@@ -1,157 +1,58 @@
...
+const input = processInput((await Actor.getInput<Partial<Input>>()) ?? ({} as Input));
 log.info(`Loaded input: ${JSON.stringify(input)} `);
 
 if (STANDBY_MODE) {
+    const app = createExpressApp(HOST, mcpServer);
     log.info('Actor is running in the STANDBY mode.');
-    await mcpServer.addToolsFromDefaultActors();
-    mcpServer.updateTools(getActorDiscoveryTools());
+    const tools = [searchTool, actorDefinitionTool];
+    mcpServer.updateTools([searchTool, actorDefinitionTool]);
```
MQ37:
```suggestion
```

update `mcpServer` tools only once. They are update also below the `if (input.enableActorAutoLoading) {` block.

src/main.ts
```diff
@@ -1,157 +1,58 @@
...
-const input = await processInput((await Actor.getInput<Partial<Input>>()) ?? ({} as Input));
+const input = processInput((await Actor.getInput<Partial<Input>>()) ?? ({} as Input));
 log.info(`Loaded input: ${JSON.stringify(input)} `);
 
 if (STANDBY_MODE) {
+    const app = createExpressApp(HOST, mcpServer);
     log.info('Actor is running in the STANDBY mode.');
-    await mcpServer.addToolsFromDefaultActors();
-    mcpServer.updateTools(getActorDiscoveryTools());
+    const tools = [searchTool, actorDefinitionTool];
```
MQ37, line 38: Are we not loading the default tools for the standby server now? So the user needs to specify that in the request?

src/tools/actor.ts
```diff
@@ -0,0 +1,122 @@
...
+    markInputPropertiesAsRequired,
+    shortenProperties,
+} from './utils.js';
+
+/**
+ * Adds a User-Agent header to the request config.
+ * @param config
+ * @private
+ */
+export function addUserAgent(config: AxiosRequestConfig): AxiosRequestConfig {
```
MQ37: in future we should move this somewhere else, because it would be used by the Apify client for dataset ops and so on, right?

MQ37 (commented):
Just a few comments, otherwise nice work @jirispilka  :+1:

src/main.ts
MQ37: fixed

src/mcp-server.ts
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    private setupToolHandlers(): void {
+        /**
+         * Handles the request to list tools.
+         * @param {object} request - The request object.
+         * @returns {object} - The response object containing the tools.
+         */
+        this.server.setRequestHandler(ListToolsRequestSchema, async () => {
+            const tools = Array.from(this.tools.values()).map((tool) => (tool.tool));
```
jirispilka:
```suggestion
            // TODO if there is actor-mcp as a tool, also list the tools from that Actor 
            const tools = Array.from(this.tools.values()).map((tool) => (tool.tool));
```

src/tools/actor.ts
MQ37: created subclass with the request interceptor in `mpc-apify-client.ts`

jirispilka:
I know, one test is failing x I didn't have time to do it before leaving the office.

MQ37:
@jirispilka thank you for fixing the tests :+1: Tested Actorized MCPs and normal Actors in Claude desktop (stdio) and on staging via SSE and everything works :heavy_check_mark:

jirispilka:
@MQ37 cool. Thanks! We should also test existing Actor, whether it works via SSE. I'll create a task for this

jirispilka (approved):

MQ37:
@jirispilka thank you for fixing the tests :+1: Tested Actorized MCPs and normal Actors in Claude desktop (stdio) and on staging via SSE and everything works :heavy_check_mark: 

> @MQ37 cool. Thanks! We should also test existing Actor, whether it works via SSE. I'll create a task for this

Tested again as Actor and it works :+1:

Issue #54, opened 2025-03-29
Add `--enable-adding-actors` param to NPM command


jancurn:
Some clients will add support for this, so it would be good to have that option in the local MCP command.

Also, instead of `enableActorAutoLoading` I'd call it simply `enableAddingActors` (we might have to keep backwards compatibility)

PR #63, opened 2025-04-08
fix: deprecate enableActorAutoLoading in favor of enable-adding-actors, and load tools only if not provided in query parameter

jirispilka:
fix: replace minimist with yargs-parser, deprecate enableActorAutoLoading in favor of enable-adding-actors, and load tools only if not provided in query parameter

Closes: #54


MQ37 (approved):
LGMT :+1:

PR #90, opened 2025-04-22
fix: move logic to enableAddingActors and enableDefaultActors to constructor

jirispilka:


MQ37 (approved):
tested locally in Claude desktop (STDIO) and LGTM :+1:

PR #91, opened 2025-04-22
fix: code improvements

MQ37:
**Changes:**
- better error handling and logging

### DIFF
--- a/.actor/ACTOR.md
+++ b/.actor/ACTOR.md
@@ -1,7 +1,6 @@
 # Apify Model Context Protocol (MCP) Server
 
 [![Actors MCP Server](https://apify.com/actor-badge?actor=apify/actors-mcp-server)](https://apify.com/apify/actors-mcp-server)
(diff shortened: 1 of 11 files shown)


jirispilka (changes requested):
Let's not complicate our lives with caching here. 

I think all request are cached at infra level. There is no need to add extra caching here

MQ37:
> Let's not complicate our lives with caching here.
> 
> I think all request are cached at infra level. There is no need to add extra caching here

makes sense, removed the caching stuff :+1:

jirispilka (changes requested):
Is there a reason why, this code should span multiple-lines?
```
                throw new McpError(
                    ErrorCode.InvalidParams,
                    msg,
                );
```
I think it would fit to one line quite well. The same for other errors, it can be a single line.

MQ37:
> Is there a reason why, this code should span multiple-lines?
> 
> ```
>                 throw new McpError(
>                     ErrorCode.InvalidParams,
>                     msg,
>                 );
> ```
> 
> I think it would fit to one line quite well. The same for other errors, it can be a single line.

There is no specific reason. I think should not address formatting/non-functional code style issues. A formatter should handle this once set up.

PR #103, opened 2025-04-29
chore: add tests

MQ37:
closes https://github.com/apify/actors-mcp-server/issues/57
closes https://github.com/apify/actors-mcp-server/issues/97

**by default for `npm run test` only unit tests are executed as they do not require Apify API token**

Features:
- more unit tests
- stdio integration tests
- Actor MCP server SSE and streamable HTTP integration tests
- helper tool


src/actor/utils.ts
```diff
@@ -26,3 +28,12 @@ export function getActorRunData(): ActorRunData | null {
         standbyUrl: process.env.ACTOR_STANDBY_URL,
     } : null;
 }
+
+export function createActorMCPServer(
```
jirispilka: I'm sorry, I don't really like this function. It is called only two times. I don't think it is required to create a wrapper for it

tests/integration/actor.server.test.ts
```diff
@@ -0,0 +1,440 @@
...
+    let server: ActorsMcpServer;
+    let httpServer: HttpServer;
+    const testPort = 50000;
+    const testHost = `http://localhost:${testPort}`;
+    const serverStartWaitTimeMillis = 100;
+
+    beforeEach(async () => {
+        // same as in main.ts
+        // TODO: unify
+        server = createActorMCPServer();
```
jirispilka: server is created twice. I would really remove the createActorMCPServer function. Can you handle TODO as well?

src/createActorMCPServer.ts
jirispilka, line 1: Empty file, please delete

src/main.ts
```diff
@@ -30,11 +30,7 @@ const input = processInput((await Actor.getInput<Partial<Input>>()) ?? ({} as In
 log.info(`Loaded input: ${JSON.stringify(input)} `);
 
 if (STANDBY_MODE) {
-    const mcpServer = new ActorsMcpServer({
-        enableAddingActors: Boolean(input.enableAddingActors),
-        enableDefaultActors: false,
-    });
-
+    const mcpServer = createActorMCPServer(input);
```
jirispilka: Let's not use the createActorMCPSever, we can create the class right away

tests/unit/tools.utils.test.ts
```diff
@@ -0,0 +1,374 @@
...
+        };
+
+        const result = shortenProperties(properties);
+
+        // Check that properties were not modified
+        expect(result).toEqual(properties);
+    });
+});
+
+describe('actors', () => {
```
jirispilka: This code duplicates actor.test.ts

jirispilka (changes requested):

jirispilka (changes requested):
I didn’t have time to look at the test in detail,  
but overall, it looks good. I’ve left a few comments that should be addressed before merging.

src/createActorMCPServer.ts
MQ37: did not notice, thank you :+1:

tests/unit/tools.utils.test.ts
MQ37: previously I wanted to move that and probably forgot, thank you for noticing :+1: Removed this duplicate code and renamed `actor.test.ts` to `tools.actor.test.ts`

src/actor/utils.ts
MQ37: I created this so the tests and actual Actor logic is consistent and defined in single place - let's use the class right away :+1:

tests/integration/actor.server.test.ts
MQ37: Removed the `createActorMCPServer` wrapper and TODO since I copied the logic from `main.ts` so it's unified. 

> server is created twice.

You mean there is separate server for each `describe` block? If so that is correct, I wanted to split the different transport tests so they can run in parallel so it spawns two server instances on different ports. Since the Actor MCP server logic is coupled and only single-tenant the actual tests (`it` blocks) must run in serial.

PR #114, opened 2025-05-13
fix: Use a new API to get Actor default build

jirispilka:


MQ37 (approved):
haven't tested but LGTM :+1:

PR #117, opened 2025-05-15
feat: normal Actor tools cache

MQ37:


src/mcp/actors.ts
```diff
@@ -56,26 +56,12 @@ export async function getActorStandbyURL(actorID: string, apifyToken: string, st
     return `https://${actorRealID}.${standbyBaseUrl}`;
 }
 
+// TODO: optimize this to only use /builds/default endpoint with Actor name replacing / with ~
```
jirispilka: I don't think we want to mix sdk use and rest api use. Please remove this TODO

src/mcp/actors.ts
```diff
@@ -56,26 +56,12 @@ export async function getActorStandbyURL(actorID: string, apifyToken: string, st
...
 export async function getActorDefinition(actorID: string, apifyToken: string): Promise<ActorDefinition> {
```
jirispilka, line 59: On top of that. This function duplicates: getActorDefinition in build.ts. It would be better to unify it right away

src/tools/actor.ts
```diff
@@ -58,6 +59,10 @@ export async function callActorGetDataset(
     }
 }
 
+// Cache for normal Actor tools
+const normalActorToolsCache = new LruCache<ToolCacheEntry>({
```
jirispilka: I would put this at the top of the file

src/tools/actor.ts
```diff
@@ -83,13 +88,31 @@ export async function getNormalActorsAsTools(
...
+    const tools: ToolWrap[] = [];
+    const actorsLoadedFromCache: string[] = [];
+    for (const actorID of actors) {
+        const cacheEntry = normalActorToolsCache.get(actorID);
+        if (cacheEntry && cacheEntry.expiresAt > Date.now()) {
+            tools.push(cacheEntry.tool);
+            actorsLoadedFromCache.push(actorID);
+        }
+    }
+    const actorsToLoad = actors.filter((actorID) => !actorsLoadedFromCache.includes(actorID));
```
jirispilka:
```suggestion
    const tools: ToolWrap[] = [];
    const actorsToLoad: string[] = [];
    for (const actorID of actors) {
        const cacheEntry = normalActorToolsCache.get(actorID);
        if (cacheEntry && cacheEntry.expiresAt > Date.now()) {
            tools.push(cacheEntry.tool);
        } else {
            actorsToLoad.push(actorID);
        }
    }
```

src/tools/actor.ts
```diff
@@ -83,13 +88,31 @@ export async function getNormalActorsAsTools(
...
     const getActorDefinitionWithToken = async (actorId: string) => {
         return await getActorDefinition(actorId, apifyToken);
     };
-    const results = await Promise.all(actors.map(getActorDefinitionWithToken));
-    const tools: ToolWrap[] = [];
-    for (const result of results) {
+    const results = await Promise.all(actorsToLoad.map(getActorDefinitionWithToken));
+
+    // Zip the results with their corresponding actorIDs
+    for (let i = 0; i < results.length; i++) {
```
jirispilka, line 113: Please, you don't have to do change this! You already have actorId in result, see: actorDefinitions.id = actor.id;

jirispilka (changes requested):

src/mcp/actors.ts
MQ37: I would keep that, when SDK is not optimal we can write better solution (1 req vs 2 reqs). At scale it makes a difference - we don't have to solve that now but we can keep the TODO

src/mcp/actors.ts
MQ37: Agree, we need to rework that logic but I did not want to touch that since you made changes there (at least I think you mentioned that)

src/tools/actor.ts
MQ37: Did not notice that, thank you for mentioning :+1:

src/mcp/actors.ts
MQ37: will remote TODO and create issue instead :+1:

src/mcp/actors.ts
MQ37: https://github.com/apify/actors-mcp-server/issues/120

src/mcp/actors.ts
MQ37: https://github.com/apify/actors-mcp-server/issues/119

src/tools/actor.ts
MQ37: We aren't storing the original user input in `result`. We encourage users to use `{user}/{actor}`, but they can supply just `{actorObjectID}` (like `3ox4R101TgZz67sLr`). This is why the param is named `actorIdOrName`. This won't work since `getActorDefinition` isn't storing that value. We need the exact input from the user because if they use `actorObjectID`, it would always miss the cache since we only store the resolved value, and we cannot retrieve `actorObjectID` without an API call. So, the cache would not make sense if we always store the entry with `actorObjectID` key. We need to handle both possibilities.

MQ37: will rename the variable to `actorIdOrName` instead of `actorID` so it makes more sense and add a comment

jirispilka: ohh, you are right! I missed that, thanks

jirispilka (approved):

jirispilka (approved):
Thank you!

PR #122, opened 2025-05-16
feat: Add Actor runs API, dataset API

jirispilka:
- **fix: Use a new API to get Actor default build`**
- **feat: add get-actor and get-user-runs-list**
- **feat: add dataset, Actor runs, fix tests**

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -203,7 +203,6 @@ export class ActorsMcpServer {
             }
 
             // TODO - if connection is /mcp client will not receive notification on tool change
-
             // Find tool by name or actor full name
             const tool = Array.from(this.tools.values())
                 .find((t) => t.tool.name === name || (t.type === 'actor' && (t.tool as ActorTool).actorFullName === name));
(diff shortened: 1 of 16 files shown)


src/tools/actor.ts
```diff
@@ -31,29 +39,32 @@ import {
...
  */
 export async function callActorGetDataset(
     actorName: string,
     input: unknown,
     apifyToken: string,
     callOptions: ActorCallOptions | undefined = undefined,
-): Promise<object[]> {
-    const name = actorName;
+    limit = ACTOR_RUN_DATASET_OUTPUT_MAX_ITEMS,
+): Promise<{ actorRun: ActorRun, datasetInfo: Dataset | undefined, items: PaginatedList<Record<string, unknown>> }> {
```
MQ37: I would create type for this `{ actorRun: ActorRun, datasetInfo: Dataset | undefined, items: PaginatedList<Record<string, unknown>> }`

MQ37 (approved):
Just a minor comment otherwise LGTM :+1:

jirispilka:
@MQ37 This PR has grown quite a bit — sorry about that! Once again, we ended up working on the same part of the code, and when I started fixing the unit tests, I went down a rabbit hole and couldn’t stop myself. 😅

I really like the tests you added — they’re super helpful, thanks!

That said, there’s one test I believe is tied to the reset function, which always loads the default Actor. So even when the server is called with a specific Actor selected, the default one still gets loaded — which shouldn’t happen.

We’ll need to fix that.

Also, I ran into an issue where, after the test completed, the process was still hanging around and consuming 100% CPU.

If you could please review again. Mainly the test suite and mcp/server.ts 🙏🏻

src/actor/server.ts
```diff
@@ -68,12 +68,13 @@ export function createExpressApp(
         try {
             log.info(`Received GET message at: ${Routes.SSE}`);
             const input = parseInputParamsFromUrl(req.url);
-            if (input.actors || input.enableAddingActors) {
```
MQ37, line 71: `loadToolsFromUrl` already handler Actor adding tools, see `loadToolsFromUrl -> processParamsGetTools`. Let's keep it like this (it is more readable) but we should rework in future.

src/mcp/server.ts
```diff
@@ -126,75 +116,78 @@ export class ActorsMcpServer {
...
+        return Array.from(this.tools.values())
+            .filter((tool) => tool.type === 'actor')
+            .map((tool) => (tool.tool as ActorTool).actorFullName);
+    }
+
+    /**
+     * Returns a list of Actor IDs that are registered as MCP servers.
+     * @returns {string[]} - An array of Actor MCP server Actor IDs (e.g., 'apify/actors-mcp-server').
+     */
+    private listActorMcpServerToolIds(): string[] {
```
MQ37, line 142: original implementation used `set` to list only unique Actor IDs/names and since there may be multiple tools from single Actorized MCP server this will list the server Actor multiple times. Please implement it using set as in the original implementation

jirispilka: Thanks, good point!

PR #126, opened 2025-05-20
feat: Experimenting with oauth

valekjo:
Seems that we won't need any changes related to MCP + oAuth in the Actor itself :)

But I struggled with this when running the platform locally.

Specs: https://www.notion.so/apify/Apify-as-OAuth-Provider-Docs-192f39950a22807b83a0c8fcbdac90cc

### DIFF
--- a/src/apify-client.ts
+++ b/src/apify-client.ts
@@ -16,17 +16,10 @@ function addUserAgent(config: AxiosRequestConfig): AxiosRequestConfig {
     return updatedConfig;
 }
 
-export function getApifyAPIBaseUrl(): string {
-    // Workaround for Actor server where the platform APIFY_API_BASE_URL did not work with getActorDefinition from actors.ts
-    if (process.env.APIFY_IS_AT_HOME) return 'https://api.apify.com';
-    return process.env.APIFY_API_BASE_URL || 'https://api.apify.com';
-}
-
 export class ApifyClient extends _ApifyClient {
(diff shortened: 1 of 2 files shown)


src/apify-client.ts
```diff
@@ -16,17 +16,10 @@ function addUserAgent(config: AxiosRequestConfig): AxiosRequestConfig {
     return updatedConfig;
 }
 
-export function getApifyAPIBaseUrl(): string {
-    // Workaround for Actor server where the platform APIFY_API_BASE_URL did not work with getActorDefinition from actors.ts
-    if (process.env.APIFY_IS_AT_HOME) return 'https://api.apify.com';
-    return process.env.APIFY_API_BASE_URL || 'https://api.apify.com';
-}
```
valekjo, line 23: Note: The workaround did not work when running the platform locally, and probably would not work on staging as well.
Where was the `APIFY_API_BASE_URL` env var missing?

jirispilka: Great catch, this was actually a hack from @MQ37 

This package is imported in `apify-mcp-server` where `APIFY_API_BASE_URL` is defined 🙈.
Here is a TODO for that https://github.com/apify/apify-mcp-server/blob/0b336190178ad85beb1ee83956924d98c0eb129b/src/main.ts#L18

I'm not sure if we'll have time to fix it this week.

PR #130, opened 2025-05-30
fix: error and input handling

MQ37:
- fixes tool error and input handling
- fixed test suite inconsistencies

### DIFF
--- a/src/tools/actor.ts
+++ b/src/tools/actor.ts
@@ -245,11 +245,20 @@ export const getActor: ToolEntry = {
         ajvValidate: ajv.compile(zodToJsonSchema(getActorArgs)),
         call: async (toolArgs) => {
             const { args, apifyToken } = toolArgs;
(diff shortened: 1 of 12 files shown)


src/tools/actor.ts
```diff
@@ -1,6 +1,6 @@
 import type { Client } from '@modelcontextprotocol/sdk/client/index.js';
 import { Ajv } from 'ajv';
-import type { ActorCallOptions, ActorRun, Dataset, PaginatedList } from 'apify-client';
+import { type ActorCallOptions, type ActorRun, type Dataset, type PaginatedList } from 'apify-client';
```
jirispilka: This look unnecessary verbose

jirispilka (approved):
Looks good, only one nit (questions)

src/tools/actor.ts
MQ37: yeah, vscode :smile:

PR #136, opened 2025-06-11
fix: explicitly clear resources

jirispilka:
Explicitly clear resources in close() function. But it might not help with memory leak (unless there was some handler registered)


MQ37 (approved):
LGTM 👍

Issue #80, opened 2025-04-16
Deploy pre-release (beta) using label, remove pre-release when label is removed and PR is merged?


jirispilka:

MQ37:
it seems that there is not way to run a CICD job when a tag is removed

MQ37:
another alternative solution is to use github branches for beta releases that we can later remove (NPM can install from github). But for simplicity I would keep the NPM implementation

jirispilka:
> another alternative solution is to use github branches for beta releases that we can later remove (NPM can install from github). But for simplicity I would keep the NPM implementation

Can you please add the comments into the PR description, this is difficult to track

PR #142, opened 2025-06-13
fix: disable search for rental Actors

MQ37:
closes https://github.com/apify/actors-mcp-server/issues/135

### DIFF
--- a/src/tools/store_collection.ts
+++ b/src/tools/store_collection.ts
@@ -39,7 +39,7 @@ export async function searchActorsByKeywords(
     limit: number | undefined = undefined,
     offset: number | undefined = undefined,
     pricingModel: ApifyStorePricingModel | undefined = undefined,
-): Promise<ActorStorePruned[] | null> {
+): Promise<ActorStorePruned[]> {
     const client = new ApifyClient({ token: apifyToken });
     const results = await client.store().list({ search, limit, offset, pricingModel });
     return results.items.map((x) => pruneActorStoreInfo(x));
--- a/src/tools/store_collection.ts
+++ b/src/tools/store_collection.ts
@@ -69,19 +69,26 @@ export const searchActorsArgsSchema = z.object({
 });
 
 /**
- * Filters out actors with the 'FLAT_PRICE_PER_MONTH' pricing model (rental actors)..
+ * Filters out actors with the 'FLAT_PRICE_PER_MONTH' pricing model (rental actors),
+ * unless the actor's ID is present in the user's rented actor IDs list.
  *
- * Returns new array of Actors excluding those with 'FLAT_PRICE_PER_MONTH' pricing model.
+ * This is necessary because the Store list API does not support filtering by multiple pricing models at once.
  *
  * @param actors - Array of ActorStorePruned objects to filter.
- * @returns Array of actors excluding those with 'FLAT_PRICE_PER_MONTH' pricing model.
+ * @param userRentedActorIds - Array of actor IDs that the user has rented.
+ * @returns Array of actors excluding those with 'FLAT_PRICE_PER_MONTH' pricing model,
+ *          except for actors that the user has rented (whose IDs are in userRentedActorIds).
  */
 function filterRentalActors(
     actors: ActorStorePruned[],
+    userRentedActorIds: string[],
 ): ActorStorePruned[] {
     // Store list API does not support filtering by two pricing models at once,
     // so we filter the results manually after fetching them.
-    return actors.filter((actor) => (actor.currentPricingInfo.pricingModel as ActorPricingModel) !== 'FLAT_PRICE_PER_MONTH');
+    return actors.filter((actor) => (
+        actor.currentPricingInfo.pricingModel as ActorPricingModel) !== 'FLAT_PRICE_PER_MONTH'
+        || userRentedActorIds.includes(actor.id),
+    );
 }
 
 /**
@@ -102,7 +109,7 @@ export const searchActors: ToolEntry = {
         inputSchema: zodToJsonSchema(searchActorsArgsSchema),
         ajvValidate: ajv.compile(zodToJsonSchema(searchActorsArgsSchema)),
         call: async (toolArgs) => {
-            const { args, apifyToken } = toolArgs;
+            const { args, apifyToken, userRentedActorIds } = toolArgs;
             const parsed = searchActorsArgsSchema.parse(args);
             let actors = await searchActorsByKeywords(
                 parsed.search,
@@ -110,7 +117,7 @@ export const searchActors: ToolEntry = {
                 parsed.limit + ACTOR_SEARCH_ABOVE_LIMIT,
                 parsed.offset,
             );
-            actors = filterRentalActors(actors || []).slice(0, parsed.limit);
+            actors = filterRentalActors(actors || [], userRentedActorIds || []).slice(0, parsed.limit);
 
             return { content: actors?.map((item) => ({ type: 'text', text: JSON.stringify(item) })) };
         },
--- a/src/tools/store_collection.ts
+++ b/src/tools/store_collection.ts
@@ -76,7 +76,8 @@ export const searchActorsArgsSchema = z.object({
  *
  * @param actors - Array of ActorStorePruned objects to filter.
  * @param userRentedActorIds - Array of Actor IDs that the user has rented.
- * @returns Array of Actors excluding those with 'FLAT_PRICE_PER_MONTH' pricing model (= rental Actors), except for Actors that the user has rented (whose IDs are in userRentedActorIds).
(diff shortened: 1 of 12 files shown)


src/types.ts
```diff
@@ -190,3 +190,5 @@ export interface ToolCacheEntry {
     expiresAt: number;
     tool: ToolEntry;
 }
+
+export type ApifyStorePricingModel = 'FREE' | 'FLAT_PRICE_PER_MONTH' | 'PRICE_PER_DATASET_ITEM';
```
jirispilka: can we define it the same way as in apify-core?

```
export const ACTOR_PRICING_MODEL = {
    /** Rental actors */
    FLAT_PRICE_PER_MONTH: 'FLAT_PRICE_PER_MONTH',
    FREE: 'FREE',
    /** Pay per result (PPR) actors */
    PRICE_PER_DATASET_ITEM: 'PRICE_PER_DATASET_ITEM',
    /** Pay per event (PPE) actors */
    PAY_PER_EVENT: 'PAY_PER_EVENT',
} as const;
export type ACTOR_PRICING_MODEL = ValueOf<typeof ACTOR_PRICING_MODEL>
```
and then use ACTOR_PRICING_MODEL.FREE

src/tools/store_collection.ts
```diff
@@ -68,6 +69,61 @@ export const searchActorsArgsSchema = z.object({
         .describe('Filters the results by the specified category.'),
 });
 
+/**
+ * Filters out actors with the 'FLAT_PRICE_PER_MONTH' pricing model (rental actors).
+ *
+ * @param actors - Array of ActorStorePruned objects to filter.
+ * @returns Array of actors excluding those with 'FLAT_PRICE_PER_MONTH' pricing model.
+ */
+function filterRentalActors(
```
jirispilka, line 82: what about removeRentalActors, that would be more descriptive?

src/tools/store_collection.ts
```diff
@@ -68,6 +69,61 @@ export const searchActorsArgsSchema = z.object({
...
+ * Fallback function to fetch actors if no rental actors are found.
+ * Fetches both free and pay-per-result actors and merges them in a zig-zag order.
+ *
+ * @param search - Search keywords for actors.
+ * @param apifyToken - Apify API token.
+ * @param limit - Maximum number of actors to return.
+ * @param offset - Number of actors to skip from the start.
+ * @returns Array of ActorStorePruned objects, alternating between free and pay-per-result actors.
+ */
+async function getFallbackActors(
```
jirispilka: getFallbackActors, what does it mean? I would put it explicitly (search ppe,free,pay par item). IMO, you are missing PPE here!

src/tools/store_collection.ts
```diff
@@ -88,12 +144,25 @@ export const searchActors: ToolEntry = {
         call: async (toolArgs) => {
             const { args, apifyToken } = toolArgs;
             const parsed = searchActorsArgsSchema.parse(args);
-            const actors = await searchActorsByKeywords(
+            let actors = await searchActorsByKeywords(
                 parsed.search,
                 apifyToken,
                 parsed.limit,
```
jirispilka: Alternative approach: Increase the limit to a higher number, for example, parsed.limit + 50. Then, filter out rentals and finally use slice(0, parsed.limit). Although this method is inefficient, the extra data transfer isn't very costly, and it would simplify the implementation and we wouldn't need extra calls to get PPE, FREE, and PRICE_PER_DATASET_ITEM (3 extra calls anyway).

src/tools/store_collection.ts
```diff
@@ -68,6 +69,61 @@ export const searchActorsArgsSchema = z.object({
...
+ * @param offset - Number of actors to skip from the start.
+ * @returns Array of ActorStorePruned objects, alternating between free and pay-per-result actors.
+ */
+async function getFallbackActors(
+    search: string,
+    apifyToken: string,
+    limit: number | undefined,
+    offset: number | undefined,
+): Promise<ActorStorePruned[]> {
+    const freeActors = await searchActorsByKeywords(
```
jirispilka: If you want to keep this logic (I would not). Let's do it Promise.all([])

jirispilka (changes requested):

jirispilka (changes requested):
Thank you!

However, I would simplify the implementation and take a different approach—please see my comments.

There’s a major issue, though: What if the user has rental Actors?
We should also search for Actors the user is renting. For that, we’d need to access the database... because this server can also be run locally

If we want to support this in actors-mcp-server, we’ll need to introduce a new API.

FYI @mtrunkat

src/tools/store_collection.ts
MQ37: thank you for noticing, I am missing the PPE - I copied possible values from https://docs.apify.com/api/v2/store-get where we don't mention PPE (will notify others on slack). Will rename to something else - this func is used as fallback non-rental Actors are not found in the first search

src/tools/store_collection.ts
MQ37: remove could imply that it changes the input array but this returns a new, filtered out, array instead. But we can rename that, it's styling/subjective thing

jirispilka:
I discussed this with Mara. We should also return rental Actors for a particular user.

However, since the Actors MCP server can be run locally without a database connection:

- For local deployments, we won’t return rentals at all.
- For remote deployments, we can pass a MongoDB connection to the Actors MCP server and handle users rentals there.

Later on, we can introduce an API to provide this functionality. However, I'm not sure if such an API would be needed for anything else. I'll share this in Product Insights.

src/tools/store_collection.ts
```diff
@@ -68,6 +69,61 @@ export const searchActorsArgsSchema = z.object({
...
+        'PRICE_PER_DATASET_ITEM',
+    );
+    const allActors: ActorStorePruned[] = [];
+    // Push Actors in zig-zag order to ensure that we return all Actors
+    // in relevant order.
+    const maxLength = Math.max(freeActors?.length || 0, payPerResultActors?.length || 0);
+    for (let i = 0; i < maxLength; i++) {
+        if (freeActors && freeActors[i]) allActors.push(freeActors[i]);
+        if (payPerResultActors && payPerResultActors[i]) allActors.push(payPerResultActors[i]);
+    }
```
MichalKalita: I am not sure if lodash is available in this project. This code will be better readable when `zip` function is used https://lodash.com/docs/4.17.15#zip

src/tools/store_collection.ts
MQ37: you are right, let's use this much simpler logic, three separate calls and their merging is complicated :+1:

MichalKalita (approved):

MQ37:
@jirispilka @MichalKalita updated to logic, we are now passing list of users rented Actor IDs from `apify-mcp-server` to the search store tool call and handling that here without passing any mongo connection (I don't think we should do this in open source repo and internal MongoDB structure should remain hidden). If list of rented Actors is passed we filter out all no-rented rental Actors, if not then we just filter out all rental Actors.

please see https://github.com/apify/apify-mcp-server/pull/80

src/tools/store_collection.ts
```diff
@@ -68,6 +68,29 @@ export const searchActorsArgsSchema = z.object({
...
+ * unless the actor's ID is present in the user's rented actor IDs list.
+ *
+ * This is necessary because the Store list API does not support filtering by multiple pricing models at once.
+ *
+ * @param actors - Array of ActorStorePruned objects to filter.
+ * @param userRentedActorIds - Array of actor IDs that the user has rented.
+ * @returns Array of actors excluding those with 'FLAT_PRICE_PER_MONTH' pricing model,
+ *          except for actors that the user has rented (whose IDs are in userRentedActorIds).
+ */
+function filterRentalActors(
```
jirispilka:
```suggestion
/**
 * Filters out actors with the 'FLAT_PRICE_PER_MONTH' pricing model (rental actors),
 * unless the actor's ID is present in the user's rented actor IDs list.
 *
 * This is necessary because the Store list API does not support filtering by multiple pricing models at once.
 *
 * @param actors - Array of ActorStorePruned objects to filter.
 * @param userRentedActorIds - Array of Actor IDs that the user has rented.
 * @returns Array of Actors excluding those with 'FLAT_PRICE_PER_MONTH' pricing model (= rental Actors), except for Actors that the user has rented (whose IDs are in userRentedActorIds).
 */
function filterRentalActors(
```

src/types.ts
```diff
@@ -199,3 +202,6 @@ export interface ToolCacheEntry {
     expiresAt: number;
     tool: ToolEntry;
 }
+
+// Utility type to get the union of values of an object type
```
jirispilka:
```suggestion
// Utility type to get a union of values from an object type
```

src/tools/store_collection.ts
jirispilka: It has somehow messed up the diff. The only thing I did was to change actor -> Actor (at all places)

Issue #64, opened 2025-04-09
How to add parameters to apify/rag-web-browser


xldistance:
I would like to add {"requestTimeoutSecs": 60,"scrapingTool": "browser-playwright"} parameter to apify/rag-web-browser, how do I change        
 ```
"actors-mcp-server": {
            "command": "npx",
            "args": ["-y", "@apify/actors-mcp-server"
        ],
            "env": {
                "APIFY_TOKEN": "xxx"
            },
        },
```

jirispilka:
Hi  I'm not sure about your use-case, I have tried, and it works, see attached screenshot and this [run](https://console.apify.com/view/runs/APTT8qQocoW01KZP4) Apify console

![Image](https://github.com/user-attachments/assets/201131cd-0fcf-4d54-8aa3-299a454d7651)

PR #145, opened 2025-06-16
fix: notifications

MQ37:
Use MCP request handler extras for notification sending - this is needed so streamable POST SSE notifications work correctly. Also fixes notifications in streamable transport for the standby Actor mode which were not working - this needs POST SSE but it should be safe because the SSE sessions are usually short-lived until all tool call or other requests are finished and then new is created for each POST request.

prerequisite to https://github.com/apify/apify-mcp-server/pull/69
related to https://github.com/apify/apify-mcp-server/issues/64

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,4 +1,6 @@
 import type { Server } from '@modelcontextprotocol/sdk/server/index.js';
+import type { RequestHandlerExtra } from '@modelcontextprotocol/sdk/shared/protocol.js';
+import type { Notification, Request } from '@modelcontextprotocol/sdk/types.js';
 import type { ValidateFunction } from 'ajv';
 import type { ActorDefaultRunOptions, ActorDefinition } from 'apify-client';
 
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -80,6 +82,13 @@ export interface ActorTool extends ToolBase {
 export type InternalToolArgs = {
     /** Arguments passed to the tool */
     args: Record<string, unknown>;
+    /** Extra data given to request handlers.
+     *
+     * Can be used to send notifications from the server to the client.
+     *
+     * For more details see: https://github.com/modelcontextprotocol/typescript-sdk/blob/f822c1255edcf98c4e73b9bf17a9dd1b03f86716/src/shared/protocol.ts#L102
+     */
```
jirispilka, line 90:
```suggestion
     * For more details see: https://github.com/modelcontextprotocol/typescript-sdk/blob/f822c1255edcf98c4e73b9bf17a9dd1b03f86716/src/shared/protocol.ts
     */
```
I would remove the link to a specific line of code. It'll be obsolete. Sooner or later, someone will update this file, and it will point to the wrong line. So it is better to provide less specific reference than a wrong one.

jirispilka (approved):
One small nit. Otherwise, ok. Thank you!

src/types.ts
MQ37: This is why the link is blob, it should thus point to the specific, pinned, version of file

jirispilka: Oh I see, fair enough

PR #147, opened 2025-06-18
feat: dynamic actor loading is enabled by default

MichalKalita:
closes https://github.com/apify/actors-mcp-server/issues/144

### DIFF
--- a/src/input.ts
+++ b/src/input.ts
@@ -24,7 +24,7 @@ export function processInput(originalInput: Partial<Input>): Input {
             log.warning('enableActorAutoLoading is deprecated, use enableAddingActors instead');
             input.enableAddingActors = input.enableActorAutoLoading === true || input.enableActorAutoLoading === 'true';
         } else {
-            input.enableAddingActors = false;
+            input.enableAddingActors = true;
         }
     } else {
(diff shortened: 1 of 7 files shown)


MQ37 (changes requested):
integration tests are failing, please check with `APIFY_TOKEN="your token" npm run test:integration` and fix them

src/input.ts
```diff
@@ -24,7 +24,7 @@ export function processInput(originalInput: Partial<Input>): Input {
             log.warning('enableActorAutoLoading is deprecated, use enableAddingActors instead');
             input.enableAddingActors = input.enableActorAutoLoading === true || input.enableActorAutoLoading === 'true';
         } else {
-            input.enableAddingActors = false;
+            input.enableAddingActors = true;
```
MichalKalita, line 27: This simple update changes the logic in apify-mcp-server. When the setting is not set, it's enabled by default. If it is set (by URL or config object), it uses the value as before.

MichalKalita:
Tests are updated, and all should pass.

MQ37 (commented):
please add test case for disabling the Actor adding tools by passing `false` and check if they are actually NOT loaded, otherwise LGTM :+1:

jirispilka (approved):
Thank you!

PR #150, opened 2025-06-24
fix: optimize dataset items truncation logic

### LINKED ISSUE #121: Optimize data truncation and improve logic when data are truncated
Context:

```
                  const itemContents = items.items.map((item: Record<string, unknown>) => {
                        const text = JSON.stringify(item).slice(0, ACTOR_OUTPUT_MAX_CHARS_PER_ITEM);
                        return text.length === ACTOR_OUTPUT_MAX_CHARS_PER_ITEM
                            ? { type: 'text', text: `${text} ... ${ACTOR_OUTPUT_TRUNCATED_MESSAGE}` }
                            : { type: 'text', text };
                    });
```

MQ37:
closes https://github.com/apify/actors-mcp-server/issues/121

Before we just truncated the stringified dataset JSON, which almost always broke the JSON structure and tool result context (not all fields were present). This new logic pops dataset items one by one from the back either until the max character limit is reached or until only one item is remaining. Then, we improve the truncated info message by guiding the LLM to use the truncated count and offset in the get dataset items tool. This logic keeps the JSON structure valid and ensures that the tool/Actor results fit into the LLM context.

Note: Specific/bad Actors may have weird output schemas and can technically dump the whole output into a single huge dataset item, but I think we should not handle this case - the Actor creator should ensure user and LLM friendly output.

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -43,8 +43,11 @@ export const defaults = {
 
 // Actor output const
 export const ACTOR_OUTPUT_MAX_CHARS_PER_ITEM = 5_000;
-export const ACTOR_OUTPUT_TRUNCATED_MESSAGE = `Output was truncated because it will not fit into context.`
-    + `There is no reason to call this tool again! You can use ${HelperTools.DATASET_GET_ITEMS} tool to get more items from the dataset.`;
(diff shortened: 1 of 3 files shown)


MQ37:
@jirispilka what do you think about this logic? Tested with Claude desktop and never reached a context window limit error when making a tool call.

Issue #155, opened 2025-07-01
Add generic call-actor tool


MQ37:
So client that do not support dynamic tool discovery can call any Actor after they tried adding one. Be sure to mention that in the tool call output to guide the LLM. When running non-existent Actor mention that if the Actor was recently added maybe the client does not support dynamic tools discovery and should try to get Actor details (to get input schema) and then use the generic call-actor tool.

MQ37:
The response to guide the LLM to use this generic tool should actually be returned in the add-actor tool result since when calling non-existent tool the "request" can be aborted at the client level and not even be sent to the MCP server.

MichalKalita:
> The response to guide the LLM to use this generic tool should actually be returned in the add-actor tool result since when calling non-existent tool the "request" can be aborted at the client level and not even be sent to the MCP server.

I don't understand how it can be returned when LLM will not call 'add-actor'

It will call only:
1. search-actor
2. call-actor

It will be added dynamically when it is called. When the user writes the exact Actor name in the prompt, LLM can call it directly without using any other tool.

MQ37:
> > The response to guide the LLM to use this generic tool should actually be returned in the add-actor tool result since when calling non-existent tool the "request" can be aborted at the client level and not even be sent to the MCP server.
> 
> I don't understand how it can be returned when LLM will not call 'add-actor'
> 
> It will call only:
> 1. search-actor
> 2. call-actor
> 
> It will be used dynamically when it is called. When the user writes the exact Actor name in the prompt, LLM can call it directly without using any other tool.


The idea is that the client should at least try to add the actor and if it does not work then it can use the generic tool.

PR #156, opened 2025-07-02
feat(test): Evaluate MCP tools usage

MichalKalita:
Closes https://github.com/apify/ai-team/issues/30

It's a simple test to run Agent with a prompt and check how the tools were used.
The target is to be sure the tool description is good and LLM doesn't make mistakes with tool selection and arguments.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24,10 +24,13 @@
         "actors-mcp-server": "dist/stdio.js"
       },
       "devDependencies": {
+        "@ai-sdk/openai": "^1.3.22",
         "@anthropic-ai/sdk": "^0.33.1",
         "@anthropic-ai/tokenizer": "^0.0.4",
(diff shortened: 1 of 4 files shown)


MichalKalita:
This MCP server is not compatible with the Mastra MCP client. We allow adding tools (Actors) at runtime inside LLM requests, while Mastra only allows adding tools between requests. This may be a problem for more clients.

We have two options:
1. Implement #155 
2. Use on MCP client other than Mastra ( but it's itself a problem, when popular MCP clients don't work, it's itself an evaluation fail)

@jirispilka @MQ37

MichalKalita:
@jirispilka I'm closing this PR. We want a tool that allows us to conduct A/B tests in complex scenarios, collect all metrics, and decide which way is better.

PR #161, opened 2025-07-07
feat: call-actor tool

### LINKED ISSUE #155: Add generic call-actor tool

MichalKalita:
closes #155

Generic tool for calling Actors dynamically, for clients without support for the dynamic tools list.

TODO:
- [x] allow call only added tools
- [x] add to add-tools description info, "'call-actor' can be used when the added tool is not found"
- [x] add to this tool info "use 'add-actor' before this tool"
- [x] merge branch feat/fewer-tools--improve-tool-output and setup as target

### DIFF
--- a/src/tools/actor.ts
+++ b/src/tools/actor.ts
@@ -241,9 +241,11 @@ export async function getActorsAsTools(
         }),
     );
 
+    const clonedActors = structuredClone(actorsInfo);
+
     // Filter out nulls and separate Actors with MCP servers and normal Actors
-    const actorMCPServersInfo = actorsInfo.filter((actorInfo) => actorInfo && actorInfo.webServerMcpPath) as ActorInfo[];
-    const normalActorsInfo = actorsInfo.filter((actorInfo) => actorInfo && !actorInfo.webServerMcpPath) as ActorInfo[];
+    const actorMCPServersInfo = clonedActors.filter((actorInfo) => actorInfo && actorInfo.webServerMcpPath) as ActorInfo[];
+    const normalActorsInfo = clonedActors.filter((actorInfo) => actorInfo && !actorInfo.webServerMcpPath) as ActorInfo[];
 
     const [normalTools, mcpServerTools] = await Promise.all([
         getNormalActorsAsTools(normalActorsInfo),
@@ -315,7 +317,6 @@ export const callActor: ToolEntry = {
                 throw new Error('APIFY_TOKEN environment variable is not set.');
             }
             try {
-                // FIXME: Bug, every call add "**REQUIRED**" to the description of the input properties
                 const [actor] = await getActorsAsTools([actorName], apifyToken);
 
                 if (!actor) {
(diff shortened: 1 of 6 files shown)


src/tools/actor.ts
```diff
@@ -287,3 +289,94 @@ export const getActor: ToolEntry = {
...
+                        content: [{
+                            type: 'text',
+                            text: `Actor '${actorName}' is not added. Add it with tool '${HelperTools.ACTOR_ADD}'. ${toolsText}`,
+                        }],
+                    };
+                }
+                return {
+                    content: [{
+                        type: 'text',
+                        text: `Actor '${actorName}' is not added. ${toolsText}`,
```
MQ37: Please add help text here that the user should add the Actor to the query param ?actors= comma separated list or if using cli --actors cli arg list

src/tools/actor.ts
```diff
@@ -287,3 +289,94 @@ export const getActor: ToolEntry = {
...
+        description: `Call Actor and get dataset results. Call without input and result response with requred input properties. Actor MUST be added before calling, use ${HelperTools.ACTOR_ADD} tool before.`,
+        inputSchema: zodToJsonSchema(callActorArgs),
+        ajvValidate: ajv.compile(zodToJsonSchema(callActorArgs)),
+        call: async (toolArgs) => {
+            const { apifyMcpServer, args, apifyToken } = toolArgs;
+            const { actorName, input, callOptions } = callActorArgs.parse(args);
+
+            const actors = apifyMcpServer.listActorToolNames();
+            if (!actors.includes(actorName)) {
+                const toolsText = actors.length > 0 ? `Added Actors are: ${actors.join(', ')}` : 'Not added Actors yet.';
```
MQ37: I would personally use word "Available" instead of added Actors.

src/tools/actor.ts
```diff
@@ -287,3 +289,94 @@ export const getActor: ToolEntry = {
...
+                );
+
+                return {
+                    content: items.items.map((item: Record<string, unknown>) => ({
+                        type: 'text',
+                        text: JSON.stringify(item),
+                    })),
+                };
+            } catch (error) {
+                console.error(`Error calling Actor: ${error}`);
```
MQ37: Please use log.error from package @apify/log (see other modules using "log.error")

MQ37 (changes requested):
Thank you, just a couple of comments. Please be sure to also add integration tests to the tests/integration/suite.ts for the call-actor tool. I would also personally move the tool to a separate file tools/call-actor.ts (but it's fine for now, we should refine the strucutre later on - there is an issue for that).

MichalKalita:
> Thank you, just a couple of comments. Please be sure to also add integration tests to the tests/integration/suite.ts for the call-actor tool. I would also personally move the tool to a separate file tools/call-actor.ts (but it's fine for now, we should refine the strucutre later on - there is an issue for that).

I wrote two simple tests: run them after the Actor is added to check if they work, and run them without adding an actor.

PR #162, opened 2025-07-08
feat: apify docs tools

MQ37:
Changes:
- added Apify docs search and fetch tools
- updated README

closes https://github.com/apify/ai-team/issues/39

### DIFF
--- a/src/utils/turndown.ts
+++ b/src/utils/turndown.ts
@@ -1,0 +1,32 @@
+import TurndownService from "turndown";
+
+export const turndown = new TurndownService();
+
+// Remove non-visible elements
+turndown.remove('script');
+turndown.remove('style');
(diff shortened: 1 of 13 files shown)


src/utils/turndown.ts
```diff
@@ -0,0 +1,32 @@
+import TurndownService from 'turndown';
+
+export const turndown = new TurndownService();
```
MichalKalita: It's fragile, have a global mutable thing. Export only new function "html-to-markdown" where `turndown.turdown(input)` will be called. The rest of project doesn't need to know which library is used.

MQ37:
> @MQ37 this will work only for the remote server, right?

yes - we just use algolia API and then fetch docs.apify.com page html and convert to markdown

PR #164, opened 2025-07-09
fix: generic actor-call tool and hide it behind beta flag

MQ37:
Improve the generic call tool description and enable it only if `?beta=1` or `--beta` flag is present.


MichalKalita (approved):
looks good

PR #172, opened 2025-07-16
feat: add tool selection by feature param

MQ37:
closes https://github.com/apify/actors-mcp-server/issues/166

**Only `docs` tools category is currently enabled by default (always)**

This PR introduces the `tools` (`--tools` or `?tools`) comma separated list param so users can choose which tools are loaded into their MCP session.

Possible values:
- docs: Search and fetch Apify documentation tools.
- runs: Get Actor runs list, run details, and logs from a specific Actor run.
- storage: Access datasets, key-value stores, and their records.

Example:
```
--tools docs,runs
?tools=docs,storage
```

Also vibe refactored the tool loading logic so it's not duplicated in the `stdio.ts`

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -23,9 +23,9 @@ import {
     SERVER_NAME,
     SERVER_VERSION,
 } from '../const.js';
-import { addRemoveTools, betaTools, callActorGetDataset, defaultTools, getActorsAsTools } from '../tools/index.js';
+import { addRemoveTools, betaTools, callActorGetDataset, defaultTools, featureTools, getActorsAsTools } from '../tools/index.js';
 import { actorNameToToolName, decodeDotPropertyNames } from '../tools/utils.js';
-import type { ActorMcpTool, ActorTool, HelperTool, ToolEntry } from '../types.js';
+import type { ActorMcpTool, ActorTool, FeatureToolKey, HelperTool, ToolEntry } from '../types.js';
 import { connectMCPClient } from './client.js';
 import { EXTERNAL_TOOL_CALL_TIMEOUT_MSEC } from './const.js';
 import { processParamsGetTools } from './utils.js';
@@ -173,6 +173,13 @@ export class ActorsMcpServer {
         const actorsToLoad: string[] = [];
         const toolsToLoad: ToolEntry[] = [];
         const internalToolMap = new Map([...defaultTools, ...addRemoveTools, ...betaTools].map((tool) => [tool.tool.name, tool]));
+        // Add all feature tools
+        for (const key of Object.keys(featureTools)) {
+            const tools = featureTools[key as FeatureToolKey];
+            for (const tool of tools) {
+                internalToolMap.set(tool.tool.name, tool);
+            }
+        }
 
         for (const tool of toolNames) {
             // Skip if the tool is already loaded
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -23,9 +23,9 @@ import {
     SERVER_NAME,
     SERVER_VERSION,
 } from '../const.js';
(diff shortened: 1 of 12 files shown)


src/tools/index.ts
```diff
@@ -1,45 +1,53 @@
...
-    // getUserDatasetsList,
-    // getUserKeyValueStoresList,
     getActorDetailsTool,
-    helpTool,
     searchActors,
-    searchApifyDocsTool,
-    fetchApifyDocsTool,
 ];
 
+export const featureTools = {
```
jirispilka: I don't have strong opinion but what about toolCategories? I think it is more descriptive?

src/tools/index.ts
```diff
@@ -1,45 +1,53 @@
...
+        getDatasetItems,
+        getKeyValueStore,
+        getKeyValueStoreKeys,
+        getKeyValueStoreRecord,
+        getUserDatasetsList,
+        getUserKeyValueStoresList,
+    ],
+};
+
 export const betaTools = [
```
jirispilka: should we include betaTools into the toolsCategories? I think it would simplify the logic, why to have to switches for the same thing? If we want we can call it `preview`?

jirispilka (approved):
Looks good to me. The only change I would do is rename things and remove the beta flag. What do you think?

src/tools/index.ts
MQ37: Agree :+1:

src/tools/index.ts
MQ37: we can include that, let's call it `preview` as you mentioned

MQ37:
> Looks good to me. The only change I would do is rename things and remove the beta flag. What do you think?

Agree, renamed and moved the beta tools to `preview` tools category :+1: Thank you for review

src/stdio.ts
```diff
@@ -88,8 +102,18 @@ if (!process.env.APIFY_TOKEN) {
...
+
+    // Create an Input object from CLI arguments
+    const input: Input = {
+        actors: actorList.length ? actorList : [],
+        enableAddingActors,
+        tools: toolCategoryKeys as ToolCategory[],
+    };
+
+    // Use the shared tools loading logic
+    const tools = await loadToolsFromInput(input, process.env.APIFY_TOKEN as string, actorList.length === 0);
```
MichalKalita, line 115: Missing validation if tools list is only list of available options.

src/mcp/server.ts
```diff
@@ -172,7 +166,14 @@ export class ActorsMcpServer {
         const loadedTools = this.listAllToolNames();
         const actorsToLoad: string[] = [];
         const toolsToLoad: ToolEntry[] = [];
-        const internalToolMap = new Map([...defaultTools, ...addRemoveTools, ...betaTools].map((tool) => [tool.tool.name, tool]));
+        const internalToolMap = new Map([...defaultTools, ...addRemoveTools].map((tool) => [tool.tool.name, tool]));
+        // Add all category tools
+        for (const key of Object.keys(toolCategories)) {
```
MichalKalita: `Object.values(toolCategories).flat()` will be better here. Key is used only for get value from object.

MichalKalita (changes requested):

src/stdio.ts
MQ37: When the user inputs invalid option it just ignores it. What is goal of validation here? Do you think we should refuse to start the MCP server and exit with error code when the option is not valid?

src/mcp/server.ts
MQ37: Agree, thank you :+1:

src/stdio.ts
MichalKalita: OK, we can ignore it. There is no good way to send LLM information about the wrong MCP server being used, so it's easier to ignore it.

MQ37: we agreed to not solve this as of now and ignore bad options

PR #173, opened 2025-07-16
feat: progress notification

### LINKED ISSUE #149: Send progress notification to prevent timeout
If the server sends progress, it will renew the timeout in the client. The inspector shows a default timeout of 10s and a total timeout of 60s. 

Image

The server should send some progress info to prevent a timeout on the client side.

https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/progress#progress

MichalKalita:
Closes #149

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -430,10 +430,6 @@ export class ActorsMcpServer {
                         ? createProgressTracker(progressToken, extra.sendNotification)
                         : null;
 
-                    if (progressTracker) {
-                        await progressTracker.updateProgress(0, `Starting: ${internalTool.name}`);
-                    }
-
                     const res = await internalTool.call({
                         args,
                         extra,
@@ -444,7 +440,7 @@ export class ActorsMcpServer {
                     }) as object;
 
                     if (progressTracker) {
-                        await progressTracker.complete(`Completed: ${internalTool.name}`);
+                        progressTracker.stop();
                     }
 
                     return { ...res };
@@ -506,10 +502,6 @@ export class ActorsMcpServer {
                             progressTracker,
                         );
 
-                        if (progressTracker) {
-                            await progressTracker.complete(`Completed: ${actorTool.actorFullName}`);
-                        }
-
                         return {
                             content: items.items.map((item: Record<string, unknown>) => {
                                 return {
@@ -520,7 +512,7 @@ export class ActorsMcpServer {
                         };
                     } finally {
                         if (progressTracker) {
-                            progressTracker.stopPeriodicUpdates();
+                            progressTracker.stop();
                         }
                     }
                 }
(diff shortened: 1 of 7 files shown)


src/utils/progress.ts
```diff
@@ -0,0 +1,105 @@
...
+
+                    // Stop polling if actor finished
+                    if (status === 'SUCCEEDED' || status === 'FAILED' || status === 'ABORTED' || status === 'TIMED-OUT') {
+                        this.stopPeriodicUpdates();
+                    }
+                }
+            } catch {
+                // Silent fail - continue polling
+            }
+        }, 5_000);
```
MQ37: please move the hardcoded value to some const in `consts.ts`

src/mcp/server.ts
```diff
@@ -425,6 +426,16 @@ export class ActorsMcpServer {
                 // Handle internal tool
                 if (tool.type === 'internal') {
                     const internalTool = tool.tool as HelperTool;
+
+                    // Only create progress tracker for call-actor tool
+                    const progressTracker = internalTool.name === 'call-actor'
+                        ? createProgressTracker(progressToken, extra.sendNotification)
+                        : null;
+
+                    if (progressTracker) {
```
MQ37: do you actually pass the `progressTracker` to the `call-actor` internal tool?

src/utils/progress.ts
```diff
@@ -0,0 +1,105 @@
...
+                const { status, statusMessage } = run;
+
+                // Only send notification if status or statusMessage changed
+                if (status !== lastStatus || statusMessage !== lastStatusMessage) {
+                    lastStatus = status;
+                    lastStatusMessage = statusMessage || '';
+
+                    // Calculate progress based on status
+                    let progress = 0;
+                    if (status === 'RUNNING') progress = 50;
```
MQ37: based on the spec the progress value must always be increasing https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/progress#progress. We need to change this logic

src/mcp/server.ts
MichalKalita: Only for call-actor, the progress tracker is started.

MQ37: where exactly do you pass that to the `call-actor`? Either `internalTool.call()` func call does not receive the `progressTracker` or I am just blind :smile:

MQ37: since in the call Actor get Dataset function call you pass the `progressTracker` and it handles the actual Actor run and the status message

src/utils/progress.ts
MichalKalita: OK, i updated it. And I removed total value, it's not required.

jirispilka (approved):
Thanks! Looks nice. Once @MQ37 reviews it, we can merge it and test on staging together with other changes.

And we can do a release today.

src/utils/progress.ts
MQ37: nice, this adheres to the spec :+1:

MQ37 (approved):
LGTM :+1:

src/mcp/server.ts
MichalKalita: Updated, progressTracker is send as parameter.

PR #176, opened 2025-07-18
feat: Update README, make it shorter and concise

jirispilka:
I've decided to remove any content that isn't strictly necessary.

The README had become too lengthy, making it easy to lose track of important information.

My initial goal was to have a single source of truth at docs.apify.com/.../mcp, but I wasn’t brave enough to remove everything from the GitHub repository just yet.

As a result, the current state is a compromise. However, moving forward, we should aim to maintain only one authoritative page for MCP settings and configuration. This will ensure updates are easy and we won’t need to propagate changes across multiple repositories.

close https://github.com/apify/ai-team/issues/21


MQ37 (approved):
Nice, I like this :+1:

MichalKalita (approved):

PR #179, opened 2025-07-22
fix: export internals for apify-mcp-server

MQ37:
Export internals used in `apify-mcp-server` repo to prevent importing from `dist/`.

related to https://github.com/apify/apify-mcp-server/issues/102

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,11 @@
     "node": ">=18.0.0"
   },
   "main": "dist/index.js",
+  "exports": {
+    ".": "./dist/index.js",
+    "./internals": "./dist/index-internals.js",
+    "./internals.js": "./dist/index-internals.js"
+  },
   "bin": {
     "actors-mcp-server": "./dist/stdio.js"
   },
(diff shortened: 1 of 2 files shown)


jirispilka (changes requested):
I still think it would be better to expose only the `ActorsMcpServer` class.
Otherwise, it becomes too easy to break something.

Do we really need to export all these functions?
Could we include them as part of the `ActorsMcpServer` class instead?

MQ37:
> I still think it would be better to expose only the `ActorsMcpServer` class. Otherwise, it becomes too easy to break something.
> 
> Do we really need to export all these functions? Could we include them as part of the `ActorsMcpServer` class instead?

We can include the `actorNameToToolName` and `parseInputParamsFromUrl` functions in the `ActorsMcpServer` class, however I don't think it makes sense to include tool related "config", like categories, in this class. But I think in a long run we should get rid of this class to make the architecture more scalable and less isolated in the class.

@MichalKalita what do you think?

MichalKalita:
Fun fact, `react` uses more dangerous names. https://github.com/facebook/react/blob/edac0dded99d56e7d66a88da83e874761e3e937a/fixtures/fiber-debugger/src/App.js#L10

jirispilka (approved):
For the record, exporting more tools will just make things harder for us.
Once something is exposed, we shouldn’t change it, or it might break in apify-mcp.

Let's approve it and we will see.

PR #190, opened 2025-07-24
feat: Add tool to get dataset schema so LLMs can understand dataset structure without fetching everything

jirispilka:
feat: Add tool to get dataset schema so LLMs can understand dataset structure without fetching everything (mainly for Zuzka)

Placed under storage since that group isn’t enabled by default.

Also, I added runId and datasetId injection into context, hopefully it won't break anything affect existing flows.


src/tools/dataset.ts
```diff
@@ -112,3 +113,132 @@ export const getDatasetItems: ToolEntry = {
...
+                const datasetItems = datasetResponse.items;
+
+                if (datasetItems.length === 0) {
+                    return { content: [{ type: 'text', text: `Dataset '${parsed.datasetId}' is empty.` }] };
+                }
+
+                // Clean the dataset items by removing empty arrays
+                const cleanedDatasetItems = datasetItems.map((item) => removeEmptyArrays(item));
+
+                try {
```
MQ37: I do not like the nested try blocks

MQ37 (approved):
LGTM, I just do not like the nested try blocks, but it works so it does not matter.

MichalKalita (approved):
👍 for fallbackSchema approach

PR #194, opened 2025-07-29
docs: tools overview table

MQ37:
merge after https://github.com/apify/apify-mcp-server/pull/109

related to https://github.com/apify/actors-mcp-server/issues/185

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ For example, to enable all tools, use `npx @apify/actors-mcp-server --tools docs
 
 ### Overview of available tools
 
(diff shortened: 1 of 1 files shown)


jirispilka:
I think we've discussed including some general description only with a link to https://mcp.apify.com/tools

MQ37:
> I think we've discussed including some general description only with a link to https://mcp.apify.com/tools, but I'm fine with this too

I decided to include a simplified version of the table in the README.md so users can see it right away in the repo. If that becomes a hassle to maintain, we can replace it with just a link.

MQ37:
After today's sync I removed the link to the `mcp.apify.com/tools` as this will be removed and replaced with a `?dumpJson=1`. I think this should still be merged to provide an overview of all available tools and what tools are enabled by default. @jirispilka

Issue #181, opened 2025-07-23
Present search and Actor details in human readable format


jirispilka:
For the `search-actors` and `get-actor-details` tools, we shouldn’t just return raw JSON—we should return properly formatted Markdown.

The Markdown should include all the information visible on the Actor page (see example: [https://apify.com/apify/rag-web-browser](https://apify.com/apify/rag-web-browser)):

* [] **search-actors (Markdown):** title, URL, image, description, usage, pricing, etc.
* [] **actor details (Markdown):** the same set of information

**Goal:**
Present consistent and visually appealing information.

**Rationale:**
When returning JSON, the LLM sometimes omits important fields like the URL or image. Returning Markdown ensures all relevant details are shown clearly and consistently.

jirispilka:
```
Here are the top 10 Actors matching "Twitter scraper":

# [Twitter Scraper](https://apify.com/apidojo/tweet-scraper) (`apidojo/tweet-scraper`)
- Developer: [APIDojo]([https://apify.com/apidojo](https://apify.com/apidojo/tweet-scraper)) (by community / by Apify)
- Description: ....
- Categories: Social media, Lead generation
- User rating: 3.8 of 5 stars
- Pricing: This Actor charges per results (in this case named tweet) ...
- Stats: 18.2k total users, 10k active users, 771 bookmarks
```

PR #195, opened 2025-07-29
feat: Return Actor card in markdown

jirispilka:
- **fix: standardize tool description formatting with string concatenation**
- **feat: Return markdown from get-actor-details.ts**
- **feat: Return markdown search-actors tools**
- **feat: Improve search tool description**

close: #181

### DIFF
--- a/src/utils/actor-card.ts
+++ b/src/utils/actor-card.ts
@@ -62,6 +62,18 @@ export function formatActorToActorCard(
             statsParts.push(`${stats.totalUsers.toLocaleString()} total users, ${stats.totalUsers30Days.toLocaleString()} monthly users`);
         }
 
+        // Add success rate for last 30 days if available
+        if ('publicActorRunStats30Days' in stats && stats.publicActorRunStats30Days) {
+            const runStats = stats.publicActorRunStats30Days as {
+                SUCCEEDED: number;
+                TOTAL: number;
+            };
+            if (runStats.TOTAL > 0) {
+                const successRate = ((runStats.SUCCEEDED / runStats.TOTAL) * 100).toFixed(1);
+                statsParts.push(`Runs succeeded: ${successRate}%`);
+            }
+        }
+
         // Add bookmark count if available (ActorStoreList only)
         if ('bookmarkCount' in actor && actor.bookmarkCount) {
             statsParts.push(`${actor.bookmarkCount} bookmarks`);
--- a/src/types.ts
+++ b/src/types.ts
@@ -157,13 +157,6 @@ export interface ToolEntry {
     tool: ActorTool | HelperTool | ActorMcpTool;
 }
 
-//  ActorStoreList for actor-search tool
-export interface ActorStats {
-    totalRuns: number;
-    totalUsers30Days: number;
-    publicActorRunStats30Days: unknown;
-}
-
 /**
  * Price for a single event in a specific tier.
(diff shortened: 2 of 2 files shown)


jirispilka:
image
image

src/utils/actor-card.ts
```diff
@@ -0,0 +1,105 @@
...
+        return formatted.replace('Mcp Server', 'MCP Server').replace('Ai', 'AI').replace('Seo', 'SEO');
+    });
+}
+
+/**
+ * Formats Actor details into an Actor card (Actor markdown representation).
+ * @param actor - Actor information from the API
+ * @returns Formatted actor card
+ */
+export function formatActorToActorCard(
```
MQ37, line 27: I think we should include the number of failed runs that was in the original search function.

src/tools/run_collection.ts
```diff
@@ -32,9 +32,9 @@ export const getUserRunsList: ToolEntry = {
     tool: {
         name: HelperTools.ACTOR_RUN_LIST_GET,
         actorFullName: HelperTools.ACTOR_RUN_LIST_GET,
-        description: `Gets a paginated list of Actor runs with run details, datasetId, and keyValueStoreId.
-           Filter by status: READY (not allocated), RUNNING (executing), SUCCEEDED (finished), FAILED (failed),
-           TIMING-OUT (timing out), TIMED-OUT (timed out), ABORTING (being aborted), ABORTED (aborted).`,
+        description: `Gets a paginated list of Actor runs with run details, datasetId, and keyValueStoreId.\n`
```
MQ37, line 35: why do we use the `+` to join the strings if we can just use single multiline `\`` string?

MQ37 (changes requested):
I would add back the failed number of runs stats that was in the original implementation. Otherwise LGTM :+1:

src/tools/run_collection.ts
jirispilka: Other tools already used that. I find noindent ugly. I'm sorry. If we agree on using template literals with noindent  across the codebase, I'll adapt. 

So what do you say @MQ37 @MichalKalita

src/utils/actor-card.ts
jirispilka: Yeah, I haven’t explained this. Good catch. 

 We should only display information available on the Actor page. There is Actor "succeeded", which I personally find misleading, for example, mcp-tester-client shows 0% because it never finishes. 
 
 I agree this wasn’t ideal, so I added "run-succeeded" instead.

src/tools/run_collection.ts
MQ37: I also like how the `+` looks but updating that and splitting the lines is a hassle. Let's keep that for now :+1:

MQ37 (approved):
LGTM :+1:

src/tools/run_collection.ts
MichalKalita: I usually prefer ``` because of simplicity. It's easier to copy/paste content.
But there is no need to refactor all places and update them.

MichalKalita (approved):

src/tools/run_collection.ts
jirispilka: Ok, I've created issue for that: https://github.com/apify/actors-mcp-server/issues/197

PR #196, opened 2025-07-30
feat: handle Apify specific input schema props

MQ37:
Added helpers to inject Apify-specific schema structures for:
- resourcePicker arrays
- keyValue arrays
- globs arrays
- pseudoUrls arrays
- proxy objects (additional Apify proxy fields)
- Request list source (additional http request fields)

I believe this **should** be all Apify specific fields that require is to build their inner nested schemas from https://github.com/apify/apify-shared-js/blob/master/packages/input_schema/src/schema.json.

closes https://github.com/apify/actors-mcp-server/issues/182


jirispilka (approved):
Really like how the code was simplified, great job!

Moving things into apify-properties.ts made actor.ts and the whole structure much cleaner.

One question: do you think apify-properties.ts belongs in the tool directory? 

I always thought that folder was mainly for tool specs. Maybe it could go in utils.ts (though that file can get pretty big). 

Not sure what’s best, just wanted to mention it in case you have a better idea.

src/tools/utils.ts
```diff
@@ -279,3 +278,16 @@ export function decodeDotPropertyNames(properties: Record<string, unknown>): Rec
     }
     return decodedProperties;
 }
+
+export function transformActorInputSchemaProperties(input: IActorInputSchema): ActorInputSchemaProperties {
+    // Deep clone input to avoid mutating the original object
+    const inputClone: IActorInputSchema = JSON.parse(JSON.stringify(input));
```
MichalKalita: use structuredClone function instead 
https://developer.mozilla.org/en-US/docs/Web/API/Window/structuredClone

tests/unit/tools.utils.test.ts
```diff
@@ -1,12 +1,136 @@
...
+        expect(result.pseudoUrls.items?.properties?.purl).toBeDefined();
+        expect(result.pseudoUrls.items?.properties?.purl.type).toBe('string');
+        expect(result.pseudoUrls.items?.properties?.method).toBeDefined();
+        expect(result.pseudoUrls.items?.properties?.method.type).toBe('string');
+        expect(result.pseudoUrls.items?.properties?.payload).toBeDefined();
+        expect(result.pseudoUrls.items?.properties?.payload.type).toBe('string');
+        expect(result.pseudoUrls.items?.properties?.userData).toBeDefined();
+        expect(result.pseudoUrls.items?.properties?.userData.type).toBe('object');
+        expect(result.pseudoUrls.items?.properties?.headers).toBeDefined();
+        expect(result.pseudoUrls.items?.properties?.headers.type).toBe('object');
```
MichalKalita, line 128: It can be easier with just `expect(result.pseudoUrls.items).toMatchInlineSnapshot()`
When we can always expect same values

tests/unit/tools.utils.test.ts
```diff
@@ -368,3 +518,272 @@ describe('decodeDotPropertyNames', () => {
...
+            },
+        };
+        const result = transformActorInputSchemaProperties(input);
+        expect(result['a-dot-b']).toBeDefined();
+        expect(result['c-dot-d-dot-e']).toBeDefined();
+        expect(result['a.b']).toBeUndefined();
+        expect(result['c.d.e']).toBeUndefined();
+    });
+
+    it('should not mutate the input object', () => {
```
MichalKalita, line 776: Function `transformActorInputSchemaProperties` should have `Readonly<IActorInputSchema>` input when mutation is disabled.
If not mutated input is really important, we can keep this test.

MichalKalita (changes requested):

tests/unit/tools.utils.test.ts
MQ37: no mutating the input is actually important since in a few tests I discovered (and also you mentioned that I think) that it can "break" things (duplicate descriptions, `**required**` and so on)

tests/unit/tools.utils.test.ts
MQ37: For now I would keep this and we can rework that later if needed

PR #202, opened 2025-08-04
fix: improve beta release cicd

MQ37:
This PR improves the pre-release CI/CD workflow. It now only runs when the PR has a beta tag (including when the tag is added without a new push). The calculation of the next beta release version is now more robust: it checks both NPM and Git tags for existing beta versions and uses the highest found, ensuring the next beta version is always unique and up-to-date.

closes https://github.com/apify/actors-mcp-server/issues/80

Notes:
> it seems that there is not way to run a CICD job when a tag is removed

> another alternative solution is to use github branches for beta releases that we can later remove (NPM can install from github). But for simplicity I would keep the NPM implementation

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,12 @@ npx @modelcontextprotocol/inspector node ./dist/stdio.js
 
 Upon launching, the Inspector will display a URL that you can open in your browser to begin debugging.
 
+## 🐦 Canary PR releases
+
+Due to the current architecture where Apify MCP is split across two repositories, this one containing the core MCP logic and the private [apify-mcp-server](https://github.com/apify/apify-mcp-server) repository that handles the actual server implementation for [mcp.apify.com](https://mcp.apify.com), development can be challenging as changes need to be synchronized between both repositories.
+
+You can create a canary release from your PR branch by adding the `beta` tag. This will test the code and publish the package to [pkg.pr.new](https://pkg.pr.new/) which you can then use, for example, in a staging environment to test before actually merging the changes. This way we do not need to create new NPM releases and keep the NPM versions cleaner. The workflow runs whenever you commit to a PR branch that has the `beta` tag or when you add the `beta` tag to an already existing PR. For more details check out [the workflow file](.github/workflows/pre_release.yaml).
+
 # 🐛 Troubleshooting (local MCP server)
 
 - Make sure you have `node` installed by running `node -v`.


MQ37:
> IMO, this has a lot of custom logic, and I’m not sure we really need to add it here. I think we should aim for as unified a deployment as possible across apify packages

I checked few Apify repose and it seems that this beta release workflow is not centralized and each repo handles that separately. I think developer flexibility is important and thus would also keep this decentralized, per repo.

These repos use **a variant** of `before-beta-release.js`:
- https://github.com/apify/apify-client-js/blob/master/.github/scripts/before-beta-release.js
- https://github.com/apify/apify-cli/blob/master/.github/scripts/before-beta-release.ts

These repos use on push to master model:
- https://github.com/apify/apify-sdk-js/blob/master/.github/workflows/test-and-release.yaml
- https://github.com/apify/crawlee/blob/master/.github/workflows/test-ci.yml

MichalKalita (approved):
I don't fully understand GitHub yaml configs, but JS part looks good.

B4nan:
So what exactly are you trying to solve here? Do you need canary builds for PRs, or for stuff in master? Normally, we just run the canary builds for every commit to master, making this dependent on reopening a merge PR feels quite weird to me (that's what this PR is doing, right?)

> The calculation of the next beta release version is now more robust: it checks both NPM and Git tags for existing beta versions and uses the highest found

This doesn't sound right to me, first of all, I really dislike the idea of git tags for every commit to master, I'd just drop those - and base the next vesion only on NPM, that is the single source of truth to me. Also, I would prefer checking the version from NPM API instead of package.json

B4nan:
This is what we use in crawlee for detection of the next canary version number:

https://github.com/apify/crawlee/blob/e4b7f69bca6113a4622738ffd92cf8d5dfe44819/scripts/copy.ts#L68-L99

MQ37:
> So what exactly are you trying to solve here? 

The Apify MCP is spread across two repos, this one and the `apify-mcp-server`, private one that uses the core logic from this repo. For some PRs that change specific internals we would like to publish our beta/canary release to the NPM so we can test that on staging `mpc.apify.com` (using the private `apify-mcp-server` repo) before merging PR in this repo. So I don't think it's right to automatically release canary version for each PR that does not need that.

I included the git tag logic so there is a second source of truth, since one time the beta release script cause this:

> last time for some reason the beta version jumped straight from stable [0.2.15](https://www.npmjs.com/package/@apify/actors-mcp-server/v/0.2.15) to [0.3.1-beta.0](https://www.npmjs.com/package/@apify/actors-mcp-server/v/0.3.1-beta.0) for some reason 🤷‍♂️

We can remove the git tag logic and add more logging so we can then trace what causes this version jump if that happens again for some reason.

jirispilka:
Thanks, guys. IMO, https://pkg.pr.new/ looks like exactly what we need to create an ad-hoc release for testing

B4nan:
Great to hear that, we'll probably adopt it too in other projects if it worked well for you guys.

jirispilka (changes requested):
Really cool that it’s working!

I think it deserves a bit of documentation in the README. If it’s something others might adopt, it would be great to briefly document why we’re using it (what problem it solves) and how it works.

MQ37:
> Really cool that it’s working!
> 
> I think it deserves a bit of documentation in the README. If it’s something others might adopt, it would be great to briefly document why we’re using it (what problem it solves) and how it works.

Makes sense, documented that in "Canary releases" section of "Development" README main heading :+1:

PR #210, opened 2025-08-06
Feat/subset of input schema

MichalKalita:
closes https://github.com/apify/actors-mcp-server/issues/184

### DIFF
--- a/src/tools/actor.ts
+++ b/src/tools/actor.ts
@@ -210,25 +210,20 @@ export async function getActorsAsTools(
 
     const actorsInfo: (ActorInfo | null)[] = await Promise.all(
         actorIdsOrNames.map(async (actorIdOrName) => {
-            const actorDefinitionPrunedCached = actorDefinitionPrunedCache.get(actorIdOrName);
-            if (actorDefinitionPrunedCached) {
-                return {
-                    actorDefinitionPruned: actorDefinitionPrunedCached,
-                    webServerMcpPath: getActorMCPServerPath(actorDefinitionPrunedCached),
-
-                } as ActorInfo;
+            let actorDefinition = actorDefinitionPrunedCache.get(actorIdOrName);
+            if (!actorDefinition) {
+                actorDefinition = await getActorDefinition(actorIdOrName, apifyToken);
             }
 
-            const actorDefinitionPruned = await getActorDefinition(actorIdOrName, apifyToken);
-            if (!actorDefinitionPruned) {
-                log.error('Actor not found or definition is not available', { actorIdOrName });
+            if (!actorDefinition) {
                 return null;
             }
+
             // Cache the pruned Actor definition
-            actorDefinitionPrunedCache.set(actorIdOrName, actorDefinitionPruned);
+            actorDefinitionPrunedCache.set(actorIdOrName, actorDefinition);
             return {
-                actorDefinitionPruned,
-                webServerMcpPath: getActorMCPServerPath(actorDefinitionPruned),
+                actorDefinitionPruned: actorDefinition,
+                webServerMcpPath: getActorMCPServerPath(actorDefinition),
             } as ActorInfo;
         }),
     );
@@ -263,7 +258,6 @@ export const callActor: ToolEntry = {
     type: 'internal',
     tool: {
         name: HelperTools.ACTOR_CALL,
-        actorFullName: HelperTools.ACTOR_CALL,
         description: `Call an Actor and get the Actor run results. If you are not sure about the Actor input, you MUST get the Actor details first, which also returns the input schema using ${HelperTools.ACTOR_GET_DETAILS}. The Actor MUST be added before calling; use the ${HelperTools.ACTOR_ADD} tool first. By default, the Apify MCP server makes newly added Actors available as tools for calling. Use this tool ONLY if you cannot call the newly added tool directly, and NEVER call this tool before first trying to call the tool directly. For example, when you add an Actor "apify/instagram-scraper" using the ${HelperTools.ACTOR_ADD} tool, the Apify MCP server will add a new tool ${actorNameToToolName('apify/instagram-scraper')} that you can call directly. If calling this tool does not work, then and ONLY then MAY you use this tool as a backup.`,
         inputSchema: zodToJsonSchema(callActorArgs),
         ajvValidate: ajv.compile(zodToJsonSchema(callActorArgs)),
--- a/src/tools/actor.ts
+++ b/src/tools/actor.ts
@@ -10,6 +10,7 @@ import { ApifyClient } from '../apify-client.js';
 import {
     ACTOR_ADDITIONAL_INSTRUCTIONS,
     ACTOR_MAX_MEMORY_MBYTES,
+    ADVANCED_INPUT_KEY,
     HelperTools,
 } from '../const.js';
 import { getActorMCPServerPath, getActorMCPServerURL } from '../mcp/actors.js';
@@ -46,7 +47,7 @@ export type CallActorGetDatasetResult = {
  *
  * @param {string} actorName - The name of the actor to call.
  * @param {ActorCallOptions} callOptions - The options to pass to the actor.
- * @param {unknown} input - The input to pass to the actor.
+ * @param {ActorInput} simplifiedInput - The input to pass to the actor.
  * @param {string} apifyToken - The Apify token to use for authentication.
  * @param {ProgressTracker} progressTracker - Optional progress tracker for real-time updates.
  * @returns {Promise<{ actorRun: any, items: object[] }>} - A promise that resolves to an object containing the actor run and dataset items.
@@ -54,11 +55,17 @@ export type CallActorGetDatasetResult = {
  */
 export async function callActorGetDataset(
     actorName: string,
-    input: unknown,
+    simplifiedInput: Record<string, unknown>,
     apifyToken: string,
     callOptions: ActorCallOptions | undefined = undefined,
     progressTracker?: ProgressTracker | null,
 ): Promise<CallActorGetDatasetResult> {
+    let input = simplifiedInput;
+    if (input[ADVANCED_INPUT_KEY]) {
+        const { [ADVANCED_INPUT_KEY]: advancedInput, ...rest } = input;
+        input = { ...advancedInput, ...rest };
+    }
+
     try {
         log.info(`Calling Actor ${actorName} with input: ${JSON.stringify(input)}`);
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -29,6 +29,8 @@ export interface ISchemaProperties {
 
     sectionCaption?: string; // Optional section caption for grouping properties
     sectionDescription?: string; // Optional section description for grouping properties
+
+    additionalProperties?: boolean;
 }
 
 export interface IActorInputSchema {
--- a/src/types.ts
+++ b/src/types.ts
@@ -26,6 +26,9 @@ export interface ISchemaProperties {
 
     properties?: Record<string, ISchemaProperties>;
     required?: string[];
+
+    sectionCaption?: string; // Optional section caption for grouping properties
+    sectionDescription?: string; // Optional section description for grouping properties
 }
 
 export interface IActorInputSchema {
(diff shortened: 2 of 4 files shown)


src/types.ts
```diff
@@ -142,20 +147,27 @@ export interface ActorMcpTool extends ToolBase {
...
- * Used to store and manage tools of different types uniformly.
+ * Type discriminator for tools - indicates whether a tool is internal or Actor-based.
  */
-export interface ToolEntry {
-    /** Type of the tool (internal or actor) */
-    type: ToolType;
-    /** The tool instance */
-    tool: ActorTool | HelperTool | ActorMcpTool;
-}
+export type ToolType = ToolEntry['type'];
```
jirispilka: ToolType is not used anywhere. I think we should remove it

src/tools/actor.ts
```diff
@@ -46,19 +47,25 @@ export type CallActorGetDatasetResult = {
  *
  * @param {string} actorName - The name of the actor to call.
  * @param {ActorCallOptions} callOptions - The options to pass to the actor.
- * @param {unknown} input - The input to pass to the actor.
+ * @param {ActorInput} simplifiedInput - The input to pass to the actor.
```
jirispilka: ActorIInput is not present in the input arguments. It should be simplefied input?

jirispilka:
@MichalKalita please ask for review only once it is ready for a review.

For example, I'm getting bunch of errors when running `npm run build`

```
npm run build                   

> @apify/actors-mcp-server@0.3.2 build
> tsc -b src

src/main.ts:59:68 - error TS2345: Argument of type '{}' is not assignable to parameter of type 'Record<string, unknown>'.
  Index signature for type 'string' is missing in type '{}'.

59     const { items } = await callActorGetDataset(input.debugActor!, input.debugActorInput!, process.env.APIFY_TOKEN, options);
                                                                      ~~~~~~~~~~~~~~~~~~~~~~

src/tools/actor.ts:13:5 - error TS2305: Module '"../const.js"' has no exported member 'ADVANCED_INPUT_KEY'.

13     ADVANCED_INPUT_KEY,
       ~~~~~~~~~~~~~~~~~~

src/tools/actor.ts:66:19 - error TS2698: Spread types may only be created from object types.

66         input = { ...advancedInput, ...rest };
                     ~~~~~~~~~~~~~~~~

src/tools/build.ts:9:35 - error TS2305: Module '"../const.js"' has no exported member 'ADVANCED_INPUT_KEY'.

9 import { ACTOR_README_MAX_LENGTH, ADVANCED_INPUT_KEY, HelperTools } from '../const.js';
                                    ~~~~~~~~~~~~~~~~~~

src/utils/tools.ts:1:10 - error TS2305: Module '"../const.js"' has no exported member 'ADVANCED_INPUT_KEY'.

1 import { ADVANCED_INPUT_KEY } from '../const.js';
           ~~~~~~~~~~~~~~~~~~

````

File: tests/integration/suite.ts [hunk 3]
```diff
@@ -110,20 +110,33 @@ export function createIntegrationTestsSuite(
...
         it('should list all default tools and two loaded Actors', async () => {
             const actors = ['apify/website-content-crawler', 'apify/instagram-scraper'];
             const client = await createClientFn({ actors, enableAddingActors: false });
             const names = getToolNames(await client.listTools());
-            expect(names.length).toEqual(defaultTools.length + actors.length);
-            expectToolNamesToContain(names, DEFAULT_TOOL_NAMES);
-            expectToolNamesToContain(names, actors.map((actor) => actorNameToToolName(actor)));
+            expect(names).toMatchInlineSnapshot(`
+              [
+                "get-actor-details",
```
MQ37, line 132: If we change the tool names we need to update the test, this is fragile, I prefer the previous logic - it was not nice but was definitely more robust.

src/tools/actor.ts
```diff
@@ -46,19 +48,25 @@ export type CallActorGetDatasetResult = {
...
 export async function callActorGetDataset(
     actorName: string,
-    input: unknown,
+    simplifiedInput: Record<string, unknown>,
     apifyToken: string,
     callOptions: ActorCallOptions | undefined = undefined,
     progressTracker?: ProgressTracker | null,
 ): Promise<CallActorGetDatasetResult> {
+    let input = simplifiedInput;
+    if (input[ADVANCED_INPUT_KEY]) {
```
MQ37, line 65:
```suggestion
    // Moves the advances input properties into the root object
    if (input[ADVANCED_INPUT_KEY]) {
```

src/tools/actor.ts
```diff
@@ -202,29 +214,34 @@ async function getMCPServersAsTools(
...
             const actorDefinitionPruned = await getActorDefinition(actorIdOrName, apifyToken);
             if (!actorDefinitionPruned) {
                 log.error('Actor not found or definition is not available', { actorName: actorIdOrName });
                 return null;
             }
-            // Cache the pruned Actor definition
-            actorDefinitionPrunedCache.set(actorIdOrName, actorDefinitionPruned);
+
+            // Cache canonical full schema without mutation
+            actorDefinitionPrunedCache.set(cacheKey, actorDefinitionPruned);
```
MQ37: Why this rework? This implementation always sets the cache key and thus is resetting the TTL so the cache will never expire and if there is new Actor build we will never re-fetch it. Is my reasoning correct or did I miss something?

src/utils/tools.ts
```diff
@@ -8,6 +9,24 @@ export function getToolPublicFieldOnly(tool: ToolBase) {
     return {
         name: tool.name,
         description: tool.description,
-        inputSchema: tool.inputSchema,
+        inputSchema: simplifiedSchema(tool.inputSchema as IActorInputSchema),
     };
 }
+
+function simplifiedSchema(schema: IActorInputSchema): IActorInputSchema {
```
MQ37: Please provide dostring of what this function does.

src/stdio.ts
```diff
@@ -79,6 +80,11 @@ Note: Tools that enable you to search Actors from the Apify Store and get their
 `,
         example: 'docs,runs,storage',
     })
+    .option('full-actor-schema', {
+        type: 'boolean',
+        default: false,
+        describe: 'Enable full Actor schema for all Actors.',
```
MQ37: Please provide detailed explanation what this does. That by default it allows you to provide only basic input parameters to the Actor and you need to enable this if you need more advanced parameters.

tests/integration/suite.ts
```diff
@@ -486,6 +499,37 @@ export function createIntegrationTestsSuite(
             await client.close();
         });
 
+        it(`should make ${ADVANCED_INPUT_KEY} in Actor properties available`, async () => {
+            const client = await createClientFn({ enableAddingActors: true, fullActorSchema: false });
+            await client.callTool({ name: HelperTools.ACTOR_ADD, arguments: { actor: 'compass/crawler-google-places' } });
+            // Get input type for actor 'compass-slash-crawler-google-places'
+            const tools = await client.listTools();
+            const googlePlacesTool = tools.tools.find((tool) => tool.name === 'compass-slash-crawler-google-places');
```
MQ37: Here please use the Actor name (`compass/crawler-google-places`) and use the `actorNameToToolName` util func

tests/integration/suite.ts
```diff
@@ -486,6 +499,37 @@ export function createIntegrationTestsSuite(
             await client.close();
         });
 
+        it(`should make ${ADVANCED_INPUT_KEY} in Actor properties available`, async () => {
+            const client = await createClientFn({ enableAddingActors: true, fullActorSchema: false });
+            await client.callTool({ name: HelperTools.ACTOR_ADD, arguments: { actor: 'compass/crawler-google-places' } });
```
MQ37, line 504: same here as below please

MQ37 (changes requested):
Went through it and I like the refactored and simplified parts - mainly the ToolEntry type. There is couple of nits and one important question regarding the caching that needs clarification before merging.

Note: Please also be sure to document this in docs.apify.com and this repo README.md.

src/tools/actor.ts
jirispilka: This issue is still there - rename `ActorInput` to `simplifiedInput`

src/tools/actor.ts
```diff
@@ -119,6 +127,7 @@ export async function callActorGetDataset(
  */
 export async function getNormalActorsAsTools(
     actorsInfo: ActorInfo[],
+    fullActorSchema: boolean,
```
jirispilka, line 131: fullActorSchema is missing in the doc

src/tools/actor.ts
jirispilka: Good catch. actorDefinition is always fetched, regardless of cache

src/tools/utils.ts
```diff
@@ -279,9 +279,50 @@ export function decodeDotPropertyNames(properties: Record<string, unknown>): Rec
...
+
+    const mainInputs = propertiesEntries.slice(0, firstSectionCaptionIndex);
+    const advancedInputs = propertiesEntries.slice(firstSectionCaptionIndex);
+
+    const propObject: Record<string, ISchemaProperties> = Object.fromEntries(mainInputs);
+    // Use a basic object container here; additional details (like title/description) are not required at this stage
+    propObject.advancedInput = {
+        type: 'object',
+        title: 'Advanced Inputs',
+        description: 'These inputs are considered advanced and are not required for basic functionality.',
```
jirispilka: I think the description could be more detailed—specifically, explaining why we separate basic and advanced. It should be described in a way that makes sense to a human reader.

src/tools/utils.ts
```diff
@@ -279,9 +279,50 @@ export function decodeDotPropertyNames(properties: Record<string, unknown>): Rec
...
+        properties: Object.fromEntries(advancedInputs),
+    } as unknown as ISchemaProperties;
+
+    return propObject;
+}
+
+export function transformActorInputSchemaProperties(
```
jirispilka, line 309: This is nice. IMO it would be a bit more readable and easier to debug without the recursion. But this is personal preference only.

src/tools/actor.ts
```diff
@@ -129,7 +138,10 @@ export async function getNormalActorsAsTools(
         if (actorDefinitionPruned) {
             const schemaID = getToolSchemaID(actorDefinitionPruned.actorFullName);
             if (actorDefinitionPruned.input && 'properties' in actorDefinitionPruned.input && actorDefinitionPruned.input) {
-                actorDefinitionPruned.input.properties = transformActorInputSchemaProperties(actorDefinitionPruned.input);
+                actorDefinitionPruned.input.properties = transformActorInputSchemaProperties(
```
jirispilka, line 142: Minor comment—no need to fix now, maybe later.
We currently have three things there: basic, full, and pruned (though “pruned” is a bit misleading). I know it’s a modified schema based on some transformation, but it could also be a schema without advancedInput.

jirispilka (changes requested):

jirispilka:
One more thing, we should document it in the README (query parameter + description)

And we'll need to include it in `docs.apify.com`, and `mcp-apify-server`

src/tools/actor.ts
MichalKalita: I revert this part

src/tools/utils.ts
MichalKalita: I don't know how to write this code better.
`separateAdvancedInputsInSchema` must be called before `filterSchemaProperties` (where are removed necesseary properties)
And all modificator functions must be called on root object and advancedInputs too.

Maybe modifiers can be moved to arrow function inside this function and be called twice.

like
```
    // Deep clone input to avoid mutating the original object
    const inputClone: IActorInputSchema = structuredClone(input);

    const transform = (transformInput: IActorInputSchema) => {
        let transformedProperties = markInputPropertiesAsRequired(transformInput);
        transformedProperties = buildApifySpecificProperties(transformedProperties);
        transformedProperties = filterSchemaProperties(transformedProperties);
        transformedProperties = inferArrayItemsTypeIfMissing(transformedProperties);
        transformedProperties = shortenProperties(transformedProperties);
        transformedProperties = addEnumsToDescriptionsWithExamples(transformedProperties);
        transformedProperties = encodeDotPropertyNames(transformedProperties);
        return transformedProperties;
    };

    if (options?.separateAdvancedInputs) {
        inputClone.properties = separateAdvancedInputsInSchema(inputClone.properties);
        if (inputClone.properties.advancedInput) {
            // Recursively transform the advanced input properties
            inputClone.properties.advancedInput.properties = transform(
                inputClone.properties.advancedInput as IActorInputSchema,
            );
        }
    }

    return transform(inputClone);
```

tests/integration/suite.ts
MichalKalita: I have a problem with this tests when I worked on this issue. Tests failed and I don't know why. This test is really simple and if fails, is easy to find where is a problem, what missing and what should not be here.

src/tools/utils.ts
jirispilka: Nice! IMO this is easier to understand and debug. I was thinking of a separate function for the transformation, but this arrow function would do the trick as well

tests/integration/suite.ts
MichalKalita: actorNameToToolName cannot be used here, when adding actor, clear name from Apify Console must be used

MQ37: yes, sorry my bad

tests/integration/suite.ts
MichalKalita: All snapshots can be updated using the command `npx vitest tests --update`.

MQ37: I don't like this added complexity. But if you are both @jirispilka and @MichalKalita for using this then I am fine with keeping that.

PR #218, opened 2025-08-12
feat(input): allow empty tools and actors to allow  greater control of exposed tools

MQ37:
closes https://github.com/apify/actors-mcp-server/issues/214

also handles this feature request https://console.apify.com/actors/1lSvMAaRcadrM1Vgv/issues/uMIkLfBQgw1hy8M95

Currently, default tools (Actor details, search, and add Actor — unless explicitly disabled via `enableAddingActors`) and `docs` category tools are always enabled and cannot be disabled. This is poor UX — some users may want access only to docs and not call Actors, or may want access to just a single Actor without any other tools.

This PR addresses that. When the user does not specify any parameters (e.g., `?actors` or `?tools`), the default tools are loaded as specified in the README. However, if the user specifies, for example, `tools=storage`, only the `storage` tools category will be loaded, along with the `add-actor` tool and the default RAG web browser Actor — both of which can also be disabled if needed.

With this change, users can disable all Actors and tools entirely by setting parameters to empty values, such as `?actors=&tools=&enableAddingActors=false`. This configuration would expose no tools from the MCP server, rendering it effectively useless — but it demonstrates the fine-grained control now available.

⚠️ **Warning**: This change could break workflows where users, for example, enable only `tools=storage` but still expect `docs` tools to be available (since they're currently loaded by default). While this is a potentially "breaking" change, we should consider it minor. Users should explicitly list the tool categories they actively use and avoid relying on implicit defaults in production workflows.


tests/integration/suite.ts
```diff
@@ -414,13 +467,10 @@ export function createIntegrationTestsSuite(
...
 
-                // If the category is enabled by default, it should not be loaded again, and its tools
-                // are accounted for in the default tools.
-                const isCategoryInDefault = toolCategoriesEnabledByDefault.includes(category as ToolCategory);
-                const expectedTools = isCategoryInDefault ? [] : toolCategories[category as ToolCategory];
+                const expectedTools = toolCategories[category as ToolCategory];
                 const expectedToolNames = expectedTools.map((tool) => tool.tool.name);
 
-                expect(toolNames.length).toEqual(expectedTools.length + defaultTools.length + defaults.actors.length + addRemoveTools.length);
+                expect(toolNames.length).toEqual(expectedTools.length + defaults.actors.length + addRemoveTools.length);
```
MichalKalita: I don't like usage of `toolCategories` as dependency on source code. When some category will be removed, this test must fail, because of BC.

I think we need to test this MCP server like this
```
const client = await createClientFn({
    tools: ["docs"],
});

const loadedTools = await client.listTools();
const toolNames = getToolNames(loadedTools);
expect(toolNames).toEqual(["search-docs", "fetch-docs"])
```

MichalKalita: Helper `getToolNamesFromClientConfig` can be usefull.
```
expect(getToolNamesFromClientConfig({tools: ["docs"]})).toEqual(["search-docs", "fetch-docs"])
```

MQ37: I started working on that but did not like the solution that I came up with - it would need a bit of refactoring of the tests for a nicer solution which is not the scope of this PR. Let's keep it as it is and we can then decide how to treat the integration tests and refactor them.

MQ37:
> We’ll need to create a new PR or incorporate all the changes we discussed with Honza. In my opinion, it makes the most sense to include everything in a single PR.

I would be for creating a new PR and either base it from this branch or merge it into this branch where we would also refactor/simplify the whole input/parameter handling. I agree, that we should include everything in a one PR/release.

MichalKalita (approved):

PR #222, opened 2025-08-15
chore: add apify logo svg

MQ37:
Add file is we can update and fix the logo in https://docs.crewai.com/en/tools/automation/apifyactorstool and use the github asset link instead of wikimedia since wikiedia capitalizes filename for some reason and mint lowercases the icon name so its useless.

related to https://github.com/apify/ai-team/issues/58 (will close after crewai docs merge)


MQ37:
merging FYI @jirispilka

PR #225, opened 2025-08-19
fix: Actor input transform array items type inference

MQ37:
This PR fixes the Actor input transform issue where the MCP server did not properly build the `items` sub-property for `array` type input properties. This was caused by a misordering of the transform steps - we filtered out `editor` before building `items`, which requires the `editor` field. Also added a test for this case to prevent this in the future.


MichalKalita (approved):

MQ37:
FYI @jirispilka

PR #226, opened 2025-08-19
feat: vibe new mcp input spec

MQ37:


MQ37:
sorry guys, quite a large PR with a bit of refactoring. The main thing is to thoroughly review the test cases and see if they make sense in context of the new space and if we are not missing any scenario @jirispilka @MichalKalita

src/input.ts
```diff
@@ -1,49 +1,73 @@
...
+ * - Coerce `actors`, `tools` from string/array into trimmed arrays ('' → []).
+ * - Normalize booleans (including legacy `enableActorAutoLoading`).
+ * - Merge `actors` into `tools` so selection lives in one place.
+ *
+ * Semantics passed to the loader:
+ * - `undefined` → use defaults; `[]` → explicitly none.
  */
 export function processInput(originalInput: Partial<Input>): Input {
-    const input = originalInput as Input;
+    const input = { ...originalInput } as Input;
```
MichalKalita: This way is tricky for types. Some properties of objects can be forgotten easily.

A better way is to create constants like `actors`, `enableAddingActors`, etc.
At the end of function create new object with `return {actors, enableAddingActors, ....}` without retyping. TypeScript checks that the object is compatible with the function return type from the header.

It's easier to create a new Input type with all properties defined than modify all properties of Partial<Input> to make it fully defined.

src/utils/tools-loader.ts
```diff
@@ -20,32 +33,91 @@ export async function loadToolsFromInput(
...
+    const internalSelections: ToolEntry[] = [];
+    const actorSelectorsFromTools: string[] = [];
+    if (selectorsProvided && !selectorsExplicitEmpty) {
+        for (const selector of selectors as (string | ToolCategory)[]) {
+            const categoryTools = toolCategories[selector as ToolCategory];
+            if (categoryTools) {
+                internalSelections.push(...categoryTools);
+                continue;
+            }
+            const internalByName = getInternalToolByNameMap().get(String(selector));
```
MichalKalita, line 58: The function `getInternalTool` with a tool name parameter can be called and return the tool.

src/utils/tools-loader.ts
```diff
@@ -20,32 +33,91 @@ export async function loadToolsFromInput(
...
+    // De-duplicate by tool name for safety
+    const seen = new Set<string>();
+    const deduped = result.filter((entry) => {
+        const { name } = entry.tool;
+        if (seen.has(name)) return false;
+        seen.add(name);
+        return true;
+    });
+
+    return deduped;
```
MichalKalita, line 122: This function is a lot of code for a relatively simple thing.

I don't know why there is so much checking for input here. I hope we should use the function `processInput` and work with the "good" shape of data in code. Maybe create a type like `ClearInput`, where the list of tools will be a required `Set<string>`. Then, we don't have to check if it's defined or if items are unique.

After that this function can be implemented in 2 stages:
1. split tools to actors and tools (when includes `/` character)
2. load all tools and actors

tests/integration/suite.ts
```diff
@@ -231,6 +328,10 @@ export function createIntegrationTestsSuite(
             expect(result).toEqual(
                 {
                     content: [
+                        {
+                            text: expect.stringMatching(/^Actor finished with runId: .+, datasetId .+$/),
```
MichalKalita, line 331: nice, i didn't know this type of expect is possible

src/input.ts
MQ37: makes sense :+1:

src/utils/tools-loader.ts
MQ37: How would that simplify or make the code better? I am using function here only for lazy evaluation since import cause circular dependency. You would still need to load the map and lookup the entry. It would be useful if used in multiple places but this is only local for this file.

src/utils/tools-loader.ts
MQ37: I tried to refactor but did not like the complicated solution and I would prefer not to introduce it in this PR. I would keep it as it is and we can refactor in a different PR. The whole input processing logic deserves a proper refactor/simplification.

src/utils/tools-loader.ts
MichalKalita: It's just an opinion thing. I think my comment has easier code.
You can keep it as is.

src/tools/actor.ts
```diff
@@ -246,53 +244,84 @@ export async function getActorsAsTools(
...
+                    if (!details) {
+                        return {
+                            content: [{ type: 'text', text: `Actor information for '${actorName}' was not found. Please check the Actor ID or name and ensure the Actor exists.` }],
+                        };
+                    }
+                    return {
+                        content: [
+                            { type: 'text', text: `**Actor card**:\n${details.actorCard}` },
+                            { type: 'text', text: `**README:**\n${details.readme}` },
+                            { type: 'text', text: `**Input Schema:**\n${JSON.stringify(details.inputSchema, null, 0)}` },
```
jirispilka, line 310: README can be quite big, Perhaps we can return only input schema? Tha'ts what we want anyway

src/mcp/server.ts
```diff
@@ -477,6 +401,8 @@ export class ActorsMcpServer {
                         apifyToken,
                         userRentedActorIds,
                         progressTracker,
+                        // Passing as argument to prevent circular dependency
+                        getActorsAsTools,
```
jirispilka: I don't really like that we need to pass getActorAsTools here. There are better solutions IMO.
1. Either create a new file with this function only (I'm not sure whether it actually helps).
2. Or add a new class method to apifyMcpServer (as we are passing this instances to the internalTool.call anyway). This could something like  
``` public async loadActorsAsTools(actorIdsOrNames: string[], apifyToken: string): Promise<ToolEntry[]> {
       return getActorsAsTools(actorIdsOrNames, apifyToken);
   }
   ```
We have public functions dealing with loading tools

File: src/mcp/server.ts [hunk 7]
```diff
@@ -199,47 +177,6 @@ export class ActorsMcpServer {
...
-            await this.loadDefaultActors(process.env.APIFY_TOKEN as string);
-        }
-    }
-
-    /**
-     * Loads default tools if not already loaded.
-     * @param apifyToken - Apify API token for authentication
-     * @returns {Promise<void>} - A promise that resolves when the tools are loaded
-     */
-    public async loadDefaultActors(apifyToken: string): Promise<void> {
```
jirispilka, line 234: loadDefaultActors is being used in apify-mcp-internal. Either keep and deprecate it or we need to make changes there too

src/input.ts
```diff
@@ -1,49 +1,78 @@
...
+ * intent via `undefined` (use defaults later) vs empty (explicitly none).
  */
 import log from '@apify/log';
 
-import type { Input, ToolCategory } from './types.js';
+import type { Input, ToolSelector } from './types.js';
+
+// Helpers
+// Normalize booleans that may arrive as strings or be undefined.
+function toBoolean(value: unknown, defaultValue: boolean): boolean {
```
jirispilka: This function would deserve a test

src/input.ts
```diff
@@ -1,49 +1,78 @@
...
+    if (value === undefined) return defaultValue;
+    if (typeof value === 'boolean') return value;
+    if (typeof value === 'string') return value.toLowerCase() === 'true';
+    return defaultValue;
+}
+
+// Normalize lists from comma-separated strings or arrays.
+function normalizeList(value: string | string[] | undefined): string[] | undefined {
```
jirispilka: This function would deserve a test

src/stdio.ts
```diff
@@ -46,38 +46,32 @@ log.setLevel(log.LEVELS.ERROR);
...
-Tool categories are as follows:
-- docs: Search and fetch Apify documentation tools.
-- runs: Get Actor runs list, run details, and logs from a specific Actor run.
-- storage: Access datasets, key-value stores, and their records.
-- preview: Experimental tools in preview mode.
-
-Note: Tools that enable you to search Actors from the Apify Store and get their details are always enabled by default.
-`,
-        example: 'docs,runs,storage',
+For more details visit https://github.com/apify/actors-mcp-server`,
```
jirispilka: The link should be to mcp.apify.com (or docs)

tests/unit/input.test.ts
```diff
@@ -39,17 +41,16 @@ describe('processInput', () => {
         expect(processed.enableAddingActors).toBe(false);
     });
 
-    it('should default enableAddingActors to true when not provided', async () => {
+    it('should default enableAddingActors to false when not provided', async () => {
         const input: Partial<Input> = {
             actors: ['actor1'],
```
jirispilka: do we need to pass actors here? It is not being tested anyway

tests/unit/input.test.ts
```diff
@@ -58,28 +59,60 @@ describe('processInput', () => {
...
         const input: Partial<Input> = {
-            actors: ['actor1'],
-            // @ts-expect-error: purposely invalid key for test
             tools: ['docs', 'invalidKey', 'storage'],
         };
         const processed = processInput(input);
         expect(processed.tools).toEqual(['docs', 'invalidKey', 'storage']);
     });
+
+    it('should merge actors into tools selectors for backward compatibility', async () => {
```
jirispilka:
```suggestion
    it('should merge actors into tools for backward compatibility', async () => {
```

tests/unit/mcp.utils.test.ts
```diff
@@ -3,16 +3,18 @@ import { describe, expect, it } from 'vitest';
 import { parseInputParamsFromUrl } from '../../src/mcp/utils.js';
 
 describe('parseInputParamsFromUrl', () => {
-    it('should parse Actors from URL query params', () => {
+    it('should parse Actors from URL query params (as tools selectors)', () => {
         const url = 'https://actors-mcp-server.apify.actor?token=123&actors=apify/web-scraper';
```
jirispilka:
```suggestion
        const url = 'https://mcp.apify.com?token=123&actors=apify/web-scraper';
```

tests/unit/mcp.utils.test.ts
```diff
@@ -3,16 +3,18 @@ import { describe, expect, it } from 'vitest';
...
         const result = parseInputParamsFromUrl(url);
-        expect(result.actors).toEqual(['apify/web-scraper']);
+        expect(result.tools).toEqual(['apify/web-scraper']);
+        expect(result.actors).toBeUndefined();
     });
 
-    it('should parse multiple Actors from URL', () => {
+    it('should parse multiple Actors from URL (as tools selectors)', () => {
         const url = 'https://actors-mcp-server.apify.actor?actors=apify/instagram-scraper,lukaskrivka/google-maps';
```
jirispilka:
```suggestion
        const url = 'https://mcp.apify.com?actors=apify/instagram-scraper,lukaskrivka/google-maps';
```

tests/unit/mcp.utils.test.ts
```diff
@@ -39,9 +41,10 @@ describe('parseInputParamsFromUrl', () => {
         expect(result.enableAddingActors).toBe(false);
     });
 
-    it('should handle Actors as string parameter', () => {
+    it('should handle Actors as string parameter (as tools selectors)', () => {
         const url = 'https://actors-mcp-server.apify.actor?actors=apify/rag-web-browser';
```
jirispilka:
```suggestion
        const url = 'https://mcp.apify.com?actors=apify/rag-web-browser';
```

tests/unit/mcp.utils.test.ts
```diff
@@ -3,16 +3,18 @@ import { describe, expect, it } from 'vitest';
 import { parseInputParamsFromUrl } from '../../src/mcp/utils.js';
```
jirispilka, line 3: Please change to: `mcp.apify.com` in the entire file

tests/unit/mcp.utils.test.ts
```diff
@@ -3,16 +3,18 @@ import { describe, expect, it } from 'vitest';
...
 
 describe('parseInputParamsFromUrl', () => {
-    it('should parse Actors from URL query params', () => {
+    it('should parse Actors from URL query params (as tools selectors)', () => {
```
jirispilka:
```suggestion
    it('should parse Actors from URL query params (as tools)', () => {
```

tests/unit/mcp.utils.test.ts
```diff
@@ -3,16 +3,18 @@ import { describe, expect, it } from 'vitest';
...
         const url = 'https://actors-mcp-server.apify.actor?token=123&actors=apify/web-scraper';
         const result = parseInputParamsFromUrl(url);
-        expect(result.actors).toEqual(['apify/web-scraper']);
+        expect(result.tools).toEqual(['apify/web-scraper']);
+        expect(result.actors).toBeUndefined();
     });
 
-    it('should parse multiple Actors from URL', () => {
+    it('should parse multiple Actors from URL (as tools selectors)', () => {
```
jirispilka:
```suggestion
    it('should parse multiple Actors from URL (as tools)', () => {
```

tests/unit/mcp.utils.test.ts
```diff
@@ -39,9 +41,10 @@ describe('parseInputParamsFromUrl', () => {
         expect(result.enableAddingActors).toBe(false);
     });
 
-    it('should handle Actors as string parameter', () => {
+    it('should handle Actors as string parameter (as tools selectors)', () => {
```
jirispilka:
```suggestion
    it('should handle Actors as string parameter (as tools)', () => {
```

tests/unit/input.test.ts
```diff
@@ -4,20 +4,22 @@ import { processInput } from '../../src/input.js';
 import type { Input } from '../../src/types.js';
 
 describe('processInput', () => {
-    it('should handle string actors input and convert to array', async () => {
+    it('should handle string actors input and convert to tools selectors', async () => {
```
jirispilka:
```suggestion
    it('should handle string actors input and convert to tools', async () => {
```

tests/unit/input.test.ts
```diff
@@ -4,20 +4,22 @@ import { processInput } from '../../src/input.js';
...
             actors: 'actor1, actor2,actor3',
         };
         const processed = processInput(input);
-        expect(processed.actors).toEqual(['actor1', 'actor2', 'actor3']);
+        expect(processed.tools).toEqual(['actor1', 'actor2', 'actor3']);
+        expect(processed.actors).toBeUndefined();
     });
 
-    it('should keep array actors input unchanged', async () => {
+    it('should move array actors input into tools selectors', async () => {
```
jirispilka:
```suggestion
    it('should move array actors input into tools', async () => {
```

README.md
```diff
@@ -140,47 +140,75 @@ One of the most powerful features of using MCP with Apify is dynamic tool discov
...
+- `docs`
+- `apify/rag-web-browser`
+
+It is thus the same as if you configured the `tools` parameter like this:
+
+**For the hosted server:**
+```
+https://mcp.apify.com?tools=actors,docs,apify/rag-web-browser
+```
+
```
jirispilka: Please add: `Use https://mcp.apify.com/ to configure your server and then copy the configuration to your client.`

README.md
```diff
@@ -140,47 +140,75 @@ One of the most powerful features of using MCP with Apify is dynamic tool discov
...
+**For the CLI:**
+```bash
+npx @apify/actors-mcp-server --tools actors,docs,apify/rag-web-browser
+```
+
+If the tools parameter is specified, only the listed tools or categories will be enabled - no default tools will be included.
+
+> **⚠️ Important Recommendation**
```
jirispilka:
```suggestion
> **⚠️ Important recommendation**
```

README.md
```diff
@@ -140,47 +140,75 @@ One of the most powerful features of using MCP with Apify is dynamic tool discov
...
 | `get-dataset-items` | storage | Retrieve items from a dataset with support for filtering and pagination. |  |
+| `get-dataset-schema` | storage | Generate a JSON schema from dataset items. |  |
 | `get-key-value-store` | storage | Get metadata about a specific key-value store. |  |
 | `get-key-value-store-keys`| storage | List the keys within a specific key-value store. |  |
 | `get-key-value-store-record`| storage | Get the value associated with a specific key in a key-value store. |  |
 | `get-dataset-list` | storage | List all available datasets for the user. |  |
 | `get-key-value-store-list`| storage | List all available key-value stores for the user. |  |
+| `add-actor` | experimental | Add an Actor as a new tool for the user to call. |  |
+
+### Tools configuration
```
jirispilka, line 173: In the tools configuration, you first provide config for remote server, then for CLI and then again for remote server. I understand the reason but it is kind of breaking the flow. I would put remote config togeather and keep CLI at the end

jirispilka, line 173: I think it would be good to also add backward compatibility explanation, something like this:

### Backward compatibility

The v2 configuration preserves backward compatibility with v1 usage. Notes:

- `actors` param (URL) and `--actors` flag (CLI) are still supported.
  - Internally they are merged into `tools` selectors.
  - Examples: `?actors=apify/rag-web-browser` ≡ `?tools=apify/rag-web-browser`; `--actors apify/rag-web-browser` ≡ `--tools apify/rag-web-browser`.
- `enable-adding-actors` (CLI) and `enableAddingActors` (URL) are supported but deprecated.
  - Prefer `tools=experimental` or including the specific tool `tools=add-actor`.
  - Behavior remains: when enabled with no `tools` specified, the server exposes only `add-actor`; when categories/tools are selected, `add-actor` is also included.
- `enableActorAutoLoading` remains as a legacy alias for `enableAddingActors` and is mapped automatically.
- Defaults remain compatible: when no `tools` are specified, the server loads `actors`, `docs`, and `apify/rag-web-browser`.
  - If any `tools` are specified, the defaults are not added (same as v1 intent for explicit selection).
- `call-actor` is now included by default via the `actors` category (additive change). To exclude it, specify an explicit `tools` list without `actors`.

Existing URLs and commands using `?actors=...` or `--actors` continue to work unchanged.

src/tools/actor.ts
MQ37: I agree :+1:

src/mcp/server.ts
MQ37: I will look into how to solve this more elegantly :+1:

src/mcp/server.ts
MQ37: There will be changes needed even if we keep this function. I just wanted you initial review and then I need to also prepare draft of integrating the new (upcoming version after these merges) into the `apify-mcp-server-internal`. We will see if we keep this method or not.

src/mcp/server.ts
MQ37: Instead of passing that the `getActorsAsTools` I created the method in MCP server class as you suggested :+1:

PR #227, opened 2025-08-20
fix: dockerhub init unauth server start and tool list

MQ37:
In order to publish the server to dockerhub it needs to be able to pass their build task. This task build the docker image and starts the container with the server and tries to list the tools, which Apify server was failing at because of the dummy token passed where the Apify API was throwing an error. Changed the logic to delete the dummy token and use the API as unauth which is sufficient for loading the Actor tools and listing.

### DIFF
--- a/src/apify-client.ts
+++ b/src/apify-client.ts
@@ -31,6 +31,7 @@ export class ApifyClient extends _ApifyClient {
          * for server start and listing of tools.
          */
         if (options.token?.toLowerCase() === 'your-apify-token') {
+            // eslint-disable-next-line no-param-reassign
             delete options.token;
         }
 


MQ37:
> Well, this is really a hack, should we keep the token empty in the docker?
> 
> So that we would remove it only when empty, and we don't have to check for a specific value

In the dockerhub `server.yaml` it is defined as:
```yaml

    secrets:
        - name: apify-mcp-server.apify_token
          env: APIFY_TOKEN
          example: your-apify-token
```

I would keep this logic. Since now it will work for dokerhub and we can also keep the behaviour for users that execute the CLI without (or with empty) token and print the error message that they need to set the token.

PR #228, opened 2025-08-20
feat: cancellable Actor run

MichalKalita:
Closes: https://github.com/apify/actors-mcp-server/issues/160

Tested in:
-  Claude desktop: ❌  doesn't work, in Claude logs is missing a notification for cancel run
- VSCode: ✅ works
- Integration tests: ✅ MCP client has a manual request method, with the option to AbortSignal. Internally, the MCP Client sends a cancel notification when the signal is called. https://github.com/modelcontextprotocol/typescript-sdk/blob/1f5950be01927cb8e4e555cb5584c44db1604f39/src/shared/protocol.ts#L620 

I tried to write a test using manually sending a cancel notification, but I need a request ID, and I don't know where to find it because it's managed internally in the MCP client.

### DIFF
--- a/src/tools/actor.ts
+++ b/src/tools/actor.ts
@@ -398,11 +398,6 @@ The step parameter enforces this workflow - you cannot call an Actor without fir
 
                 return { content };
             } catch (error) {
-                if (error instanceof Error && error.message === 'Operation cancelled') {
-                    // Receivers of cancellation notifications SHOULD NOT send a response for the cancelled request
-                    // https://modelcontextprotocol.io/specification/2025-06-18/basic/utilities/cancellation#behavior-requirements
-                    return { };
-                }
                 log.error('Error with Actor operation', { error, actorName, step });
                 return {
                     content: [


src/tools/actor.ts
```diff
@@ -58,22 +58,46 @@ export async function callActorGetDataset(
     apifyToken: string,
     callOptions: ActorCallOptions | undefined = undefined,
     progressTracker?: ProgressTracker | null,
+    abortSignal?: AbortSignal,
```
jirispilka, line 58: I have this warning in IDE: Parameter abortSignal is not described in JSDoc

src/tools/actor.ts
```diff
@@ -58,22 +58,46 @@ export async function callActorGetDataset(
...
 ): Promise<CallActorGetDatasetResult> {
     try {
         const client = new ApifyClient({ token: apifyToken });
         const actorClient = client.actor(actorName);
 
-        // Start the actor run but don't wait for completion
+        // Check if already aborted
+        if (abortSignal?.aborted) {
+            throw new Error('Operation cancelled');
```
jirispilka: you are throwing exception that is caugth locally

MQ37 (commented):
As I already menitoned in person, please change the Actor in tests to either `apify/python-example` or `apify/rag-web-browser`. These are already on staging and I would prefer to not add a new Actor there to our testing suite. If that is not possible we can include this Actor (if truly needed). Otherwise LGTM, thank you :+1:

src/tools/actor.ts
```diff
@@ -58,22 +58,46 @@ export async function callActorGetDataset(
...
         if (progressTracker) {
             progressTracker.startActorRunUpdates(actorRun.id, apifyToken, actorName);
         }
 
-        // Wait for the actor to complete
-        const completedRun = await client.run(actorRun.id).waitForFinish();
+        // Create abort promise that handles both API abort and race rejection
+        const abortPromise = async () => new Promise<never>((_, reject) => {
+            abortSignal?.addEventListener('abort', async () => {
+                // Abort the actor run via API
```
jirispilka, line 76: Based on the spec: Both parties SHOULD log cancellation reasons for debugging

So should we add an INFO log here: MCP client requested cancellation, actorName ....

MichalKalita: Logs added

src/tools/actor.ts
MichalKalita: I refactored this function, to not throw error, but returns null

src/tools/actor.ts
```diff
@@ -45,31 +45,73 @@ export type CallActorGetDatasetResult = {
...
         }
 
-        // Wait for the actor to complete
-        const completedRun = await client.run(actorRun.id).waitForFinish();
+        // Create abort promise that handles both API abort and race rejection
+        const abortPromise = async () => new Promise<typeof CLIENT_ABORT>((resolve) => {
+            abortSignal?.addEventListener('abort', async () => {
+                // Abort the actor run via API
+                try {
+                    await client.run(actorRun.id).abort({ gracefully: true });
```
jirispilka: Is there is a reason to do abort gracefully?

src/tools/actor.ts
```diff
@@ -45,31 +45,73 @@ export type CallActorGetDatasetResult = {
...
+    abortSignal: AbortSignal,
+): Promise<CallActorGetDatasetResult | null>; // With abort signal, null is returned if the actor was aborted by the client
 export async function callActorGetDataset(
     actorName: string,
     input: unknown,
     apifyToken: string,
     callOptions: ActorCallOptions | undefined = undefined,
     progressTracker?: ProgressTracker | null,
-): Promise<CallActorGetDatasetResult> {
+    abortSignal?: AbortSignal,
```
jirispilka: Do we really need to overload this function?

The caller can handle the null result
```
const result = await callActorGetDataset({
  actorName,
  input,
  apifyToken,
  callOptions,
  progressTracker,
  // abortSignal can be passed here if available
});

if (result === null) {
  return {};
}

const { runId, datasetId, items } = result;
```

src/mcp/server.ts
```diff
@@ -540,6 +549,12 @@ export class ActorsMcpServer {
                         });
                         content.push(...itemContents);
                         return { content };
+                    } catch (error) {
+                        if (error instanceof Error && error.message === 'Operation cancelled') {
```
jirispilka: Where (and why) is Operation cancelled raised?

jirispilka (changes requested):
I left a couple of questions, if you can please address them

src/tools/actor.ts
MichalKalita: Nope, I will switch it to false.

MichalKalita:
I tested the cancellation of the job working for the actor and the call-actor tool. Checked with VSCode.
I keep tests there, only for the streamable transfer type.

I found a bug where the client in test is not closed when some test fails. It's fixed when the client is always closed in `afterEach`

PR #232, opened 2025-08-21
fix: dxt rename sever to apify-mcp-server

MQ37:


MQ37:
FYI @jirispilka @MichalKalita

PR #235, opened 2025-08-22
feat: Add deployment to smithery

jirispilka:


src/index.ts
```diff
@@ -1,8 +1,70 @@
...
+            .then((tools) => {
+                server.upsertTools(tools);
+                return true;
+            })
+            .catch((error) => {
+                // eslint-disable-next-line no-console
+                console.error('Failed to load tools:', error);
+                return false;
+            });
+        return server.server;
```
jirispilka: @MichalKalita I would be glad if you can help with this issue or push me in a right direction.

Dockerfile
```diff
@@ -1,32 +1,31 @@
-# Stage 1: Build the project
-FROM node:24-alpine AS builder
+FROM node:22.12-alpine AS builder
```
MQ37: question: The NodeJS 22 has end of Active support in 2 months, why the switch? https://endoflife.date/nodejs

Dockerfile
```diff
@@ -1,32 +1,31 @@
...
+
 COPY src ./src
 COPY tsconfig.json ./
 
-# Build the project
 RUN npm run build
 
-# Stage 2: Set up the runtime environment
-FROM node:24-alpine
+FROM node:22-alpine AS release
```
MQ37: same here - the version switch

Dockerfile
```diff
@@ -1,32 +1,31 @@
...
+ENV NODE_ENV=production
+ENV APIFY_TOKEN=your-api-key-here
+
 # Install production dependencies only
-RUN npm ci --omit=dev
+RUN npm ci --ignore-scripts --omit=dev
 
 # Set the entry point for the container
-ENTRYPOINT ["node", "dist/stdio.js"]
+ENTRYPOINT ["node", "dist/index.js"]
```
MQ37: **change: this breaks the DockerHub integration.**

When running `task build -- --tools apify-mcp-server` I am now getting:
```
 - SecretsUsedInArgOrEnv: Do not use ARG or ENV instructions for sensitive data (ENV "APIFY_TOKEN") (line 25)
2025/08/22 14:25:21 initializing mcp/apify-mcp-server: 
exit status 1
task: Failed to run task "build": exit status 1
```

To reproduce, clone https://github.com/docker/mcp-registry, change `servers/apify-mcp-server/server.yaml` branch in yaml to `feat/smithery` from `master`, then install https://taskfile.dev/ and run the command.

src/index.ts
```diff
@@ -1,8 +1,70 @@
+#!/usr/bin/env node
 /*
  This file provides essential functions and tools for MCP servers, serving as a library.
  The ActorsMcpServer should be the only class exported from the package
+
+ Also, it serves as the main entry point for smithery deployment.
```
MQ37: question: Why are we using the `index.ts` as the smithery entrypoint? Is that actually needed? Can we just create a new file for that to keep it separate?

Dockerfile
jirispilka: It was copy-paste from template, fixed

Dockerfile
jirispilka: Copy-paste from a template, fixed

src/index.ts
jirispilka: Yeah, I had it somewhere else before, and it wasn’t working. But in retrospect, I only tried pushing it once or twice. I can give it another try.

src/smithery.ts
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env node
+/*
+ This file provides essential functions and tools for MCP servers, serving as a library.
+ The ActorsMcpServer should be the only class exported from the package
+
+ Also, it serves as the main entry point for smithery deployment.
+*/
```
jirispilka:
```suggestion
/*
 This file serves as the main entry point for smithery deployment.
*/
```

jirispilka:
Currently, deployment is failing

image

package.json
```diff
@@ -7,6 +7,7 @@
     "node": ">=18.0.0"
   },
   "main": "dist/index.js",
+  "module": "./src/smithery.ts",
```
MichalKalita, line 10: Info I get from LLM how to use main and module:

```
- main should point to the CommonJS build (usually in dist/).
- module should point to the ESM build (also in dist/), not to the TypeScript source.
- Both should export the same API.
- Ensure all referenced files are published.
```

The problem is that we have other exported APIs in the main and module. I think it will break our internal server if Typescript looks at the module and sees only Smithery, not ActorsMcpServer.

jirispilka: I'm sorry for the late answer, as smithery deployment was deprioritized.
I'll fix it based on your comment.

jirispilka: It does not work at smithery, when I use:
```
    "main": "dist/index.js",
    "module": "dist/index.js", 

#15 6.878 ❌ Build failed: Error: Entry file specified in package.json not found at /app/dist/index.js.
```

Because smithery is not building the project 🤷🏻

MichalKalita: If some settings in Smithery are possible (like changing the root directory), we can create a directory with the package only for Smithery.

It looks like Smithery requirements are not compatible with normal package usage.

jirispilka: I'll try to deploy remote server, if possible

jirispilka:
Smithery now support remote URLs

We have tried but there is an issue with auth 
Context: https://apify.slack.com/archives/C04URRT3934/p1764239977289259

PR #253, opened 2025-09-02
ON HOLD feat: skyfire payments

MQ37:
related to https://github.com/apify/apify-mcp-server-internal/pull/149

This PR prepares the core MCP repo for custom payment agentic key provider.

## Backstory 

Till now we just received Apify token through Auth header Bearer token or `?token` query param. This now changes as we want to implement the Skyfire for agentic payments. So we will be supporting both Apify tokens and also [Skyfire payment tokens](https://docs.skyfire.xyz/docs/explore-products#pay) (PAY) as input.

## Main idea

The main ideas is the Apify token logic remains as it is and we will also accept the PAY token through the `skyfire-pay-id` header or the Auth Bearer auth header (we can recognize Apify token (starts either with `apify_api` or `integration_api` for OAuth). We will then pass that to the MCP tools as we did with the Apify token and then forward it to the Apify API server.

## Technical details

For this I introduced the `AuthToken` type so we can clearly mark from which provider (currently only Apify or Skyfire) the token is since we may support multiple providers in future and they may not be easily recognizable by their string value but for example only via the header they come from - this needs to be somehow passed from the HTTP server in this wrapper.

## Additional changes

- Removed two unused legacy functions `getRealActorID()` and `getActorDefinition()`.
- Also added integation tests for the MCP progress notificaiton for the Actor tool calls so we can easily verify they actually work in production.

### DIFF
--- a/src/apify-client.ts
+++ b/src/apify-client.ts
@@ -29,14 +29,11 @@ export function getApifyAPIBaseUrl(): string {
  * @param authToken
  * @private
  */
-function addSkyfireHeader(config: AxiosRequestConfig, authToken?: AuthToken): AxiosRequestConfig {
-    if (authToken?.type === 'skyfire') {
-        const updatedConfig = { ...config };
-        updatedConfig.headers = updatedConfig.headers ?? {};
-        updatedConfig.headers['skyfire-pay-id'] = authToken.value;
-        return updatedConfig;
-    }
-    return config;
+function addSkyfireHeader(config: AxiosRequestConfig, authToken: AuthToken): AxiosRequestConfig {
+    const updatedConfig = { ...config };
+    updatedConfig.headers = updatedConfig.headers ?? {};
+    updatedConfig.headers['skyfire-pay-id'] = authToken.value;
+    return updatedConfig;
 }
 
 export class ApifyClient extends _ApifyClient {
--- a/src/apify-client.ts
+++ b/src/apify-client.ts
@@ -3,7 +3,7 @@ import { ApifyClient as _ApifyClient } from 'apify-client';
 import type { AxiosRequestConfig } from 'axios';
 
 import { USER_AGENT_ORIGIN } from './const.js';
-import type { AuthToken } from './types.js';
+import type { AuthInfo } from './types.js';
 
 /**
  * Adds a User-Agent header to the request config.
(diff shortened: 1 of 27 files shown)


MichalKalita (approved):
:+1:

src/mcp/server.ts
```diff
@@ -160,7 +160,7 @@ export class ActorsMcpServer {
     * @param toolNames - Array of tool names to ensure are loaded
     * @param apifyToken - Apify API token for authentication
```
jirispilka: apifyToken -> authToken

src/mcp/server.ts
```diff
@@ -196,8 +196,8 @@ export class ActorsMcpServer {
      * @param apifyToken - Apify API token for authentication
```
jirispilka: apifyToken -> authToken

src/apify-client.ts
```diff
@@ -22,23 +23,59 @@ export function getApifyAPIBaseUrl(): string {
...
 }
 
+/**
+ * Adds Skyfire header to the request config if needed.
+ * @param config
+ * @param authToken
+ * @private
+ */
+function addSkyfireHeader(config: AxiosRequestConfig, authToken?: AuthToken): AxiosRequestConfig {
+    if (authToken?.type === 'skyfire') {
```
jirispilka: This is an extra if, you are calling this function to addSkyfireHeader. It is the caller responsibility to check that the type is skyfire. You have it below, line 72

jirispilka (changes requested):
Well, the generated summary is nice but IMO completely use-less, srry  :).
I would appreciate your thoughts more than AI ones.

What about this approach?

I'm not really a fan of passing around `AuthToken`  everywhere, when, most of the time, we just need actual token string. It's a bit clunky 

Instead of the AuthToken object, we'll just use plain strings. The ApifyClient will handle figuring out if it's an Apify token or Skyfire token internally, so tools don't need to worry.

We can expose this function for classification, to be reusable (or we can duplicate it). 
  
Why this is better:
 • cleaner API - tools just pass the token string
 • classification logic lives where it's actually used 
 • simpler code

We can also live with the `authToken` but it seems to be a bit overkill.

Yes, I'm being pedantic because you said it was vibe-coded 😜

src/apify-client.ts
MQ37: Agree, that this is too defensive :+1:

PR #260, opened 2025-09-03
feat: improve actor tool output

MQ37:


src/utils/schema-generation.ts
```diff
@@ -0,0 +1,110 @@
...
+        try {
+            const fallbackSchema = toJsonSchema(processedItems, {
+                arrays: { mode: 'first' },
+            }) as JsonSchemaArray;
+            return fallbackSchema;
+        } catch {
+            // If all attempts fail, return null
+            return null;
+        }
+    }
```
MichalKalita: There is no need to make try/catch recursion

```suggestion
    // Try to generate schema with full options first
    try {
        const schema = toJsonSchema(processedItems, {
            arrays: { mode: arrayMode },
        }) as JsonSchemaArray;
        return schema;
    } catch { /* ignore */ }

    try {
        const fallbackSchema = toJsonSchema(processedItems, {
            arrays: { mode: 'first' },
        }) as JsonSchemaArray;
        return fallbackSchema;
    } catch { /* ignore */ }

    // If all attempts fail, return null
    return null;
```

tests/integration/suite.ts
```diff
@@ -95,11 +95,13 @@ export function createIntegrationTestsSuite(
...
             client = await createClientFn();
             const tools = await client.listTools();
-            expect(tools.tools.length).toEqual(defaultTools.length + defaults.actors.length);
+            expect(tools.tools.length).toEqual(defaultTools.length + defaults.actors.length + 1);
 
             const names = getToolNames(tools);
             expectToolNamesToContain(names, DEFAULT_TOOL_NAMES);
             expectToolNamesToContain(names, DEFAULT_ACTOR_NAMES);
+            expect(names).toContain('get-actor-output');
+            await client.close();
```
MichalKalita, line 104: all `await client.close()` are not needed, because of `afterEach` on line 90.
AfterEach is safe, it's called if any test failed.

src/utils/schema-generation.ts
MQ37: Makes sense, I just copied the code from `get-dataset-schema`, will refactor :+1:

tests/integration/suite.ts
MQ37: I just put it there, just in case :D

MichalKalita (approved):

tests/integration/suite.ts
MichalKalita: It's not removed 😄

MQ37: I did not remove it :smile:

MQ37: I would keep that, it should not cause any issues, right?

src/utils/actor-response.ts
```diff
@@ -0,0 +1,54 @@
...
+
+If you need to retrieve additional data, use the "get-actor-output" tool with: datasetId: "${datasetId}". Be sure to limit the number of results when using the "get-actor-output" tool, since you never know how large the items may be, and they might exceed the output limits.
+`;
+
+    const itemsPreviewText = result.previewItems.length > 0
+        ? JSON.stringify(result.previewItems)
+        : `No items available for preview—either the Actor did not return any items or they are too large for preview. In this case, use the "get-actor-output" tool.`;
+
+    // Build content array
+    const content: ({ type: 'text'; text: string })[] = [
```
jirispilka: local variable is redundant, use return instead

src/utils/actor-response.ts
```diff
@@ -0,0 +1,54 @@
+import type { CallActorGetDatasetResult } from '../tools/actor';
+
+/**
+ * Builds the response content for actor tool calls.
```
jirispilka: It would be good to explain how the response is composed, especially when itemCount !== result.previewItems.length

We have a function comment that says nothing :)

src/tools/index.ts
```diff
@@ -36,6 +37,7 @@ export const toolCategories = {
         getDataset,
         getDatasetItems,
         getDatasetSchema,
+        getActorOutput,
```
jirispilka, line 40: I understand that logically, getActorOutput should be placed here BUT it might be really misleading for user. You will enable tools=actors and you will get an extra bonus of getActorOutput without any obvious reason. What about to add it into actors?

src/utils/tools-loader.ts
```diff
@@ -123,6 +124,17 @@ export async function loadToolsFromInput(
         result.push(...actorTools);
     }
 
+    /**
+     * If there is any tool that in some way, even indirectly (like add-actor), allows calling
+     * Actor, then we need to ensure the get-actor-output tool is available.
+     */
+    const hasCallActor = result.some((entry) => entry.tool.name === 'call-actor');
```
jirispilka, line 131: use HelperTools enum here

src/utils/tools-loader.ts
```diff
@@ -123,6 +124,17 @@ export async function loadToolsFromInput(
...
+    const hasActorTools = result.some((entry) => entry.type === 'actor');
+    const hasAddActorTool = result.some((entry) => entry.tool.name === 'add-actor');
```
jirispilka, line 133: use HelperTools enum here

src/const.ts
```diff
@@ -20,6 +29,7 @@ export enum HelperTools {
     ACTOR_CALL = 'call-actor',
     ACTOR_GET = 'get-actor',
     ACTOR_GET_DETAILS = 'fetch-actor-details',
+    ACTOR_OUTPUT_GET = 'get-actor-output',
```
jirispilka, line 31: let's remove unused fields as well: ACTOR_GET, ACTOR_REMOVE, APIFY_MCP_HELP_TOOL

tests/integration/internals.test.ts
```diff
@@ -33,6 +33,7 @@ describe('MCP server internals integration tests', () => {
         const expectedToolNames = [
             addTool.tool.name,
             ACTOR_PYTHON_EXAMPLE,
+            'get-actor-output',
```
jirispilka, line 36: use HelperTools enum here

jirispilka (changes requested):
It works great — I played with it, and so far it’s flawless 🙂

My biggest concern is that get-actor-output shows up without users explicitly knowing it.

For example, when you configure tools, you end up with one extra tool as a kind of “bonus.” I think get-actor-output should go into the actors category. It makes sense there, typically when you call an Actor, you also need to retrieve its output.

What do you think?

src/tools/index.ts
MQ37: Good catch! This should not be here, I would personally not add this tool to any category and kept it as "magical" tool that appears when you include any tool that can interact with Actors in any way (see README change that explains this). What do you think?

src/utils/tools-loader.ts
MQ37: I liked the @MichalKalita idea that the integration tests should reflect user interaction with the MCP server and since the user does not have access to the HelperTools enum and probably uses hardcoded string we should write tests in this way - use hardcoded strings. So if we change some tool name the test fails and so we know we might break something for some users. Does it make sense to you?

jirispilka: oh, ok, makes perfect sense!

src/tools/index.ts
jirispilka: I think the fewer surprises, the better. When a user specifies something, we shouldn’t magically include something else. That feels like a big surprise and just asks for trouble — what if the user doesn’t want it at all? And how would they switch it off? There’s no way…

Instead, I’d add it to the actors category (alongside call-actor), and maybe also to other categories if needed.

But, t should also be included whenever an Actor is added 🤔, ahh

tests/integration/suite.ts
MichalKalita: Manually closing is required only when more clients will be in a single test.

Every client will be closed by 
```
afterEach(async () => {
            await client?.close();
            client = undefined;
        });
```
Even if the test failed, afterEach is always run, and the client will be closed and disconnected. Without this afterEach, each test should use try..finally to close client after.

It was a bug before. When one test failed, the client was not disconnected, and afterAll failed on timeout because the server was waiting for the client.

src/tools/index.ts
MQ37: I would personally keep this as a "magical" tool since user may not know when to add this tool and then might be frustrated when he only gets partial or no Actor results (we encountered user issues and reports like this already). I understand that now it seems more "magical" and is kind of a surprise but it makes the whole Actor calling and output retrieval work and this is crucial for the workflows. So I would not put this tool into any category and keep it hidden and auto-enabled if needed (based on the tool loader logic). I think there is no reason to Actually switch off this tool if you are working with Actors. Maybe there is, when you want to only call some Actor like send email and you do not care about the results, but that is miniory of the workflows and I would not complicate tool configuration for majorty only to satisfy this one use case where user might want to turn this tool off.

Issue #264, opened 2025-09-05
Error when Actorized MCP server responds with type 'json'


MichalKalita:
Example run https://console.apify.com/actors/BEwUoGEsRUQC2iUPa/runs/1BdLZGbQFokmAl84L#log

Tool call `{"name":"call-actor","arguments":{"actor":"mcp-servers/brave-search-mcp-server","step":"call","input":{"query":"random cool spot to visit in Prague","count":3}}}`

Error output

```
2025-09-05T10:52:51.891Z ERROR Error when calling tool call-actor: [
2025-09-05T10:52:51.897Z   {
2025-09-05T10:52:51.899Z     "code": "invalid_union",
2025-09-05T10:52:51.902Z     "unionErrors": [
2025-09-05T10:52:51.904Z       {
2025-09-05T10:52:51.906Z         "issues": [
2025-09-05T10:52:51.908Z           {
2025-09-05T10:52:51.910Z             "received": "json",
2025-09-05T10:52:51.912Z             "code": "invalid_literal",
2025-09-05T10:52:51.914Z             "expected": "text",
2025-09-05T10:52:51.916Z             "path": [
2025-09-05T10:52:51.919Z               "content",
2025-09-05T10:52:51.921Z               1,
2025-09-05T10:52:51.923Z               "type"
2025-09-05T10:52:51.930Z             ],
2025-09-05T10:52:51.933Z             "message": "Invalid literal value, expected \"text\""
2025-09-05T10:52:51.935Z           },
2025-09-05T10:52:51.938Z           {
2025-09-05T10:52:51.940Z             "code": "invalid_type",
2025-09-05T10:52:51.943Z             "expected": "string",
2025-09-05T10:52:51.946Z             "received": "undefined",
2025-09-05T10:52:51.948Z             "path": [
2025-09-05T10:52:51.951Z               "content",
2025-09-05T10:52:51.953Z               1,
2025-09-05T10:52:51.956Z               "text"
2025-09-05T10:52:51.959Z             ],
2025-09-05T10:52:51.961Z             "message": "Required"
2025-09-05T10:52:51.963Z           }
2025-09-05T10:52:51.966Z         ],
2025-09-05T10:52:51.968Z         "name": "ZodError"
2025-09-05T10:52:51.970Z       },
2025-09-05T10:52:51.972Z       {
2025-09-05T10:52:51.974Z         "issues": [
2025-09-05T10:52:51.976Z           {
[...]

PR #265, opened 2025-09-05
fix: error when content type is json

MichalKalita:
Fixes https://github.com/apify/apify-mcp-server/issues/264

It was a simple issue, type JSON is not allowed, it must be text.

### DIFF
--- a/src/tools/actor.ts
+++ b/src/tools/actor.ts
@@ -363,7 +363,7 @@ The step parameter enforces this workflow - you cannot call an Actor without fir
                         return {
                             content: [
                                 { type: 'text', text: `Input validation failed for Actor '${actorName}': ${errors.map((e) => e.message).join(', ')}` },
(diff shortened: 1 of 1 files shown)


src/tools/actor.ts
```diff
@@ -363,7 +363,7 @@ The step parameter enforces this workflow - you cannot call an Actor without fir
                         return {
                             content: [
                                 { type: 'text', text: `Input validation failed for Actor '${actorName}': ${errors.map((e) => e.message).join(', ')}` },
-                                { type: 'json', json: actor.tool.inputSchema },
+                                { type: 'text', text: `Input Schema:\n${JSON.stringify(actor.tool.inputSchema, null, 2)}` },
```
MQ37: change: please remove the `null, 2` and use only the `JSON.strinfigy(data)` to save on tokens, this must be LLM readable, not human readable.

MQ37 (commented):
Address the change and LGTM, thank you 👍

MQ37 (approved):

Issue #263, opened 2025-09-05
Fix Apify API errors messages and return them instead of the generic one


MQ37:
```
{"time":"2025-09-04T14:27:38.468Z","level":"ERROR","msg":"Error calling actor","error":{"name":"ApifyApiError","message":"Monthly usage hard limit exceeded","stack":"ApifyApiError: Monthly usage hard limit exceeded\n
```

In the case above the generic "An error occurred while calling the tool." is returned instead of the Apify error mesage since the `if (error instanceof ApifyApiError) {` is not probably working correctly.

PR #266, opened 2025-09-09
feat: agentic payments v2

MQ37:
Integrates Skyfire PAY tokens into the MCP server that is passed via tool input props.

**Remains backwards compatible with the existing Apify token use case.**

closes https://github.com/apify/apify-mcp-server/issues/263

### Auth middleware

- Auth middleware does not require Apify token when using `?payment=skyfire` and adds bool `skyfireMode` to `request` object.

### Apify token passing

- Instead of passing `apifyToken` throughout the whole code base the interface of the functions changes to accept `apifyClient` instead, this is requires since in the Skyfire mode the client will set `skyfire-pay-id` header instead of the Apify token bearer (this is also more flexible and future proof).

### Tools

- Injected `skyfire-pay-id` input prop into all the `actor` tools types and the `call-actor` tools (only required in the `call` step).
- When these tools are called without the `skyfire-pay-id` we return the instructions with a notice that this input is required.
- Injected Skyfire usage instructions into the Actor and `call-actor` tool descriptions and also `call-actor` tools `info` step result

### Resources

- Added usage guide/README resource explaining the Skyfire usage mode

### Other

- Removed the try catch that was just rethrowing a generic error in `callActorGetDataset`

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -53,7 +53,7 @@ export const defaults = {
 };
 
 export const ACTOR_ADDITIONAL_INSTRUCTIONS = 'Never call/execute tool/Actor unless confirmed by the user.';
-export const SKYFIRE_MIN_CHARGE_USD = 0.005;
+export const SKYFIRE_MIN_CHARGE_USD = 5.0;
 export const SKYFIRE_SELLER_ID = process.env.SKYFIRE_SELLER_SERVICE_ID;
 
 export const SKYFIRE_TOOL_INSTRUCTIONS = `To run the Actor, you need to provide a Skyfire payment ID in the \`skyfire-pay-id\` input property. You first need to create the Skyfire PAY token by calling the \`create-pay-token\` tool from the Skyfire MCP server and then provide the created token in the \`skyfire-pay-id\` input property when calling Actor tool. The Skyfire PAY token is used for billing purposes and needs to be charged with at least ${SKYFIRE_MIN_CHARGE_USD} USD.${SKYFIRE_SELLER_ID ? ` The Skyfire seller ID for Apify MCP server is "${SKYFIRE_SELLER_ID}".` : ''} When retrieving Actor output using the \`get-actor-output\` tool, you also need to provide the same Skyfire PAY token ID in the \`skyfire-pay-id\` input property.`;


src/apify-client.ts
```diff
@@ -23,22 +28,37 @@ export function getApifyAPIBaseUrl(): string {
...
         }
 
+        const { skyfirePayId, ...clientOptions } = options;
+        const requestInterceptors = [addUserAgent];
+        /**
+         * Add skyfire-pay-id header if provided.
+         */
+        if (skyfirePayId) {
+            requestInterceptors.push((config) => {
+                const updatedConfig = { ...config };
```
MichalKalita, line 50: This type of first level clone don't have any effect. Config object itself is not mutated.
This function mutates `config.headers`, and the input config object is updated.

Here is simple preview how it works
image

Use `structuredClone` for full clone. Or nothing if mutation is ok.

MichalKalita (approved):

MQ37:
> @MQ37 I think it is good.
> 
> @MichalKalita
> 
> It makes the public interface a bit more complicated.
> 
>     1. What if we remove the need to pass a token when the API allows anonymous access?
> 
>     2. Junie suggested creating a client factory function that would allow you to pass a client, a function to create a client, or a token.
> 
> 
> I’ve created this PR but haven’t tested it yet—it was vibe-coded only!!! Just a proposal: https://github.com/apify/apify-mcp-server/pull/268/files

Thank you for review :+1: Guys want to test that tomorrow before AH and I still need to do some changes (we made some decisions during todays meeting) so I would definitely refactor this - we can always refactor later.

src/apify-client.ts
MQ37: I just copied to logic from the add user agent header and it uses this exact logic. I dont quite get what is the issue exactly here? Is there any major flaw?

jirispilka:
> @jirispilka I think only API where we can safely disable is the store search since that does not really require user account. Other endpoints like get Actor, build, etc. in some cases require user account - for example when user wants to use private Actor on their own account.

oh, make sense! I forgot about private Actors.

Issue #250, opened 2025-08-29
Actor MCP servers add timeout when connecting


MQ37:
Context: https://apify.slack.com/archives/C09AUCEABDM/p1756457659801699

When tried connecting to the `mcp.apify.com?actors=mcp-servers/slidespeak-mcp-server` the slidespeak MCP Actor threw and error and failed to start MCP web server. This prevented successful client connection to mcp.apify.com. Maybe we can also introduce some retry but that would require client to support tools list changed notification so it might not be worth it as of now.

PR #272, opened 2025-09-11
fix: Actorized MCP servers have 30 seconds timeout to connect

MichalKalita:
Fixes: https://github.com/apify/apify-mcp-server/issues/250

When the connection to the Actorized MCP server is not completed in 30 seconds, this MCP server silently fails and opens a user session without this Actorized MCP server.

Potential improvements:
- [x] inside `getMCPServersAsTools` load all MCP servers in parallel
- [ ] somehow write to client if MCP server is not connected, maybe create some resource with "startup errors?" @jirispilka @MQ37
  - Not completed, there is no reliable way to return the MCP error when starting the session


MQ37:
> jakub.kopecky/browserbase-mcp-server

Created issue for that https://github.com/apify/ai-team/issues/85

tests/integration/suite.ts
```diff
@@ -940,5 +940,21 @@ export function createIntegrationTestsSuite(
 
             await client.close();
         });
+
+        // Skipping tests, MCP actors are not available in staging environment
+        // TODO: run this tests only for local environment
```
MQ37: Maybe we should create some init script that adds all the required Actors to the environment via API

src/tools/actor.ts
```diff
@@ -231,17 +232,25 @@ async function getMCPServersAsTools(
...
-            actorsMCPServerTools.push(...serverTools);
+            return serverTools;
         } finally {
             if (client) await client.close();
         }
-    }
+    });
+
+    // Wait for all actors to be processed in parallel
+    const actorToolsArrays = await Promise.all(actorToolPromises);
```
MQ37, line 251: notice: for the most use cases it should be fine, but we should be aware of rate limits

src/mcp/client.ts
```diff
@@ -4,24 +4,68 @@ import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/
 
 import log from '@apify/log';
 
+import { ACTORIZED_MCP_CONNECTION_TIMEOUT_MSEC } from './const.js';
 import { getMCPServerID } from './utils.js';
 
+class TimeoutError extends Error {
```
MQ37: nit: I would more this to a separate file `errors.ts`

src/mcp/client.ts
```diff
@@ -4,24 +4,68 @@ import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/
...
+        client = await createMCPSSEClient(url, token);
+        return client;
+    } catch (error) {
+        if (error instanceof TimeoutError) {
+            log.warning('Connection to MCP server using SSE transport timed out', { url });
+            return null;
+        }
+
+        log.error('Failed to connect to MCP server using SSE transport', { cause: error });
+        throw error;
```
MQ37, line 45: nit/question: Why is the try block here? We are just catching and then re-throwing. I would remove the try block for the SSE.

tests/integration/suite.ts
MQ37: But this test is not critical as of now, so we can solve this later. But I would definitely create an issues for that. If you guys agree, please create and issue for this. @jirispilka @MichalKalita

MQ37 (approved):
Just a few nits/questions, otherwise LGTM

tests/integration/suite.ts
jirispilka: I would add only single server here. I understand that you want to test multiple servers but this will be difficult to replicate at other environments

jirispilka (commented):
I left only a few comments.  Can you please re-consider the number of mcp-servers u use for testing?

src/mcp/client.ts
MichalKalita: If I throw exception, connection will not be opened.
When TimeoutError occurs, only warning will be logged to us and connection will be opened, but without this MCP server.

src/tools/actor.ts
MichalKalita: I didn't find any rate limiting docs in the public docs on internal Notion. What type of limits do you mean?

tests/integration/suite.ts
MichalKalita: I will keep there only single server, ACTOR_MCP_SERVER_ACTOR_NAME

MichalKalita:
@jirispilka I kept only a single MCP server.

jirispilka (approved):
Looks good, thanks!

PR #273, opened 2025-09-12
feat: add dev get html skeleton tool

MQ37:
This adds a `dev` tool category with a `get-html-skeleton` tool that returns HTML structure of given page URL for building selectors when building scrapers.

The tool uses pagination to adhere to the max tool output length so I added HTML cache to keep the results for 5 minutes so LLM can fetch multiple pages faster if needed.

### DIFF
--- a/src/tools/get-html-skeleton.ts
+++ b/src/tools/get-html-skeleton.ts
@@ -30,10 +30,10 @@ const getHtmlSkeletonArgs = z.object({
         .optional()
         .default(false)
         .describe('Whether to enable JavaScript rendering. Enabling this may increase the time taken to retrieve the HTML skeleton.'),
-    page: z.number()
+    chunk: z.number()
         .optional()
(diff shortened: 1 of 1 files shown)


src/tools/get-html-skeleton.ts
```diff
@@ -0,0 +1,101 @@
...
+
+                if (!firstItem.html) {
+                    return buildMCPResponse([`The scraping Actor (${ACTOR_RAG_WEB_BROWSER}) did not return any HTML content for the URL: ${parsed.url}. Please check the Actor run for more details: ${run.id}`]);
+                }
+
+                strippedHtml = stripHtml(firstItem.html);
+                getHtmlSkeletonCache.set(parsed.url, strippedHtml);
+            }
+
+            // Pagination logic
```
jirispilka, line 87: Let's rename this to chunk instead of pagination

Issue #247, opened 2025-08-27
Add support for Actorized MCP server to `call-actor` tool


MQ37:
Right now the Actorized MCP servers cannot be used through the `call-actor` tool. We need to figure out a way how to support them through this tool. Maybe leverage the multi-step nature the tool?

PR #274, opened 2025-09-17
feat: call-actor add support for MCP server Actors

MQ37:
closes https://github.com/apify/apify-mcp-server/issues/247

**This is initial implementation, we will discuss this with JC and improve.**

This PR adds support for using MCP server Actors via the `call-actor` tool.

Proposed workflow:

- First the MCP client (LLM) uses the `call-actor` for the MCP Actor with `step=info`, this connects to the standby Actor MCP server, lists the tools and returns a list of all Actor MCP tools with their input schema
- To call any MCP server Actor tools the client has to use special Actorn name name syntax `{mcp actor name}:{toolname}` in `step=call`

Other changes:

- Moved `call-actor` to a separate file and the related logic to `utils/`

Notes:

**In case of Skyfire we just return MCP server (standby) Actors are not supported.**


src/tools/call-actor.ts
```diff
@@ -0,0 +1,238 @@
+import type { Client } from '@modelcontextprotocol/sdk/client/index.js';
```
jirispilka: renaming actor.ts to call-actor.ts make sense but it'll create a mege conflict in other PR. I can fix but it is extra work, just for renaming file :shrug

src/utils/actor.ts
```diff
@@ -1,6 +1,316 @@
...
+            actorsMCPServerTools.push(...serverTools);
+        } finally {
+            if (client) await client.close();
+        }
+    }
+
+    return actorsMCPServerTools;
+}
+
+export async function getActorsAsTools(
```
jirispilka: Again, this just begs for merge conflict and extra work for us (without much benefit :( crying a lot))

MQ37:
> Looks good, but I would really love if I would have to solve merge conflicts with this PR: #275
> 
> And I told you that :(

Sorry, already did that before you told me :( Will revert the changes and we can then create a refactor PR.

MQ37:
reverted all the refactors, please give final approve and I will merge @jirispilka

PR #277, opened 2025-09-18
fix: duplicate skyfire description when listing tools multiple times

MQ37:
Currently when listing multiple times the description is always appended and the tool description explodes. This fixes it so we modify the description only on tool upsert.


jirispilka (approved):
Thanks!

PR #281, opened 2025-09-19
fix: Change dxt to mcpb in release cicd

MQ37:


jirispilka:
@MQ37 I can't add you to the review, for some reason, can you please check?

BTW. we need to fix this in order to release apify-mcp-server :)

PR #290, opened 2025-09-24
fix: Update github workflows

jirispilka:
I think we need to add a workflow to fix auto-versioning of deployment. The step, which update_changelog causes the release.yaml to fail when the release is triggered with auto-mode.

I Haven't tested this yet


MQ37:
for now we decided to pospone this PR and just use the `custom` (manual) versioning which is working as expected

PR #299, opened 2025-09-30
feat: swap call-actor with add-actor for actors tool category for supported clients

MQ37:
closes https://github.com/apify/ai-team/issues/93

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,7 +20,7 @@
         "apify-client": "^2.12.6",
         "cheerio": "^1.1.2",
         "express": "^4.21.2",
-        "mcp-client-capabilities": "^0.0.4",
+        "mcp-client-capabilities": "^0.0.5",
         "to-json-schema": "^0.2.5",
         "turndown": "^7.2.0",
         "yargs": "^17.7.2",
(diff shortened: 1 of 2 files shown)


src/utils/tools-loader.ts
```diff
@@ -68,6 +71,14 @@ export async function loadToolsFromInput(
             }
 
             const categoryTools = toolCategories[selector as ToolCategory];
+
+            // Handler client capabilities logic for 'actors' category to swap call-actor for add-actor
+            // if client supports dynamic tools.
+            if (selector === 'actors' && doesMcpClientSupportDynamicTools(initializeRequestData)) {
+                internalSelections.push(...categoryTools.filter((t) => t.tool.name !== 'call-actor'));
```
jirispilka: Should we use the name from the tools enum?

jirispilka (approved):
Looks good to me. I don't llike that we need to pass InitializedRequest around but I don't see a better way. 

Thanks. I update the readme a bit

src/utils/tools-loader.ts
MQ37: Yes, missed that - thank you :+1:

Issue #282, opened 2025-09-19
Fix the tool output markdown format based on JC sync


MQ37:
Context: https://docs.google.com/document/d/1JScVXMhRjMuztqcHqL94WhvlofNjxrAj34Ani9LOz-k/edit?tab=t.0#heading=h.scx5khoaj0eq

MQ37:
@jirispilka We already did markdown formatting for tools where it makes sense and now as we want to implement the TOON format https://github.com/apify/apify-mcp-server/issues/913 this is not really relevant I think. I would close this issue.

jirispilka:
Yes, let's close it

PR #320, opened 2025-10-23
feat: improve tool output json markdown format

MQ37:
related to https://github.com/apify/apify-mcp-server/issues/282

Improved the output format by wrapping JSON in the backtick JSON markdown block (as per JC's feedback)

### DIFF
--- a/src/tools/actor.ts
+++ b/src/tools/actor.ts
@@ -407,7 +407,7 @@ EXAMPLES:
                             }
                             const toolsResponse = await client.listTools();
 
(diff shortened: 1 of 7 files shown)


jirispilka (changes requested):
Thanks! But le'ts not use indent:
```
JSON.stringify(datasets, null, 2)
```

For some reason it cost almost double of the tokens: https://github.com/apify/apify-mcp-server/pull/283#issuecomment-3323763589

"On real data from Google map, JSON to markdown is 11% smaller for normal JSON (JSON.stringify without formatting params) and 44% smaller than formatted JSON JSON.stringify(json, null, 2)"

PR #321, opened 2025-10-25
feat: Update search-actors tool

jirispilka:
I really struggled to get GPT to work, so I ended up analyzing agentic prompts to better understand their tool instructions.

I had to update the system prompt so that GPTs would actually recognize that there are tools available to use.
```
You are a helpful assistant with a set of tools.

Follow these rules regarding tool calls:
1. ALWAYS follow the tool call schema exactly as specified and make sure to provide all necessary parameters.
2. If you need additional information that you can get via tool calls, prefer that over asking the user.
3. Only use the standard tool call format and the available tools.
```

Other changes:
- Changed tool description and arguments description
- Refactored evaluation

Edit - there is a failed test cases analysis file that can be used to continue with other tools

### DIFF
--- a/evals/2025-11-04-failed-cases-analysis.md
+++ b/evals/2025-11-04-failed-cases-analysis.md
@@ -1,0 +1,275 @@
+# Failed Cases Analysis & Implementation Guide
+
+## Summary
+
+103 failed test cases across 6 experiments. fix by improving tool descriptions (most important factor per `evals/README.md`).
+
+**Failed cases:** 103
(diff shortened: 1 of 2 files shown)


jirispilka:
Performance on the complete dataset (only tool exact mathc)

[Before](https://app.phoenix.arize.com/s/apify/datasets/RGF0YXNldDo1NA==/experiments):
image

[After](https://app.phoenix.arize.com/s/apify/datasets/RGF0YXNldDo3NQ==/experiments)
image

Issue #115, opened 2025-05-13
Improve Tool interface typing


MQ37:
Use something like this so the tool type is more explicit:
```

interface Tool {
  type: string;
  name: string;
  ...
}

interface InternalTool extends Tool {
  type: 'internal';
}

interface ActorTool extends Tool {
  type: 'actor';
}
```

PR #323, opened 2025-10-31
refactor: simplify the tool types

MQ37:
closes https://github.com/apify/apify-mcp-server/issues/115

This bothered me for a long time and decided to give it a go today.  I believe these code changes make the codebase more maintainable and simple. **This was partially vibe coded.**

Changes:
- got rid of the wrapper tool type
- use only specific types based in the type discriminator - no nesting of the tool object
- simplified code where I could

### DIFF
--- a/tests/integration/suite.ts
+++ b/tests/integration/suite.ts
@@ -405,7 +405,7 @@ export function createIntegrationTestsSuite(
             const result = await client.callTool({
                 name: HelperTools.STORE_SEARCH,
                 arguments: {
-                    search: query,
+                    keywords: query,
                     limit: 5,
                 },
             });
@@ -422,7 +422,7 @@ export function createIntegrationTestsSuite(
             const result = await client.callTool({
                 name: HelperTools.STORE_SEARCH,
                 arguments: {
-                    search: 'rental',
+                    keywords: 'rental',
                     limit: 100,
                 },
             });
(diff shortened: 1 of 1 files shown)


src/types.ts
```diff
@@ -60,24 +61,29 @@ export type ActorDefinitionPruned = Pick<ActorDefinitionWithDesc,
...
     ajvValidate: ValidateFunction;
 }
 
+/**
+ * Type for MCP SDK's inputSchema constraint.
+ * Extracted directly from the MCP SDK's ToolSchema to ensure alignment with the specification.
+ * The MCP SDK requires inputSchema to have type: "object" (literal) at the top level.
+ * Use this type when casting schemas that have type: string to the strict MCP format.
+ */
+export type McpInputSchema = z.infer<typeof ToolSchema>['inputSchema'];
```
jirispilka: Can you think of better name than McpInputSchema? This is really vague. `ToolInputSchema` ?

jirispilka (approved):
Really nice! Thanks!

I only check couple of tools as the rest of the changes were quite similar (for other tools)

Consider renaming `McpInputSchema`. This is not descriptive IMO

src/types.ts
MQ37: good point, `ToolInputSchema` is much better :+1:

PR #326, opened 2025-11-04
fix: Update readme, add image with clients

jirispilka:
fix: Update readme, clients

(well I did not move the clients to the top)
I felt it was more important to first explain what the MCP server does.

FYI @lukas-bekr

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ You can find detailed instructions for setting up the MCP server in the [Apify d
 # 🤖 MCP clients
 
 Apify MCP Server is compatible with any MCP client that adheres to the [Model Context Protocol](https://modelcontextprotocol.org/), but the level of support for dynamic tool discovery and other features may vary between clients.
-Therefore, the server uses [mcp-client-capabilities](https://github.com/apify/mcp-client-capabilities) to detect client capabilities and adjust its behavior accordingly.
+<!--Therefore, the server uses [mcp-client-capabilities](https://github.com/apify/mcp-client-capabilities) to detect client capabilities and adjust its behavior accordingly.-->
 
 To interact with the Apify MCP server, you can use clients such as: [Claude Desktop](https://claude.ai/download), [Visual Studio Code](https://code.visualstudio.com/), or [Apify Tester MCP Client](https://apify.com/jiri.spilka/tester-mcp-client).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -66,34 +66,29 @@ You can use the Apify MCP Server in two ways:
...
+# 🤖 MCP clients
 
-To interact with the Apify MCP server, you can use various MCP clients, such as:
-- [Claude Desktop](https://claude.ai/download)
-- [Visual Studio Code](https://code.visualstudio.com/)
-- [Apify Tester MCP Client](https://apify.com/jiri.spilka/tester-mcp-client)
-- Other clients at [https://modelcontextprotocol.io/clients](https://modelcontextprotocol.io/clients)
-- More clients at [https://glama.ai/mcp/clients](https://glama.ai/mcp/clients)
+Apify MCP Server is compatible with any MCP client that adheres to the [Model Context Protocol](https://modelcontextprotocol.org/), but the level of support for dynamic tool discovery and other features may vary between clients.
+Therefore, the server uses [mcp-client-capabilities](https://github.com/apify/mcp-client-capabilities) to detect client capabilities and adjust its behavior accordingly.
```
MQ37: nit: Actually we are not using the client capabilities as of now - the logic is commented out; since the `add-actor` is not working as expected with the new LLM models. But I agree that we should link this package to promote it.

README.md
```diff
@@ -66,34 +66,29 @@ You can use the Apify MCP Server in two ways:
...
-| Client | Dynamic Tool Discovery | Notes |
-| --- | --- | --- |
-| **Claude.ai (web)** | ✅ Full | |
-| **Claude Desktop** | 🟡 Partial | Tools may need to be reloaded manually in the client. |
-| **VS Code (Genie)** | ✅ Full | |
-| **Apify Tester MCP Client** | ✅ Full | Designed for testing Apify MCP servers. |
+| Client                      | Dynamic Tool Discovery | Notes                                                 |
+|-----------------------------|------------------------|-------------------------------------------------------|
+| **Claude.ai (web)**         | ✅ Full                 |                                                       |
+| **Claude Desktop**          | 🟡 Partial             | Tools may need to be reloaded manually in the client. |
```
MQ37: change required: Claude.ai web behaves the same as the Claude desktop, right? In both cases the `add-actor` requires manual tool reload. So I would either put full in both cases or both mark as partial.

README.md
```diff
@@ -66,34 +66,29 @@ You can use the Apify MCP Server in two ways:
...
+| **VS Code (Genie)**         | ✅ Full                 |                                                       |
+| **Cursor**                  | ✅ Full                 |                                                       |
+| **Apify Tester MCP Client** | ✅ Full                 | Designed for testing Apify MCP servers.               |
```
MQ37: Let's also add OpenCodes since we used that for the demo.

```suggestion
| **Apify Tester MCP Client** | ✅ Full                 | Designed for testing Apify MCP servers.               |
| **OpenCode**                        | ✅ Full                 |                                                                                       |
```

MQ37 (approved):
Thanks! Please address the comments before merging.

PR #329, opened 2025-11-07
feat: segment telemetry

MQ37:
This PR implements the Segment telemetry.

Implemented based on https://www.notion.so/apify/Apify-MCP-server-analytics-2a1f39950a2280c4a9a5c8154dfc0f1a.

related to https://github.com/apify/ai-team/issues/56 (I would mark completed manually after we verify in prod).

**This PR does not implement the `anonymousId` (Device ID) yet for unauth requests (coming soon) - I would implement this with the PR where we are going to implement the unauth requests for docs.**

Other changes:
- added `TELEMETRY_ENABLED` env var to doppler. Localhost=false, for staging and prod it is enabled.
- added `TELEMETRY_ENV` with options `dev` and `prod`

Bonus features:
- implemented reading Apify token from the `~/.apify/auth.json` file that is created by Apify CLI so if you are a CLI user you can use the server without specifying the env var.


src/actor/server.ts
```diff
@@ -157,12 +171,25 @@ export function createExpressApp(
...
             } else if (!sessionId && isInitializeRequest(req.body)) {
-            // New initialization request - use JSON response mode
+            // New initialization request
                 transport = new StreamableHTTPServerTransport({
                     sessionIdGenerator: () => randomUUID(),
                     enableJsonResponse: false, // Use SSE response mode
                 });
-                const mcpServer = new ActorsMcpServer({ setupSigintHandler: false, initializeRequestData: req.body as InitializeRequest });
+                // Extract telemetry query parameters
+                const urlParams = new URL(req.url, `http://${req.headers.host}`).searchParams;
```
jirispilka, line 181: This code is duplicated, BUT, we aim to remove the actor/** anyway. So it is fine for now

package.json
```diff
@@ -42,6 +43,7 @@
     "@apify/datastructures": "^2.0.3",
     "@apify/log": "^2.5.16",
     "@modelcontextprotocol/sdk": "^1.18.1",
+    "@segment/analytics-node": "^2.3.0",
```
katzino, line 46: Ideally, I would use `@apify-packages/analytics` package directly. Mainly because
- it comes with typed event names
- it forces you to add all requires properties to events and its metadata

src/const.ts
```diff
@@ -104,3 +106,13 @@ export const ALGOLIA = {
...
 
 export const APIFY_STORE_URL = 'https://apify.com';
+
+// Telemetry
+export type TelemetryEnv = 'dev' | 'prod';
+
+export const TELEMETRY_ENV = {
+    DEV: 'dev',
+    PROD: 'prod',
+} as const;
```
katzino: Nit: not important since it won't change much but you can extrapolate typings from original const
```suggestion
export const TELEMETRY_ENV = {
    DEV: 'dev',
    PROD: 'prod',
} as const;
export type TelemetryEnv = (typeof TELEMETRY_ENV)[keyof typeof TELEMETRY_ENV];
```

src/telemetry.ts
```diff
@@ -0,0 +1,59 @@
+import { Analytics } from '@segment/analytics-node';
+
+import { DEFAULT_TELEMETRY_ENV, TELEMETRY_ENV, type TelemetryEnv } from './const.js';
+import type { ToolCallTelemetryProperties } from './types.js';
+
+const DEV_WRITE_KEY = '9rPHlMtxX8FJhilGEwkfUoZ0uzWxnzcT';
+const PROD_WRITE_KEY = 'cOkp5EIJaN69gYaN8bcp7KtaD0fGABwJ';
```
katzino, line 17: Should those be passed from doppler? Also I checked what we have there and it doesn't match the values you're using here 😅

src/telemetry.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const SEGMENT_EVENTS = {
+    TOOL_CALL: 'MCP Tool Call',
+};
```
katzino: As I mentioned above, when using `@apify-packages/analytics` it defines what events can be used. We're trying to have all events defined in one place so it's easier for everyone to understand what we're actually tracking.

src/telemetry.ts
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * Gets the telemetry environment, defaulting to 'prod' if not provided or invalid
+ */
+export function getTelemetryEnv(env?: string | null): TelemetryEnv {
+    return (env === TELEMETRY_ENV.DEV || env === TELEMETRY_ENV.PROD) ? env : DEFAULT_TELEMETRY_ENV;
+}
+
+// Map to store singleton Segment Analytics clients per environment
+const analyticsClients = new Map<TelemetryEnv, Analytics>();
```
katzino: I don't think this should be needed. You will only need a single client at the time, no need to define it for all envs.

src/types.ts
```diff
@@ -293,3 +293,21 @@ export type DatasetItem = Record<number | string, unknown>;
  * Can be null or undefined in case of Skyfire requests.
  */
 export type ApifyToken = string | null | undefined;
+
+/**
+ * Properties for tool call telemetry events sent to Segment.
+ */
+export interface ToolCallTelemetryProperties {
+    app_name: string;
```
katzino: Every event that we send to segment across all platforms comes with `app` key that is defined here
https://github.com/apify/apify-core/blob/develop/src/packages/consts/src/analytics.ts#L4

As far as I know, you're already tracking page views on MCP, so ideally use the same `app` value here as well.

src/telemetry.ts
```diff
@@ -0,0 +1,59 @@
...
+export function trackToolCall(
+    userId: string,
+    env: TelemetryEnv,
+    properties: ToolCallTelemetryProperties,
+): void {
+    const client = getOrInitAnalyticsClient(env);
+
+    // TODO: Implement anonymousId tracking for device/session identification
+    client.track({
+        userId: userId || '',
```
katzino: Basically if `userId === ''` and `anonymousId` isn't provided segment won't be able to create an event as far as I remember. You can check it out how we handle server event on the web here
https://github.com/apify/apify-web/blob/develop/src/common/services/analytics.ts#L8

In a nutshell, if you're not able to obtain `anonymousId` anywhere, we use `crypto.randomUUID()`

katzino (commented):
I wouldn't personally refer to this initiative as `telemetry` but `analytics` instead. Or at least this is what we do in terms of segment usage across all our codebase.

src/telemetry.ts
jirispilka: You are correct, in 99.9% of cases it won't be needed. But we expose telemetry options at remote server too.

Users can opt out from telemetry by setting: `mcp.apify.com?telemetry-enabled=false`

But you can also enable different env `mcp.apify.com?telemetry-enabled=true&telemetry-env=prod` (this is setup at request level) or `telemetry-env=dev`

Thinking about this again, I'll remove `telemetry-env` query param, it does not really make sense. It was there for convenience only

package.json
jirimoravcik: Sadly this package is public, so we can't use a private package

src/actor/server.ts
```diff
@@ -78,7 +80,19 @@ export function createExpressApp(
                 rt: Routes.SSE,
                 tr: TransportType.SSE,
             });
-            const mcpServer = new ActorsMcpServer({ setupSigintHandler: false });
+            // Extract telemetry query parameters
+            const urlParams = new URL(req.url, `http://${req.headers.host}`).searchParams;
+            const telemetryEnabledParam = urlParams.get('telemetry-enabled');
+            const telemetryEnvParam = urlParams.get('telemetry-env');
+            const telemetryEnabled = telemetryEnabledParam !== 'false'; // Default to true
+            const telemetryEnv: TelemetryEnv = getTelemetryEnv(telemetryEnvParam);
```
jirimoravcik: I see this repeated below, can we extract it to some utility function?

tests/helpers.ts
```diff
@@ -13,17 +13,18 @@ export interface McpClientOptions {
     tools?: (ToolCategory | string)[]; // Tool categories, specific tool or Actor names to include
     useEnv?: boolean; // Use environment variables instead of command line arguments (stdio only)
     clientName?: string; // Client name for identification
+    telemetryEnabled?: boolean; // Enable or disable telemetry (default: false for tests)
+    telemetryEnv?: TelemetryEnv; // Telemetry environment (default: 'prod', only used when telemetryEnabled is true)
```
jirimoravcik: `telemetry?: { enabled: boolean, env: TelemetryEnv }`

tests/helpers.ts
```diff
@@ -33,6 +34,22 @@ export async function createMcpSseClient(
     if (tools !== undefined) {
         url.searchParams.append('tools', tools.join(','));
     }
+    // Append telemetry parameters
+    if (telemetryEnabled !== undefined) {
+        url.searchParams.append('telemetry-enabled', telemetryEnabled.toString());
+    }
+    if (telemetryEnv !== undefined && telemetryEnabled !== false) {
+        url.searchParams.append('telemetry-env', telemetryEnv);
+    }
```
jirimoravcik: What if the values are `null` or just an empty string? `!== undefined` feels too specific

src/const.ts
```diff
@@ -104,3 +106,12 @@ export const ALGOLIA = {
 export const PROGRESS_NOTIFICATION_INTERVAL_MS = 5_000; // 5 seconds
 
 export const APIFY_STORE_URL = 'https://apify.com';
+
+// Telemetry
+export const TELEMETRY_ENV = {
+    DEV: 'dev',
+    PROD: 'prod',
+} as const;
```
jirimoravcik, line 114: This is a really bad practice, can we make the mapping 1:1 - either both (the keys and values) uppercase or lowercase.

src/utils/version.ts
```diff
@@ -0,0 +1,24 @@
...
+import { fileURLToPath } from 'node:url';
+
+/**
+ * Gets the package version dynamically from package.json
+ * Returns null if the file cannot be read
+ */
+export function getPackageVersion(): string | null {
+    try {
+        // Read package.json and extract version
+        // In production, this will be replaced at build time
```
jirimoravcik: What does "this will be replaced at build time" mean?

src/utils/version.ts
```diff
@@ -0,0 +1,24 @@
+import { readFileSync } from 'node:fs';
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+/**
+ * Gets the package version dynamically from package.json
+ * Returns null if the file cannot be read
+ */
+export function getPackageVersion(): string | null {
```
jirimoravcik, line 10: Could `require('./package').version` work? or even `process.env.npm_package_version` if you start it via `npm`?

This read file + JSON parsing seems almost excessive to me.

src/utils/user-cache.ts
```diff
@@ -0,0 +1,42 @@
...
+
+// LRU cache with TTL for user info - stores the raw User object from API
+const userCache = new TTLLRUCache<User>(USER_CACHE_MAX_SIZE, USER_CACHE_TTL_SECS);
+
+/**
+ * Gets user info from token, using cache to avoid repeated API calls
+ * Token is hashed before caching to avoid storing raw tokens
+ * Returns the full User object from API or null if not found
+ */
+export async function getUserIdFromToken(
```
jirimoravcik:
```suggestion
export async function getUserInfoFromTokenCached(
```
(nit) it's always a good practice to mention that the call is cached.
Also I think you're getting UserInfo, not userId, right?

src/utils/user-cache.ts
```diff
@@ -0,0 +1,42 @@
+import { createHash } from 'node:crypto';
+
+import type { User } from 'apify-client';
+
+import type { ApifyClient } from '../apify-client.js';
+import { USER_CACHE_MAX_SIZE, USER_CACHE_TTL_SECS } from '../const.js';
+import { TTLLRUCache } from './ttl-lru.js';
+
+// LRU cache with TTL for user info - stores the raw User object from API
+const userCache = new TTLLRUCache<User>(USER_CACHE_MAX_SIZE, USER_CACHE_TTL_SECS);
```
jirimoravcik: (opt) If we're using this in the internal MCP as well, it'd be nice to use `memoize` instead, as that can leverage Redis and share the cache between instances.

src/utils/user-cache.ts
```diff
@@ -0,0 +1,42 @@
...
+    // Hash token for cache key
+    const tokenHash = createHash('sha256').update(token).digest('hex');
+
+    // Check cache first
+    const cachedUser = userCache.get(tokenHash);
+    if (cachedUser) {
+        return cachedUser;
+    }
+
+    // Fetch from API
```
jirimoravcik:
```suggestion
```

src/utils/user-cache.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Returns the full User object from API or null if not found
+ */
+export async function getUserIdFromToken(
+    token: string,
+    apifyClient: ApifyClient,
+): Promise<User | null> {
+    // Hash token for cache key
+    const tokenHash = createHash('sha256').update(token).digest('hex');
+
+    // Check cache first
```
jirimoravcik:
```suggestion
```

src/mcp/server.ts
```diff
@@ -34,28 +34,76 @@ import {
...
 import { EXTERNAL_TOOL_CALL_TIMEOUT_MSEC, LOG_LEVEL_MAP } from './const.js';
 import { processParamsGetTools } from './utils.js';
 
 type ToolsChangedHandler = (toolNames: string[]) => void;
 
+/**
+ * Interface for storing and retrieving tool call counters per session.
+ * Used for tracking tool call sequence in user journeys.
+ */
+interface ToolCallCounterStore {
```
jirimoravcik: Shouldn't this be exported? I assume downstream users of the package would like to implement the interface

src/telemetry.ts
jirimoravcik: I doubt we can pass it from doppler as this is a public package that users can run locally.

src/mcp/server.ts
```diff
@@ -102,6 +154,44 @@ export class ActorsMcpServer {
...
+    private async getAndIncrementToolCallCounter(sessionId: string): Promise<number> {
+        if (this.options.toolCallCounterStore) {
+            // Use external store (Redis for HTTP/SSE)
+            return await this.options.toolCallCounterStore.getAndIncrement(sessionId);
+        }
+        // Use in-memory storage (for stdio)
+        const current = this.sessionToolCallCounters.get(sessionId) || 0;
+        const newValue = current + 1;
+        this.sessionToolCallCounters.set(sessionId, newValue);
+        return newValue;
```
jirimoravcik: Could we please refactor this into a default value for the `getAndIncrement`?
IMO this function should only do `return await this.options.toolCallCounterStore.getAndIncrement(sessionId)`

src/mcp/server.ts
```diff
@@ -666,12 +809,21 @@ Please verify the server URL is correct and accessible, and ensure you have a va
...
 Please verify the tool name, input parameters, and ensure all required resources are available.`,
                 ], true);
+            } finally {
+                // Track telemetry once at the end with determined status and execution time
+                if (telemetryData) {
+                    const execTime = Date.now() - startTime;
+                    telemetryData.tool_status = toolStatus;
+                    telemetryData.tool_exec_time_ms = execTime;
+                    trackToolCall(userId, getTelemetryEnv(this.options.telemetryEnv), telemetryData);
+                }
```
jirimoravcik: Maybe move to a separate function and just call that inside finally?

src/mcp/server.ts
```diff
@@ -102,6 +154,44 @@ export class ActorsMcpServer {
...
+        // Set telemetryEnv: explicit option > env var > default ('prod')
+        const telemetryEnv = process.env.TELEMETRY_ENV;
+        const telemetryParam = this.options.telemetryEnv;
+        this.options.telemetryEnv = getTelemetryEnv(telemetryParam ?? telemetryEnv);
+
+        // If telemetry is enabled, ensure telemetryEnv is set
+        if (this.options.telemetryEnabled && this.options.telemetryEnv === undefined) {
+            this.options.telemetryEnv = getTelemetryEnv(undefined);
+        }
+    }
```
jirimoravcik: (nit) This method is difficult to read for me

src/mcp/server.ts
```diff
@@ -263,20 +353,21 @@ export class ActorsMcpServer {
...
-                    // Clone the tool before modifying it to avoid affecting shared objects
-                    const clonedWrap = cloneToolEntry(wrap);
+                // Clone the tool before modifying it to avoid affecting shared objects
+                const clonedWrap = cloneToolEntry(wrap);
+                let modified = false;
 
+                // Handle Skyfire mode modifications
+                if (this.options.skyfireMode && (wrap.type === 'actor'
+                    || (wrap.type === 'internal' && wrap.name === HelperTools.ACTOR_CALL)
+                    || (wrap.type === 'internal' && wrap.name === HelperTools.ACTOR_OUTPUT_GET))) {
```
jirimoravcik, line 304: `wrap.type` is just a generic string, or is it some form of an enum?

src/mcp/server.ts
```diff
@@ -473,13 +562,17 @@ export class ActorsMcpServer {
             // eslint-disable-next-line prefer-const
             let { name, arguments: args, _meta: meta } = request.params;
             const { progressToken } = meta || {};
-            const apifyToken = (request.params.apifyToken || process.env.APIFY_TOKEN) as string;
+            const apifyToken = (request.params.apifyToken || this.options.token || process.env.APIFY_TOKEN) as string;
             const userRentedActorIds = request.params.userRentedActorIds as string[] | undefined;
+            // Injected for telemetry purposes
+            const mcpSessionId = request.params.mcpSessionId as string | undefined;
```
jirimoravcik: Why is the `as` cast needed here?

src/mcp/server.ts
```diff
@@ -548,6 +641,57 @@ Please check the tool's input schema using ${HelperTools.ACTOR_GET_DETAILS} tool
...
+                    mcp_protocol_version: params?.protocolVersion || '',
+                    mcp_capabilities: capabilities ? JSON.stringify(capabilities) : '',
+                    mcp_session_id: mcpSessionId || '',
+                    transport_type: this.options.transportType || '',
+                    tool_name: toolFullName,
+                    tool_status: 'succeeded', // Will be updated in finally
+                    tool_exec_time_ms: 0, // Will be calculated in finally
+                    tool_call_number: toolCallNumber,
+                };
+            }
```
jirimoravcik: This block of code would deserve a separate method

jirimoravcik (changes requested):
Looks good, I had some comments and questions. Let's make sure we don't introduce too much technical debt and complexity to just have segment analytics

src/actor/server.ts
jirispilka: I've already deleted this file in the other [PR](https://github.com/apify/apify-mcp-server/pull/338):

src/actor/server.ts
jirispilka: Removed here: https://github.com/apify/apify-mcp-server/pull/338

src/mcp/server.ts
jirispilka: If this package would be used by someone else then yes. 
But I don't think anyone is using that. 

Anyway, I moved the interfaces to types.ts and exported it

src/mcp/server.ts
jirispilka: I've not figured out a better way. I just renamed the variables

src/mcp/server.ts
jirispilka: `type` is enum here. It will be better to have `ToolType` with 'internal`, `actor`, and `actor-mcp`

I was about to change it but it would result into much larger PR, so I've created an issue [here](https://github.com/apify/apify-mcp-server/issues/339)

src/mcp/server.ts
jirispilka: Well, typescript complains there :) (I know, this is not the best answer but I didn't want to dig into that 😊 )

jirispilka: I gave it one more try, copilot is suggesting this:
```
 mcpSessionId = typeof request.params.mcpSessionId === 'string' ? request.params.mcpSessionId : undefined;
```

src/utils/user-cache.ts
jirispilka: I kept it as it is, i.e. using ttl-cache. We can revisit it later. There are more candidates for `memoize`.

I refactored the caching a bit based on the other comments

src/const.ts
jirispilka: oh, ok, I made it uppercase everywhere

tests/helpers.ts
jirispilka: This is in unit-tests. We should pass any value there, if it is defined.

But at other places it was wrong, like at server.ts. I fixed that.

src/utils/version.ts
jirispilka: This works
```
const require = createRequire(import.meta.url);
const packageJson = require('../../package.json');
...
        return packageJson.version || null;
````

`process.env.npm_package_version` - we don't always start it via npm

in the mcp-server-internal:
```
CMD [ "node", "--trace-warnings", "--stack-trace-limit=50", "./dist/main.js" ]
```

File: src/telemetry.ts (hunk 4 above)
Comment (jirispilka) [reply src/telemetry.ts:17]: Yeah, no doppler, sadly we need to keep it in the code, these variables were provided by Kuba R. (data)

File: src/mcp/server.ts (hunk 20 above)
Comment (jirispilka) [reply src/mcp/server.ts]: Thanks, changed to this:

```
    private async getAndIncrementToolCallCounter(sessionId: string): Promise<number> {
        return await this.options.telemetry!.toolCallCountStore!.getAndIncrement(sessionId);
    }
```

File: src/utils/user-cache.ts (hunk 16 above)
Comment (jirispilka) [reply src/utils/user-cache.ts]: There are more places that the memoize should be applied to. We can revisit it later.

File: src/utils/version.ts (hunk 14 above)
Comment (jirispilka) [reply src/utils/version.ts:10]: Yes, it works 

```
const require = createRequire(import.meta.url);
const packageJson = require('../../package.json');

packageJson.version
```

File: src/utils/version.ts (hunk 13 above)
Comment (jirispilka) [reply src/utils/version.ts]: Deleted

File: src/const.ts (hunk 12 above)
Comment (jirispilka) [reply src/const.ts:114]: Changed to uppercase

File: src/types.ts (hunk 7 above)
Comment (jirispilka) [reply src/types.ts]: Thanks! Changed to `app` and used `mcp` as a app name (the same as for mcp.apify.com)

File: tests/helpers.ts (hunk 10 above)
Comment (jirispilka) [reply tests/helpers.ts]: Fixed

File: src/telemetry.ts (hunk 5 above)
Comment (jirispilka) [reply src/telemetry.ts]: Thanks for pointing out `@apify-packages/analytics` I check it out. As Jirka M. said, we can't really use it here. But I've used it as an inspiration, at least

File: src/telemetry.ts (hunk 8 above)
Comment (jirispilka) [reply src/telemetry.ts]: oh, thanks! I changed it to `crypto.randomUUID()` then

File: src/telemetry.ts [hunk 26]
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/**
+ * Tracks a tool call event to Segment.
+ *
+ * @param userId - Apify user ID (null if not available)
+ * @param properties - Event properties for the tool call
+ */
+export function trackToolCall(
+    userId: string,
```
katzino: I have seen you using `crypto.randomUUID()` as a fallback but that should be attributes to `anonymousId` instead. In theory it shouldn't make much of a difference in mixpanel, but Segment is treating `userId` and `anonymousId` differently. I would maybe double check with @mr-rajce about it.

katzino (approved):
Pre-approving, just check out that one comment I have for `userId`

src/utils/userid-cache.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import type { ApifyClient } from '../apify-client.js';
+import { USER_CACHE_MAX_SIZE, USER_CACHE_TTL_SECS } from '../const.js';
+import { TTLLRUCache } from './ttl-lru.js';
+
+// LRU cache with TTL for user info - stores the raw User object from API
+const userIdCache = new TTLLRUCache<string>(USER_CACHE_MAX_SIZE, USER_CACHE_TTL_SECS);
+
+/**
+ * Gets user info from token, using cache to avoid repeated API calls
```
jirimoravcik:
```suggestion
 * Gets user ID from token, using cache to avoid repeated API calls
```

tests/unit/mcp.server.tool-call-counter.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+import { ActorsMcpServer } from '../../src/index.js';
+import type { ToolCallCounterStore } from '../../src/types.js';
+
+describe('ActorsMcpServer tool call counter', () => {
+    describe('default in-memory store', () => {
+        it('should increment counter correctly for multiple calls', async () => {
+            const server = new ActorsMcpServer({ setupSigintHandler: false });
+            // Default store should be available after construction
+            const store = server.options.telemetry!.toolCallCountStore!;
```
jirimoravcik: What does `should` mean? If we're certain, we can type it in a way that we don't need `!`. If we're not sure, we should do a check before accessing it.

src/types.ts
```diff
@@ -293,3 +293,81 @@ export type DatasetItem = Record<number | string, unknown>;
...
+    initializeRequestData?: InitializeRequest;
+    /**
+     * Telemetry configuration options.
+     */
+    telemetry?: {
+        /**
+         * Enable or disable telemetry tracking for tool calls.
+         * Defaults to true when not set.
+         */
+        enabled?: boolean;
```
jirimoravcik: Not sure if this should be optional, the parent object already is.

src/telemetry.ts
jirispilka: Good catch, I was not aware of that. I've changed it to
```
        client.track({
            ...(userId ? { userId } : { anonymousId: crypto.randomUUID() }),
            event: SEGMENT_EVENTS.TOOL_CALL,
            properties,
        });
````

File: src/telemetry.ts [hunk 30]
```diff
@@ -0,0 +1,83 @@
...
+    properties: ToolCallTelemetryProperties,
+): void {
+    const client = getOrInitAnalyticsClient();
+
+    if (!client) {
+        return;
+    }
+
+    try {
+        client.track({
```
jirimoravcik:
```suggestion
    try {
        client?.track({
```

src/telemetry.ts
```diff
@@ -0,0 +1,83 @@
...
+            return null;
+        }
+    }
+    return analyticsClient;
+}
+
+/**
+ * Tracks a tool call event to Segment.
+ *
+ * @param userId - Apify user ID (null if not available)
```
jirimoravcik, line 69: > `null` if not available

but the type is just `string`? Shouldn't it be `string | null`?

src/telemetry.ts
```diff
@@ -0,0 +1,83 @@
...
+export function getOrInitAnalyticsClient(): Analytics | null {
+    if (!analyticsClient) {
+        try {
+            const env = getTelemetryEnv(process.env.TELEMETRY_ENV);
+            const writeKey = env === TELEMETRY_ENV.PROD ? PROD_WRITE_KEY : DEV_WRITE_KEY;
+
+            analyticsClient = new Analytics({
+                writeKey,
+                flushAt: 50,
+                flushInterval: 5000,
```
jirimoravcik: Can we extract this into a constant that would mention the units? I'd assume it's `5_000` ms, but not sure from this part of code.

src/mcp/server.ts
```diff
@@ -68,9 +72,12 @@ export class ActorsMcpServer {
     private sigintHandler: (() => Promise<void>) | undefined;
     private currentLogLevel = 'info';
     public readonly options: ActorsMcpServerOptions;
+    // In-memory storage for tool call counters (used when toolCallCounterStore is not provided)
+    private sessionToolCallCounters = new Map<string, number>();
```
jirimoravcik: I think this can grow infinitely, we should implement some mechanism that would keep a limited number of items in the cache. Maybe a LRUCache would work?

src/mcp/server.ts
```diff
@@ -102,6 +110,57 @@ export class ActorsMcpServer {
         this.setupResourceHandlers();
     }
 
+    /**
+     * Gets and increments the tool call counter for a session.
+     * @param sessionId - The session ID
+     * @returns Promise resolving to the new counter value (after increment)
+     */
+    private async getAndIncrementToolCallCounter(sessionId: string): Promise<number> {
+        return await this.options.telemetry!.toolCallCountStore!.getAndIncrement(sessionId);
```
jirimoravcik: Can we type it so that we get rid of `!`?

jirimoravcik (approved):

src/telemetry.ts
jirispilka: Yeah, right, changed to:
```
                flushAt: SEGMENT_FLUSH_AT_EVENTS,
                flushInterval: SEGMENT_FLUSH_INTERVAL_MS,
```

src/mcp/server.ts
jirispilka: Good catch. In practice, this shouldn’t happen. When running locally, you usually create a single session, since stdio can’t handle multiple concurrent sessions.

For a remote server, you should provide your own `getAndIncrementToolCallCounter` with a Redis backend.
BUT if that function isn’t provided, the counter would grow indefinitely

So fixed using LRUCache.

src/telemetry.ts
jirispilka: Yeah, my mistake, I've done some changes and forgot to update it, it is `string | null` now (to handle `anonymousId` correctly)

src/mcp/server.ts
jirispilka: Well, this makes me think I should handle it in a better way (including telemetry envs). Previously, I was modifying input.options, but instead of doing that, I’ve now introduced vars
```
    private telemetryEnabled: boolean | null = null;
    private telemetryEnv: TelemetryEnv = DEFAULT_TELEMETRY_ENV;
    private toolCallCountStore: ToolCallCounterStore | undefined;
 ```
IMO this makes code cleaner, as I don't have to write `this.options.telemetry.env` etc.

File: tests/unit/mcp.server.tool-call-counter.test.ts (hunk 28 above)
Comment (jirispilka) [reply tests/unit/mcp.server.tool-call-counter.test.ts]: I've changed this part a bit more.  [See this comment](https://github.com/apify/apify-mcp-server/pull/329#discussion_r2559481058)

In nutshell, there is now `private toolCallCountStore: ToolCallCounterStore | undefined;` which will be defined only if telemetry is enabled.

But I had to change the tests to
```
            const toolCallCount = server.getToolCallCountStore();
            expect(toolCallCount).toBeDefined();

            if (!toolCallCount) {
                // TypeScript needs this for type narrowing
                expect.fail('toolCallCount should be defined');
            }
```

File: src/types.ts (hunk 29 above)
Comment (jirispilka) [reply src/types.ts]: Yeah, it might be better to be explicit here, so I marked it as required
```
         * Must be explicitly set when telemetry object is provided.
         * When telemetry object is omitted entirely, defaults to true (via env var or default).
         */
```

File: tests/unit/telemetry.test.ts
Comment (MQ37) [inline tests/unit/telemetry.test.ts:1]: note: I don't think we need tests like this. But let's keep that since it is already there.

File: src/mcp/server.ts [hunk 35]
```diff
@@ -69,8 +78,20 @@ export class ActorsMcpServer {
     private currentLogLevel = 'info';
     public readonly options: ActorsMcpServerOptions;
 
+    // Telemetry configuration (resolved from options and env vars in setupTelemetry)
+    private telemetryEnabled: boolean | null = null;
+    private telemetryEnv: TelemetryEnv = DEFAULT_TELEMETRY_ENV;
+    private toolCallCountStore: ToolCallCounterStore | undefined;
```
MQ37: question: Do we actually need this tool call counter logic? Why not just use the timestamp from the analytics that is included automatically and if we are just interested in the total session tool calls count I guess some mixpanel query could easily handle that.

tests/unit/telemetry.test.ts
jirispilka: I've removed the Actor code in other PR. I did not realize that you were relying on these test to for HTTP and SSE.

src/mcp/server.ts
jirispilka: Just for the record: I check again with Kuba R. and it is not actually required. 
Earlier it was proposed by Kuba to track user journey, so I'm going to remove it.

Thanks for pointing it out again!

PR #342, opened 2025-11-28
feat: Add server instructions

jirispilka:
- **feat: Initial server instructions**

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -161,21 +161,18 @@ These tools are called **Actors**. They enable you to extract structured data fr
 
 ## Tool dependencies and disambiguation
 
-### Mandatory dependencies
+### Tool dependencies
 - \`${HelperTools.ACTOR_CALL}\`:
   - First call with \`step="info"\` or use \`${HelperTools.ACTOR_GET_DETAILS}\` to obtain the Actor’s schema.
   - Then call with \`step="call"\` to execute the Actor.
-- \`${HelperTools.ACTOR_CALL}\` / Actor tools → \`${HelperTools.ACTOR_OUTPUT_GET}\`:
-  Use the \`datasetId\` from the Actor run to retrieve results.
-- \`${HelperTools.DOCS_SEARCH}\` → \`${HelperTools.DOCS_FETCH}\`:
-  Search returns URLs; fetch retrieves full content.
 
 ### Tool disambiguation
 - **${HelperTools.ACTOR_OUTPUT_GET} vs ${HelperTools.DATASET_GET_ITEMS}:**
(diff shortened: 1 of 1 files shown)


src/const.ts
```diff
@@ -132,3 +132,50 @@ export const DEFAULT_TELEMETRY_ENV: TelemetryEnv = TELEMETRY_ENV.PROD;
...
+- Actor execution may take time, and outputs can be large.
+- Large datasets can be paginated to retrieve results efficiently.
+
+## Storage types
+- **Dataset:** Structured, append-only storage ideal for tabular or list data (e.g., scraped items).
+- **Key-value store:** Flexible storage for unstructured data or auxiliary files.
+
+## Tool dependencies and disambiguation
+
+### Mandatory dependencies
```
MQ37: nit: I don't quite get what does "Mandatory dependencies" mean in this context? Maybe "Mandatory instructions"?

src/const.ts
```diff
@@ -132,3 +132,50 @@ export const DEFAULT_TELEMETRY_ENV: TelemetryEnv = TELEMETRY_ENV.PROD;
...
+- \`${HelperTools.ACTOR_CALL}\`:
+  - First call with \`step="info"\` or use \`${HelperTools.ACTOR_GET_DETAILS}\` to obtain the Actor’s schema.
+  - Then call with \`step="call"\` to execute the Actor.
+- \`${HelperTools.ACTOR_CALL}\` / Actor tools → \`${HelperTools.ACTOR_OUTPUT_GET}\`:
+  Use the \`datasetId\` from the Actor run to retrieve results.
```
MQ37: nit: I think the highlighted section is quite vague and I personally would be more word oriented instead of using the symbols like arrow or slash. Also I would provide exampel of the Actor tools like `apify-slash-rag-web-browser`

MQ37: Ahh I see the Actors tools example is given below, ok 😄

MQ37 (approved):
Just a few nits, otherwise LGTM 👍 I like how you included the section "Tool  disambiguation" that clearly separates the goals of each provided tool.

src/const.ts
jirispilka: Changed to Tool dependencies (if something dependent it means that is mandatory, I guess)

jirispilka: Also, I removed the two dependencies that were not really mandatory, `call-actor->get-output`, and also `search->fetch`

src/const.ts
jirispilka: right, agree, I've changed it a bit

PR #350, opened 2025-12-03
feat: Telemery add soft fail for client errors

jirispilka:
Add soft_fail status for user errors in tool telemetry.

Mark user/client errors (not found, invalid input, HTTP 4xx) as soft_fail in telemetry, while server errors (HTTP 5xx) remain failed. MCP responses unchanged (isError: true for all errors). 

I intentionally kept failed status for some tools and cases to track where LLMs are misled by tool descriptions, helping identify confusing tool descriptions

### DIFF
--- a/src/tools/fetch-apify-docs.ts
+++ b/src/tools/fetch-apify-docs.ts
@@ -43,9 +43,13 @@ USAGE EXAMPLES:
 
         // Only allow URLs starting with https://docs.apify.com
         if (!url.startsWith('https://docs.apify.com')) {
-            return buildMCPResponse([`Invalid URL: "${url}".
+            return buildMCPResponse({
+                texts: [`Invalid URL: "${url}".
 Only URLs starting with "https://docs.apify.com" are allowed.
-Please provide a valid Apify documentation URL. You can find documentation URLs using the ${HelperTools.DOCS_SEARCH} tool.`], true, TOOL_STATUS.SOFT_FAIL);
+Please provide a valid Apify documentation URL. You can find documentation URLs using the ${HelperTools.DOCS_SEARCH} tool.`],
+                isError: true,
+                toolStatus: TOOL_STATUS.SOFT_FAIL,
+            });
         }
 
         // Cache URL without fragment to avoid fetching the same page multiple times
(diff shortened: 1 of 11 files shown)


src/utils/mcp.ts
```diff
@@ -1,11 +1,22 @@
...
  * @param texts - Array of text strings to include in the response
- * @param isError - Optional flag to mark the response as an error (default: false)
+ * @param isError - Optional flag to mark the response as an error (default: false).
+ *                  This must remain MCP compliant: true for any tool-level error.
+ * @param toolStatus - Optional internal tool status used for telemetry. When provided,
+ *                     it will be attached as `_toolStatus` so the server can read it
+ *                     and strip it before sending the response to the MCP client.
  */
-export function buildMCPResponse(texts: string[], isError = false) {
+export function buildMCPResponse(
```
MQ37: nit: I would pass an object instead of adding additional positional parameter. I also think this will make the PR conflict easier to solve 😅

src/tools/build.ts
```diff
@@ -101,58 +93,3 @@ function truncateActorReadme(readme: string, limit = ACTOR_README_MAX_LENGTH): s
     const prunedReadme = lines.filter((line) => line.startsWith('#'));
     return `${readmeFirst}\n\nREADME was truncated because it was too long. Remaining headers:\n${prunedReadme.join(', ')}`;
 }
-
-const getActorDefinitionArgsSchema = z.object({
```
MQ37, line 105: cool that you got rid of this 😎

src/tools/fetch-apify-docs.ts
```diff
@@ -70,9 +76,14 @@ Please verify the URL is correct and accessible. You can search for available do
...
-                return buildMCPResponse([`Failed to fetch the documentation page at "${url}".
+                // Network/fetch errors are typically user errors (bad URL, connectivity issues)
+                return buildMCPResponse(
+                    [`Failed to fetch the documentation page at "${url}".
 Error: ${error instanceof Error ? error.message : String(error)}.
-Please verify the URL is correct and accessible. You can search for available documentation pages using the ${HelperTools.DOCS_SEARCH} tool.`], true);
+Please verify the URL is correct and accessible. You can search for available documentation pages using the ${HelperTools.DOCS_SEARCH} tool.`],
+                    true,
+                    TOOL_STATUS.SOFT_FAIL,
+                );
```
MQ37: this one is tricky but let's keep the soft fail 👍

jirispilka: Yeah, you are right .... ok, let's keep it soft fail and see

src/utils/mcp.ts
jirispilka: ok, I've updated the code to accept `options`

PR #358, opened 2025-12-09
fix: update AGENTS.md with mcp testing

jirispilka:


AGENTS.md
```diff
@@ -62,6 +62,30 @@ MANDATORY: Always check for TypeScript compilation errors before running tests o
 - Follow existing test patterns in the codebase
 - Ensure all tests pass before submitting a PR
 
+### Manual testing as MCP client
+
+You can manually test the MCP server by acting as an MCP client and invoking tools directly with real API calls.
+
+**How to test:**
+1. Connect to the MCP server (e.g., using `/mcp` command in Claude Code)
```
MQ37: Are these instructions for the human or an agent? Agent needs to be already connected to the MCP server, right? He cannot just run `/mcp` command and magically connect or can he?

khanbilal732 (approved):

AGENTS.md
jirispilka: Yeah, this is for the agent, I'll update it

PR #360, opened 2025-12-10
feat: long running tasks

MQ37:
**For PR reviewers:** I recommend reading the full commit messages as I was instructing agent to always read the changes and summarize the changes in the messages.

**Changes:**
- updated MCP SDK to latest version
- encountered issues with ZOD where it generated wrong schema for optional and default properties as they were marked as required - created PR into zod to fix this https://github.com/colinhacks/zod/pull/5532
- bumped node version to 24 (hopefully) everywheere
- implemented the long running tasks handlers - we are using in memory tasks store for stdio and Actor

- added async task execution, task are isolated by session id (another possibly way is to isolate them by user id which could be interesting use case). :warning: This is implemented only for the NORMAL Actor tools (no MCP servers), for internal tools it does not make sense. Currently the long running task tool call mode is optional and not enforced.

- in the internal repo implemented redis tasks store with multi node setup
	- wrote multi node tests that are passing
	- tasks result are stored by default for 5 minutes (we should revisit this in future). Maximum task TTL is capped at 10 minutes, clients can request more but they will get 10 minutes at maximum.

- TODO we should add long running tasks to telemetry, I would personally split that to another issue/PR since this is quite large already. Issue here https://github.com/apify/apify-mcp-server/issues/362
- for now I would keep the tasks related debug logs but we surely can remove them in future
- TODO we should also figure out how to handle the `call-actor` tool, since allowing long running tasks support there would cause even the `step=info` requests to run as a task which does not make sense - so we should probably remove this step and somehow enforce agent to use the fetch Actor details tool

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -565,6 +565,7 @@
       "integrity": "sha512-gzUt/qt81nXsFGKIFcC3YnfEAx5NkunCfnDlvuBSSFS02bcXu4Lmea0AFIUwbLWxWPx3d9p8S5QoaujKcNQxcQ==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "funding": {
         "url": "https://github.com/sponsors/colinhacks"
       }
@@ -7647,6 +7648,7 @@
       "integrity": "sha512-5gTmgEY/sqK6gFXLIsQNH19lWb4ebPDLA4SdLP7dsWkIXHWlG66oPuVvXSGFPppYZz8ZDZq0dYYrbHfBCVUb1Q==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=12"
       },
--- a/src/utils/tools.ts
+++ b/src/utils/tools.ts
@@ -14,6 +14,7 @@ export function getToolPublicFieldOnly(tool: ToolBase) {
         outputSchema: tool.outputSchema,
         annotations: tool.annotations,
         icons: tool.icons,
+        execution: tool.execution,
     };
 }
 
--- a/tests/integration/suite.ts
+++ b/tests/integration/suite.ts
@@ -253,6 +253,27 @@ export function createIntegrationTestsSuite(
             expect(names).not.toContain('fetch-apify-docs');
         });
 
+        it('should return tool with execution field when listing tools with apify/python-example', async () => {
+            const actors = [ACTOR_PYTHON_EXAMPLE];
+            client = await createClientFn({ tools: actors });
+            const tools = await client.listTools();
+
+            // Find the tool for apify/python-example
+            const pythonExampleTool = tools.tools.find((tool) => tool.name === actorNameToToolName(ACTOR_PYTHON_EXAMPLE));
+            expect(pythonExampleTool).toBeDefined();
+
+            // Verify the tool contains the execution field (as returned by getToolPublicFieldOnly)
+            expect(pythonExampleTool).toHaveProperty('execution');
+            expect(pythonExampleTool?.execution).toBeDefined();
+
+            // Verify other expected fields are present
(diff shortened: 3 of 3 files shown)


khanbilal732 (approved):

khanbilal732 (approved):

src/mcp/server.ts
```diff
@@ -483,6 +515,97 @@ export class ActorsMcpServer {
...
+            const task = await this.taskStore.getTask(taskId, mcpSessionId);
+            if (!task) {
+                // logging as this may not be just a soft fail but related to issue with the task store
+                log.error('[GetTaskRequestSchema] Task not found', { taskId, mcpSessionId });
+                throw new McpError(
+                    ErrorCode.InvalidParams,
+                    `Task "${taskId}" not found`,
+                );
+            }
+            return task;
```
jirispilka:
```suggestion
            if (task) return task;

            // logging as this may not be just a soft fail but related to issue with the task store
            log.error('[GetTaskRequestSchema] Task not found', { taskId, mcpSessionId });
            throw new McpError(ErrorCode.InvalidParams, `Task "${taskId}" not found`);
```

jirispilka: This would be better for readability

src/mcp/server.ts
```diff
@@ -483,6 +515,97 @@ export class ActorsMcpServer {
...
+                );
+            }
+            if (task.status !== 'completed' && task.status !== 'failed') {
+                throw new McpError(
+                    ErrorCode.InvalidParams,
+                    `Task "${taskId}" is not completed yet. Current status: ${task.status}`,
+                );
+            }
+            const result = await this.taskStore.getTaskResult(taskId, mcpSessionId);
+            return result;
```
jirispilka:
```suggestion
            return await this.taskStore.getTaskResult(taskId, mcpSessionId);
```

src/mcp/server.ts
```diff
@@ -93,7 +113,18 @@ export class ActorsMcpServer {
...
             {
                 capabilities: {
-                    tools: { listChanged: true },
+                    tools: {
+                        listChanged: true,
+                    },
+                    // Declare long running task support
+                    tasks: {
+                        list: {},
+                        requests: {
```
jirispilka, line 125:
```suggestion
                        list: {},
                        cancel: {},
                        requests: {
```

jirispilka: I believe the cancel should be in the capabilities

src/mcp/server.ts
MQ37: I prefer the current version, but I don't mind the suggested one.

src/mcp/server.ts
MQ37: good catch :+1:

MQ37:
> * We should use the task status from MCP (`@modelcontextprotocol/sdk/dist/esm/types.d.ts`) as we are repeating the strings: failed, working, etc at some may places

Will look into it.

> * The changes in tools, update of packages.json etc should have been a separate PR. It would make this PR more manageable

Agree, but it is already done.

> * Well, server.ts already had a lot of technical debt. Now it has nearly doubled. I'm thinking out loud here: If we are gonna merge this PR to release long-running tasks, in the coming sprints we are gonna need to ask for a couple of days for refactoring. I'm thinking about whether to postpone the release and do the refactoring right away, or accumulate the tech debt?

Honestly I would merge this technical debt and then create a separate PR for refactoring. This PR is already large and I would be against making it even larger and more complex. And yes, we definitely should reserve a few days for refactoring and giving the codebase some love.

jirispilka (approved):
We decided to merge this PR and simplify it in the coming PRs. Otherwise, this PR would be huge and difficult to review.

Issue #366, opened 2025-12-16
Tool search and fetch apify docs: we should also support Crawlee documentation


MQ37:
Currently the tools support only docs.apify.com since the Crawlee docs are in separate algolia index.

PR #369, opened 2025-12-18
feat: enhance documentation tools with Crawlee support

MQ37:
This add a new source input for the search docs tool that allows agent to choose from `apify`, `crawlee-py` and `crawlee-js` - defaults to `apify` for backward compatibility. Also adds `crawlee.dev` domain to allowed fetch docs tool domain list.

It was quite painful to get the Crawlee algolia working since it requires different args and filters and returns only url wihtout content snippet - but that is better than nothing.

Also tested locally with opencode and works quite well with just urls returned from the search.

closes https://github.com/apify/apify-mcp-server/issues/366

### DIFF
--- a/src/utils/apify-docs.ts
+++ b/src/utils/apify-docs.ts
@@ -134,28 +134,21 @@ export async function searchDocsBySource(
     const indexConfig = DOCS_SOURCES.find((idx) => idx.id === docSource);
 
     if (!indexConfig) {
-        const error = `Unknown documentation source: ${docSource}`;
-        log.error(`[Algolia] ${error}`);
-        throw new Error(error);
+        throw new Error(`Unknown documentation source: ${docSource}`);
     }
 
     const client = getAlgoliaClient(indexConfig.appId, indexConfig.apiKey);
 
-    try {
-        const searchRequest = prepareAlgoliaRequest(indexConfig, query);
-        const response = await client.search({
-            requests: [searchRequest],
-        });
+    const searchRequest = prepareAlgoliaRequest(indexConfig, query);
+    const response = await client.search({
+        requests: [searchRequest],
+    });
 
-        const results = response.results as unknown as AlgoliaResult[];
-        const searchResults = processAlgoliaResponse(results);
+    const results = response.results as unknown as AlgoliaResult[];
+    const searchResults = processAlgoliaResponse(results);
 
-        log.info(`[Algolia] Search completed successfully. Found ${searchResults.length} results for "${docSource}"`);
-        return searchResults;
-    } catch (error) {
-        log.error(`[Algolia] Search failed for "${docSource}" with query "${query}"`, error as Error);
-        throw error;
-    }
+    log.info(`[Algolia] Search completed successfully. Found ${searchResults.length} results for "${docSource}"`);
+    return searchResults;
 }
 
 /**
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -180,6 +180,21 @@ We use **4 spaces** for indentation (configured in `.editorconfig`).
 - **Internal Errors**: Use appropriate error codes (5xx for server errors), log with `log.exception` or `log.error`
 - Always handle and propagate errors clearly
 - Use custom error classes from `src/errors.ts` when appropriate
+- **Don't log then throw**: Do NOT call `log.error()` immediately before throwing. Errors are already logged by the caller or error handler. This creates duplicate logs and violates separation of concerns.
+  - ❌ Don't:
+    ```typescript
+    if (!indexConfig) {
+        const error = `Unknown documentation source: ${docSource}`;
+        log.error(`[Algolia] ${error}`);
+        throw new Error(error);
+    }
+    ```
+  - ✅ Do:
+    ```typescript
+    if (!indexConfig) {
+        throw new Error(`Unknown documentation source: ${docSource}`);
+    }
+    ```
 
 ### Code quality
 


src/const.ts
```diff
@@ -105,11 +105,51 @@ export const ACTOR_SEARCH_ABOVE_LIMIT = 50;
...
-    apiKey: 'e97714a64e2b4b8b8fe0b01cd8592870', // search only (public) API key
-    indexName: 'test_test_apify_sdk',
-};
+export const DOCS_SOURCES = [
+    {
+        id: 'apify',
+        label: 'Apify',
+        appId: 'N8EOCSBQGH',
+        apiKey: 'e97714a64e2b4b8b8fe0b01cd8592870',
+        indexName: 'test_test_apify_sdk',
```
jirispilka, line 114: What a nice index name :D (I know it is not a name you invented )

src/const.ts
```diff
@@ -105,11 +105,51 @@ export const ACTOR_SEARCH_ABOVE_LIMIT = 50;
...
+        supportsFragments: true,
+        filters: 'version:latest',
+        description:
+            'Apify Platform documentation including: Platform features, SDKs (JS, Python), CLI, '
```
jirispilka, line 117: Should we use template literals instead of string with `+`?

src/tools/search-apify-docs.ts
```diff
@@ -61,28 +83,50 @@ USAGE EXAMPLES:
...
+        // Validate docSource is one of the available indexes
+        if (!DOCS_SOURCES.find((idx) => idx.id === parsed.docSource)) {
+            const availableSources = DOCS_SOURCES.map((idx) => `"${idx.id}"`).join(', ');
+            log.softFail(`[search-apify-docs] Invalid docSource: "${parsed.docSource}". Available: ${availableSources}`);
+            return buildMCPResponse({
+                texts: [`Invalid documentation source: "${parsed.docSource}". Available sources: ${availableSources}.`],
+                isError: true,
+                toolStatus: TOOL_STATUS.SOFT_FAIL,
+            });
+        }
```
jirispilka: I believe that invalid `docSource` is caught earlier in zod validation, in `server.ts` so this check is not required and it is just make the code more complex

```
            log.debug('Validate arguments for tool', { toolName: tool.name, input: args });
            if (!tool.ajvValidate(args)) {
```

src/const.ts
```diff
@@ -105,11 +105,51 @@ export const ACTOR_SEARCH_ABOVE_LIMIT = 50;
 
 export const MCP_STREAMABLE_ENDPOINT = '/mcp';
 
-export const ALGOLIA = {
-    appId: 'N8EOCSBQGH',
-    apiKey: 'e97714a64e2b4b8b8fe0b01cd8592870', // search only (public) API key
-    indexName: 'test_test_apify_sdk',
-};
+export const DOCS_SOURCES = [
```
jirispilka, line 108: I would make type for DOCS_SOURCE, something like:
```
type ApifyDocsSource = {
    id: 'apify';
    label: string;
    appId: string;
    apiKey: string;
    indexName: string;
    supportsFragments: true;
    filters: string;
    description: string;
};

type CrawleeDocsSource = {
    id: 'crawlee-js' | 'crawlee-py';
    label: string;
    appId: string;
    apiKey: string;
    indexName: string;
    supportsFragments: false;
    typeFilter: string;
    facetFilters: Array<string | string[]>;
    description: string;
};

type DocsSource = ApifyDocsSource | CrawleeDocsSource;

export const DOCS_SOURCES: readonly DocsSource[] = [
    // ... your sources
] as const;
```

To help typescript to catch issues

.gitignore
```diff
@@ -35,3 +35,5 @@ __pycache__/
 *.pyo
 *.pyd
 .Python
+
+mcp/
```
jirispilka, line 39: Why is thi here?

src/utils/apify-docs.ts
```diff
@@ -33,65 +44,141 @@ type AlgoliaResult = {
...
-    const results = response.results as unknown as AlgoliaResult[];
-
-    const searchResults: ApifyDocsSearchResult[] = [];
-    for (const result of results) {
-        if (result.hits && result.hits.length > 0) {
-            for (const hit of result.hits) {
-                // Check the fields, just in case
-                if (!hit.url_without_anchor || !hit.content) {
-                    continue; // Skip hits with missing fields
+export async function searchDocsBySource(
```
jirispilka, line 130: it would be nice to simplify this function, ie. create two extra functions? (or at least one please :))
- prepareRequest
- processResponse

src/utils/apify-docs.ts
```diff
@@ -33,65 +44,141 @@ type AlgoliaResult = {
...
+        }
+
+        const response = await client.search({
+            requests: [searchRequest],
+        });
+
+        const results = response.results as unknown as AlgoliaResult[];
+        const searchResults: ApifyDocsSearchResult[] = [];
+
+        for (const result of results) {
```
jirispilka: The way that you process results is quite difficult to understand for me. Can we try to simplify it? Or explain with data?

.gitignore
MQ37: it's my dir local testing dir that contains opencode config

src/const.ts
MQ37: I would enforce this via linter :+1:

src/tools/search-apify-docs.ts
MQ37: true 👍

src/utils/apify-docs.ts
MQ37: yeah, it is too large :+1:

src/utils/apify-docs.ts
MQ37: I simplified it, also moved the fragment (url part after hashtag) to URL instead of keeping a separate field.

src/const.ts
MQ37: I would not add this since we are not touching this part regularly and I think the added lines of code are not worth it now.

.gitignore
jirispilka: oh, ok 👍🏻

src/const.ts
jirispilka: ok, make sense

Issue #261, opened 2025-09-04
Fix flaky cancellable Actor run test


MQ37:
Context:
```
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL  tests/integration/actor.server-streamable.test.ts > Apify MCP Server Streamable HTTP > should abort actor run on notifications/cancelled
Error: Timed out in waitUntil!
 ❯ tests/integration/suite.ts:653:22
    651| 
    652|             // Poll up to 30s for the latest run for this actor to reach ABORTED/ABORTING
    653|             await vi.waitUntil(async () => {
       |                      ^
    654|                 const runsList = await api.runs().list({ limit: 5, desc: true });
    655|                 const run = runsList.items.find((r) => r.actId === actId);

```

PR #370, opened 2025-12-18
fix: increase timeout for abort tests to reduce flakiness

MQ37:
closes https://github.com/apify/apify-mcp-server/issues/261

### DIFF
--- a/tests/integration/suite.ts
+++ b/tests/integration/suite.ts
@@ -847,7 +847,7 @@ export function createIntegrationTestsSuite(
                 .catch(() => undefined);
 
             // Abort right away
-            setTimeout(() => controller.abort(), 1000);
+            setTimeout(() => controller.abort(), 3000);
 
             // Ensure the request completes/cancels before proceeding
             await requestPromise;
@@ -866,7 +866,7 @@ export function createIntegrationTestsSuite(
                     return run.status === 'ABORTED' || run.status === 'ABORTING';
                 }
                 return false;
-            }, { timeout: 3000, interval: 500 });
+            }, { timeout: 10000, interval: 500 });
         });
 
         // Cancellation test using call-actor tool: start a long-running actor via call-actor and cancel immediately, then verify it was aborted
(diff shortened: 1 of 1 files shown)


tests/integration/suite.ts
```diff
@@ -847,7 +847,7 @@ export function createIntegrationTestsSuite(
                 .catch(() => undefined);
 
             // Abort right away
-            setTimeout(() => controller.abort(), 1000);
+            setTimeout(() => controller.abort(), 3000);
```
jirispilka, line 850: I don’t think increasing this timeout will help. It would only delay when abort is called.
IMO we could call abort immediately after start (unless there’s a window where this wouldn’t cancel the Actor run, but I don't think so :) ).

tests/integration/suite.ts
```diff
@@ -866,7 +866,7 @@ export function createIntegrationTestsSuite(
                     return run.status === 'ABORTED' || run.status === 'ABORTING';
                 }
                 return false;
-            }, { timeout: 3000, interval: 500 });
+            }, { timeout: 10000, interval: 500 });
```
jirispilka, line 869: This might help actually 👍🏻

tests/integration/suite.ts
MQ37: The issue was the abort was called imidiatelly after the Actor was started and it did not receive or ignored the abort call - it was not in the run logs. I tested multiple times and it did not time out so it is probably working - at least it is better than before.

tests/integration/suite.ts
MQ37: from my testing both were needed.

tests/integration/suite.ts
jirispilka: Interesting, this would indicate some problem at platform level 🤷🏻

MQ37: I think the problem is general inconsistency / eventual consistency problem, this is known behavior for example with Actor results in dataset, sometimes it takes really long time - few seconds, sometimes tens of seconds, to be able to retrieve the dataset results.

Issue #363, opened 2025-12-15
Failed to connect to Actorized MCP server


jirispilka:
```
{"cause":{"code":500.0,"event":{"code":500.0,"message":"Non-200 status code (500)"},"message":"SSE error: Non-200 status code (500)","name":"Error","stack":"Error: SSE error: Non-200 status code (500)\n    at _eventSource.onerror (file:///home/node/node_modules/@modelcontextprotocol/sdk/dist/esm/client/sse.js:82:31)\n    at EventSource.failConnection_fn (file:///home/node/node_modules/eventso...[truncated]"},"exception":{"code":500.0,"event":{"code":500.0,"message":"Non-200 status code (500)"},"message":"SSE error: Non-200 status code (500)","name":"Error","stack":"Error: SSE error: Non-200 status code (500)\n    at _eventSource.onerror (file:///home/node/node_modules/@modelcontextprotocol/sdk/dist/esm/client/sse.js:82:31)\n    at EventSource.failConnection_fn (file:///home/node/node_modules/eventsource/dist/index.js:234:53)\n    at file:///home/node/node_modules/eventsource/dist/index.js:76:74\n    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)"},"msg":"Failed to connect to MCP server using SSE transport","statusCode":500.0,"url":"https://T9RYyAv3bUWljCEoq.apify.actor/sse"}
```

Report from Mara in https://apify.slack.com/archives/C0L33UM7Z/p1766133036096699:
```
{"actorFullName":"jiri.spilka/weather-mcp-server","actorId":"J3RfhVsm4cNFK9Jih","error":{"message":"Error POSTing to endpoint (HTTP 400): Bad Request: No valid session ID provided","name":"Error","stack":"Error: Error POSTing to endpoint (HTTP 400): Bad Request: No valid session ID provided\n    at StreamableHTTPClientTransport.send (file:///home/node/node_modules/@modelcontextprotocol/sdk/dist/esm/client/streamableHttp.js:284:23)\n    at ...[truncated]"},"msg":"Failed to connect to MCP server"}
```

jirispilka:
> This error happens when we are trying to connect to an MCP Actor that does not have standby mode enabled but has the `webServerMcpPath`.

I saw it also for whether-mcp IIRC

MQ37:
[...]

PR #372, opened 2025-12-19
fix: update actor definition handling to include full metadata and improve MCP server identification

MQ37:
related to https://github.com/apify/apify-mcp-server/issues/363

The issue was thet if Actor had the mcp path in the `actor.json` it was treated as MCP server even though it was not standby Actor. This fixes that so the MCP server is considere only Actors that have both the path in json and also have standby mode enabled, all other Actors are considered as normal.

### DIFF
--- a/src/tools/actor.ts
+++ b/src/tools/actor.ts
@@ -314,7 +314,7 @@ export async function getActorsAsTools(
 
     const clonedActors = structuredClone(actorsInfo);
 
-    // Filter out nulls
+    // Filter out nulls - actorInfo can be null if the Actor was not found or an error occurred
     const nonNullActors = clonedActors.filter((actorInfo): actorInfo is ActorInfo => Boolean(actorInfo));
 
     // Separate Actors with MCP servers and normal Actors


src/tools/actor.ts
```diff
@@ -150,7 +150,7 @@ export async function callActorGetDataset(
  * 4. Properties are shortened using shortenProperties()
  * 5. Enums are added to descriptions with examples using addEnumsToDescriptionsWithExamples()
  *
- * @param {ActorInfo[]} actorsInfo - An array of ActorInfo objects with webServerMcpPath and actorDefinitionPruned.
+ * @param {ActorInfo[]} actorsInfo - An array of ActorInfo objects with webServerMcpPath, definition, and actor.
```
jirispilka:
```suggestion
 * @param {ActorInfo[]} actorsInfo - An array of ActorInfo objects with webServerMcpPath, definition, and Actor.
```

src/tools/actor.ts
```diff
@@ -159,22 +159,22 @@ export async function getNormalActorsAsTools(
     const tools: ToolEntry[] = [];
 
     for (const actorInfo of actorsInfo) {
-        const { actorDefinitionPruned } = actorInfo;
+        const { definition } = actorInfo;
```
jirispilka, line 162: Nice, the variable name was too long

src/tools/build.ts
```diff
@@ -13,13 +14,13 @@ import type {
  * @param {string} actorIdOrName - Actor ID or Actor full name.
  * @param {ApifyClient} apifyClient - The Apify client instance.
  * @param {number} limit - Truncate the README to this limit.
- * @returns {Promise<ActorDefinitionWithDesc | null>} - The actor definition with description or null if not found.
+ * @returns {Promise<ActorDefinitionWithInfo | null>} - The actor definition with info or null if not found.
```
jirispilka:
```suggestion
 * @returns {Promise<ActorDefinitionWithInfo | null>} - The Actor definition with info or null if not found.
```

src/tools/actor.ts
```diff
@@ -312,9 +314,14 @@ export async function getActorsAsTools(
 
     const clonedActors = structuredClone(actorsInfo);
 
-    // Filter out nulls and separate Actors with MCP servers and normal Actors
-    const actorMCPServersInfo = clonedActors.filter((actorInfo) => actorInfo && actorInfo.webServerMcpPath) as ActorInfo[];
-    const normalActorsInfo = clonedActors.filter((actorInfo) => actorInfo && !actorInfo.webServerMcpPath) as ActorInfo[];
+    // Filter out nulls
+    const nonNullActors = clonedActors.filter((actorInfo): actorInfo is ActorInfo => Boolean(actorInfo));
```
jirispilka, line 318: Is it possible to have actorInfo null? If yes, please update the comment when this happens

src/utils/actor.ts
```diff
@@ -21,10 +21,11 @@ export async function getActorMcpUrlCached(
...
 
     try {
-        const actorDefinitionPruned = await getActorDefinition(actorIdOrName, apifyClient);
-        const mcpPath = actorDefinitionPruned && getActorMCPServerPath(actorDefinitionPruned);
-        if (actorDefinitionPruned && mcpPath) {
-            const url = await getActorMCPServerURL(actorDefinitionPruned.id, mcpPath);
+        const actorDefinitionWithInfo = await getActorDefinition(actorIdOrName, apifyClient);
+        const definition = actorDefinitionWithInfo?.definition;
+        const mcpPath = definition && getActorMCPServerPath(definition);
+        if (definition && mcpPath) {
```
jirispilka:
```suggestion
        if (mcpPath) {
```

jirispilka: You can remove `defintion` because `mcpPath` is  truthy when definition exists (due to the `&&` in the first line)

src/tools/actor.ts
MQ37: we should either write a tooling for this or not care at all except readme and public documents :sweat_smile:

jirispilka: it should be in Agents.md :). By having it everywhere using capital `A` we are letting our coding agents know how it must look like :)

src/tools/actor.ts
MQ37: Yes, check out the `.map()` 3 SLOCs above, when the fetch Actor info fails it can be `null`.

jirispilka: oh, I see 🤦🏻

PR #375, opened 2025-12-22
feat(gpt-apps): add tools and widget descriptors

jakcinmarina:
**Summary**

This PR introduces the initial GPT Apps backend/tools layer, including new and updated tools for running actors, along with their widget descriptors used by the GPT Apps UI.

**Changes**

UI mode:
- added support for UI widgets via `ui` query parameter
- if value is set to `?ui=openai` it will show UI in following tools: 
  - new async `call-actor-widget` tool with UI (actor-widget)
  - new tool `fetch-actor-details-widget` to fetch Actor details with UI
  - modified `search-actors` tool to return UI if uiMode is set to `openai`, if param doesn't exist it will return previously existing output

**NOTE**
Widget descriptors are included for early review, even though the UI that renders them is not yet part of this PR

Next steps
- Add the web scaffold and shared UI components (separate PR incoming)
- Introduce the actual widget/page implementations (separate PR incoming)
- Address sync vs async execution UX once the full stack is in place

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -218,8 +218,8 @@ These tools are called **Actors**. They enable you to extract structured data fr
   - First call with \`step="info"\` or use \`${HelperTools.ACTOR_GET_DETAILS}\` to obtain the Actor's schema.
   - Then call with \`step="call"\` to execute the Actor.
   - When \`step="call"\`, supports async execution via the \`async\` parameter:
-    - When \`async: false\` or not provided (default when UI mode is disabled): Waits for completion and returns results immediately.
-    - When \`async: true\` (default when UI mode is enabled): Starts the run and returns immediately with runId. Use \`${HelperTools.ACTOR_RUNS_GET}\` to check status and retrieve results.
+  - When \`async: false\` or not provided (default when UI mode is disabled): Waits for completion and returns results immediately.
+  - When \`async: true\` (default when UI mode is enabled): Starts the run and returns immediately with runId. Use \`${HelperTools.ACTOR_RUNS_GET}\` to check status and retrieve results.
 
 ### Tool disambiguation
 - **${HelperTools.ACTOR_OUTPUT_GET} vs ${HelperTools.DATASET_GET_ITEMS}:**
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -309,7 +309,7 @@ export class ActorsMcpServer {
      * Used primarily for SSE.
      */
     public async loadToolsFromUrl(url: string, apifyClient: ApifyClient) {
-        const tools = await processParamsGetTools(url, apifyClient);
+        const tools = await processParamsGetTools(url, apifyClient, this.options.uiMode);
         if (tools.length > 0) {
             log.debug('Loading tools from query parameters');
             this.upsertTools(tools, false);
--- a/src/tools/actor.ts
+++ b/src/tools/actor.ts
@@ -402,6 +402,9 @@ EXAMPLES:
         // Additional props true to allow skyfire-pay-id
         additionalProperties: true,
     }),
+    _meta: {
+        ...getWidgetConfig(WIDGET_URIS.ACTOR_RUN)?.meta,
+    },
     annotations: {
         title: 'Call Actor',
         readOnlyHint: false,
--- a/src/tools/run.ts
+++ b/src/tools/run.ts
@@ -41,6 +41,9 @@ USAGE EXAMPLES:
 - user_input: What is the datasetId for run y2h7sK3Wc?`,
     inputSchema: z.toJSONSchema(getActorRunArgs) as ToolInputSchema,
     ajvValidate: compileSchema(z.toJSONSchema(getActorRunArgs)),
+    _meta: {
+        ...getWidgetConfig(WIDGET_URIS.ACTOR_RUN)?.meta,
+    },
     annotations: {
         title: 'Get Actor run',
         readOnlyHint: true,
--- a/tests/integration/suite.ts
+++ b/tests/integration/suite.ts
@@ -145,12 +145,14 @@ export function createIntegrationTestsSuite(
         it('should list all default tools and Actors', async () => {
             client = await createClientFn();
             const tools = await client.listTools();
-            expect(tools.tools.length).toEqual(defaultTools.length + defaults.actors.length + 1);
+            expect(tools.tools.length).toEqual(defaultTools.length + defaults.actors.length + 2);
 
             const names = getToolNames(tools);
             expectToolNamesToContain(names, DEFAULT_TOOL_NAMES);
             expectToolNamesToContain(names, DEFAULT_ACTOR_NAMES);
-            expect(names).toContain('get-actor-output');
+            expect(names).toContain(HelperTools.ACTOR_OUTPUT_GET);
+            // get-actor-run should be automatically included when call-actor is present
+            expect(names).toContain(HelperTools.ACTOR_RUNS_GET);
             await client.close();
         });
 
@@ -165,12 +167,14 @@ export function createIntegrationTestsSuite(
             const expectedActors = ['apify-slash-rag-web-browser'];
 
             const expectedTotal = expectedActorsTools.concat(expectedDocsTools, expectedActors);
-            expect(names).toHaveLength(expectedTotal.length + 1);
+            expect(names).toHaveLength(expectedTotal.length + 2);
 
             expectToolNamesToContain(names, expectedActorsTools);
             expectToolNamesToContain(names, expectedDocsTools);
             expectToolNamesToContain(names, expectedActors);
-            expect(names).toContain('get-actor-output');
+            expect(names).toContain(HelperTools.ACTOR_OUTPUT_GET);
+            // get-actor-run should be automatically included when call-actor is present
+            expect(names).toContain(HelperTools.ACTOR_RUNS_GET);
 
             await client.close();
         });
@@ -204,11 +208,13 @@ export function createIntegrationTestsSuite(
         it('should list all default tools and Actors when enableAddingActors is false', async () => {
             client = await createClientFn({ enableAddingActors: false });
             const names = getToolNames(await client.listTools());
-            expect(names.length).toEqual(defaultTools.length + defaults.actors.length + 1);
+            expect(names.length).toEqual(defaultTools.length + defaults.actors.length + 2);
 
             expectToolNamesToContain(names, DEFAULT_TOOL_NAMES);
             expectToolNamesToContain(names, DEFAULT_ACTOR_NAMES);
-            expect(names).toContain('get-actor-output');
+            expect(names).toContain(HelperTools.ACTOR_OUTPUT_GET);
+            // get-actor-run should be automatically included when call-actor is present
+            expect(names).toContain(HelperTools.ACTOR_RUNS_GET);
 
             await client.close();
         });
@@ -412,9 +418,11 @@ export function createIntegrationTestsSuite(
             const selectedToolName = actorNameToToolName(ACTOR_PYTHON_EXAMPLE);
             client = await createClientFn({ enableAddingActors: true, tools: ['actors'] });
             const names = getToolNames(await client.listTools());
-            // Only the actors category, get-actor-output and add-actor should be loaded
-            const numberOfTools = toolCategories.actors.length + 2;
+            // Only the actors category, get-actor-output, get-actor-run, and add-actor should be loaded
+            const numberOfTools = toolCategories.actors.length + 3;
             expect(names).toHaveLength(numberOfTools);
+            // get-actor-run should be automatically included when call-actor is present
+            expect(names).toContain(HelperTools.ACTOR_RUNS_GET);
             // Check that the Actor is not in the tools list
             expect(names).not.toContain(selectedToolName);
 
@@ -1038,11 +1046,13 @@ export function createIntegrationTestsSuite(
             // Test with enableAddingActors = false via env var
             client = await createClientFn({ enableAddingActors: false, useEnv: true });
             const names = getToolNames(await client.listTools());
-            expect(names.length).toEqual(defaultTools.length + defaults.actors.length + 1);
+            expect(names.length).toEqual(defaultTools.length + defaults.actors.length + 2);
 
             expectToolNamesToContain(names, DEFAULT_TOOL_NAMES);
             expectToolNamesToContain(names, DEFAULT_ACTOR_NAMES);
-            expect(names).toContain('get-actor-output');
+            expect(names).toContain(HelperTools.ACTOR_OUTPUT_GET);
+            // get-actor-run should be automatically included when call-actor is present
+            expect(names).toContain(HelperTools.ACTOR_RUNS_GET);
 
             await client.close();
         });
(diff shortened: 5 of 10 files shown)


tests/integration/suite.ts
```diff
@@ -1414,5 +1414,12 @@ export function createIntegrationTestsSuite(
                 }
             }
         });
+
+        it.runIf(options.transport === 'stdio')('should use UI_MODE env var when CLI arg is not provided', async () => {
+            client = await createClientFn({ useEnv: true, uiMode: 'openai' });
+            const tools = await client.listTools();
+            expect(tools.tools.length).toBeGreaterThan(0);
+            await client.close();
+        });
```
MQ37, line 1574: question: does this test case test anything related to the ui mode? It just list tools and check if the tool list is not empty. I think it should at least check some meta fields related to openai ui mode.

src/const.ts
```diff
@@ -44,6 +44,9 @@ export enum HelperTools {
     DOCS_SEARCH = 'search-apify-docs',
     DOCS_FETCH = 'fetch-apify-docs',
     GET_HTML_SKELETON = 'get-html-skeleton',
+    CALL_ACTOR_WIDGET = 'call-actor-widget',
+    GET_ACTOR_RUN_STATUS = 'get-actor-run-status',
+    FETCH_ACTOR_DETAILS_WIDGET = 'fetch-actor-details-widget',
```
MQ37: change: I would use the already existing fetch-actor-details tool and branch the logic based on the uiMode option provided instead of a new tool.

src/const.ts
```diff
@@ -44,6 +44,9 @@ export enum HelperTools {
     DOCS_SEARCH = 'search-apify-docs',
     DOCS_FETCH = 'fetch-apify-docs',
     GET_HTML_SKELETON = 'get-html-skeleton',
+    CALL_ACTOR_WIDGET = 'call-actor-widget',
```
MQ37: change: I would do the same with this tool but I understand that the current state of the call-actor tool is complicated with the step=info and step=call so I think this is fine to keep separate for now. I would just add a logic that if the server is in ?uiMode then it automatically uses call-actor-widget tool instead of the call-actor tool so we don't need to update the server instructions and confuse the agent. So the logic would be if client provides ?tools=call-actor or ?tools=actors when also ?uiMode=openai then it automatically swaps call-actor tool for the widget one.

src/const.ts
```diff
@@ -227,4 +230,6 @@ These tools are called **Actors**. They enable you to extract structured data fr
   \`${HelperTools.STORE_SEARCH}\` finds robust and reliable Actors for specific websites; ${RAG_WEB_BROWSER} is a general and versatile web scraping tool.
 - **Dedicated Actor tools (e.g. ${RAG_WEB_BROWSER}) vs ${HelperTools.ACTOR_CALL}:**
   Prefer dedicated tools when available; use \`${HelperTools.ACTOR_CALL}\` only when no specialized tool exists in Apify store.
+- **Async vs sync Actor tools (${HelperTools.ACTOR_CALL} vs ${HelperTools.CALL_ACTOR_WIDGET}):**
+  Default to \`${HelperTools.ACTOR_CALL}\` (synchronous, no widget) when the user asks to “run/call” and does not request background/progress/UI. Use \`${HelperTools.CALL_ACTOR_WIDGET}\` only when the user wants background/progress/UI. After starting an async run and obtaining runId, do NOT start another async run—only poll with \`${HelperTools.GET_ACTOR_RUN_STATUS}\` using that runId.
```
MQ37: question: can the agent event decide what tool to use and when? Does the chatgpt agent/LLM have this information in the context that in runs with the ui mode or not?

src/const.ts
MQ37: change: I would rename this tool to call-actor-async and I would make it more general than strictly widget / ui oriented meaning I would change the tool description so It can be used generally. Then I would also change the tool loading logic for ?uiMode=openai (can be done in tool-loader.ts) so that the call-actor tool is automatically swapped in all cases for call-actor-async (this tool) when in ui mode so we don't need to do all the complicated instructions and every Actor call in the ui mode is async by default which I think is ok and simplifies the solution. What do you think? So this means that with this tool the get-actor-run-status one would have to also be present otherwise it would not work so the logic would need to reflect this.

src/tools/actor.ts
```diff
@@ -358,6 +358,22 @@ export const callActor: ToolEntry = {
...
+- Quick-running Actors where waiting is acceptable
+- User doesn't mention "start", "run in background", "monitor progress", "widget", or "UI"
+
+**WHEN NOT TO USE THIS TOOL:**
+- User explicitly wants to "start" a run or mentions "background" or "async"
+- Long-running Actors where waiting would timeout
+- User wants to monitor progress in a UI widget
+→ In these cases, use ${HelperTools.CALL_ACTOR_WIDGET} instead
+
+This tool stays synchronous: it waits for completion and returns the results in the response. Do not pair it with ${HelperTools.CALL_ACTOR_WIDGET} for the same task.
```
MQ37: change: If we decide to go with the tool swapping logic for the call-actor-async tool I would remove these changes since the description is already too long and sometimes the agent/LLM does not respect that because of the length.

src/tools/actor.ts
```diff
@@ -393,6 +410,18 @@ EXAMPLES:
...
+        'openai/toolInvocation/invoked': 'Actor run finished (sync)',
+        'openai/widgetAccessible': false,
+        'openai/resultCanProduceWidget': false,
+        // TODO: replace with real CSP domains
+        'openai/widgetCSP': {
+            connect_domains: ['https://api.example.com'],
+            resource_domains: ['https://persistent.oaistatic.com'],
+        },
+        'openai/widgetDomain': 'https://chatgpt.com',
+    },
```
MQ37: change: if we decide for the tool swapping logic we should remove this.

src/tools/actor.ts
```diff
@@ -372,6 +388,7 @@ To fetch the full output, use the ${HelperTools.ACTOR_OUTPUT_GET} tool with the
 USAGE:
 - Always use dedicated tools when available (e.g., ${actorNameToToolName('apify/rag-web-browser')})
 - Use the generic call-actor tool only if a dedicated tool does not exist for your Actor.
+- PREFER this tool when user wants immediate results
```
MQ37: change: same here

src/mcp/server.ts
```diff
@@ -430,43 +430,160 @@ export class ActorsMcpServer {
...
+                    description: 'Interactive Actor run widget',
+                    mimeType: 'text/html+skybridge',
+                    _meta: {
+                        'openai/outputTemplate': 'ui://widget/actor-run.html',
+                        'openai/widgetAccessible': true,
+                        'openai/resultCanProduceWidget': true,
+                        // TODO: replace with real CSP domains
+                        'openai/widgetCSP': {
+                            connect_domains: ['https://api.example.com'],
+                            resource_domains: ['https://persistent.oaistatic.com'],
```
MQ37: it will be hosted at `https://mcp.apify.com` where the remote version of the Apify MCP server is.

src/mcp/server.ts
```diff
@@ -430,43 +430,160 @@ export class ActorsMcpServer {
 
     private setupResourceHandlers(): void {
         this.server.setRequestHandler(ListResourcesRequestSchema, async () => {
+            const resources = [];
```
jirispilka, line 439: Just a note, no action needed. The mcp/server.ts is super loong and we'll need to refactor it anyway. We'll add a new dir with resources.ts to make it manageable.

jirispilka (commented):
Thank you, I can see you have put a lot of work into this 💪🏻 . I appreciate it!

I reviewed it only partially, the main concern is about the new tools.
Do we really need them? :) 

`getActorRunStatus` – we don't need IMO, we have getActorRun already. We can:
- Always return structured content (not raw JSON)
- Conditionally include widget metadata when uiMode === 'openai'`

Similarly for `call-actor` vs `call-actor-widget`: I’ would prefer a single `call-actor` tool that:
- supports both sync and async modes (e.g. mode: "sync" | "async", we can over-ride defaults based on `uiMode`)
- note that two-step (info, call) flow is gonna be removed in the next PR by @MQ37

That way we don't multiply tools just for UI concerns.
Ideally, I would like to leverage the MCP tasks (all the Actor calls are async). Client starts task and polls for results.

But I'm afraid this won't be supported by OpenAI anytime soon.

Let's discuss it on a call, I'll discuss with @drobnikj to move this forward.

src/mcp/server.ts
jakcinmarina: just to check - both connect and resource domain will be https://mcp.apify.com?

src/const.ts
jakcinmarina: yes, there was some discussion to have just one tool (the comment from @jirispilka below). I am just concerned about the new implementation of `call-actor` tool. From what I understood there will be some changes where 2 step call will be gone, but not sure if that means the mcp tasks support. 
Either way it will be branched  based on the uiMode para so it's fine, I just want to confirm the exact preferred solution. cc @jirispilka

tests/integration/suite.ts
jakcinmarina: yes i should definitely do a better check :)

src/mcp/server.ts
MQ37: Based on their docs the `resource_domains` will be `https://mcp.apify.com`, that is where the mcp server runs. I am not sure about `connect_domains`, if the widget will access Apify API we might need to add `https://api.apify.com` there and I am not sure if the `https://mcp.apify.com` is needed here.

src/const.ts
MQ37: The PR is already here, I reworked the `call-actor` tool and added the mcp task support also for this tool - note that there may be some minor changes in the PR until merged. https://github.com/apify/apify-mcp-server/pull/387

src/utils/actor-card.ts
```diff
@@ -184,3 +184,87 @@ export function formatActorToStructuredCard(
...
+    };
+    currentPricingInfo: {
+        pricingModel: string;
+        pricePerResultUsd: number;
+        monthlyChargeUsd: number;
+    };
+    userActorRuns: {
+        successRate: number | null;
+    };
+} {
```
MQ37, line 219: nit: make the return separate type in types.ts

src/utils/tools.ts
```diff
@@ -1,11 +1,37 @@
 import { toolCategories } from '../tools/index.js';
-import type { HelperTool, ToolBase, ToolCategory, ToolEntry } from '../types.js';
+import type { ActorsMcpServerOptions, HelperTool, ToolBase, ToolCategory, ToolEntry } from '../types.js';
+
+type ToolPublicFieldOptions = {
+    uiMode?: ActorsMcpServerOptions['uiMode'];
+    filterOpenAiMeta?: boolean;
+};
```
MQ37, line 7: nit: move to types.ts

src/tools/actor.ts
```diff
@@ -572,11 +586,56 @@ You can search for available Actors using the tool: ${HelperTools.STORE_SEARCH}.
...
+
+                const structuredContent = {
+                    runId: actorRun.id,
+                    actorName,
+                    status: actorRun.status,
+                    startedAt: actorRun.startedAt?.toISOString() || '',
+                    input,
+                };
+
+                const response: { content: { type: 'text'; text: string }[]; structuredContent?: unknown; _meta?: unknown } = {
```
MQ37, line 613: nit: can we use some type from MCP sdk or from types.ts (crete one if needed)

tests/integration/suite.ts
```diff
@@ -1414,5 +1480,20 @@ export function createIntegrationTestsSuite(
                 }
             }
         });
+
+        it.runIf(options.transport === 'stdio')('should use UI_MODE env var when CLI arg is not provided', async () => {
```
MQ37, line 1494: nit: I would add a generic test that tests the client can connect with uiMode via the url param ?uiMode and it check by listing the tools that they provide the meta fields

src/tools/run.ts
```diff
@@ -43,17 +47,104 @@ USAGE EXAMPLES:
...
     },
     call: async (toolArgs: InternalToolArgs) => {
-        const { args, apifyToken } = toolArgs;
+        const { args, apifyToken, apifyMcpServer } = toolArgs;
         const parsed = getActorRunArgs.parse(args);
         const client = new ApifyClient({ token: apifyToken });
-        const v = await client.run(parsed.runId).get();
-        if (!v) {
-            return buildMCPResponse({ texts: [`Run with ID '${parsed.runId}' not found.`],
+
```
MQ37, line 57: change: I think we should add this tool to the default tool as the call-actor in async mode requires this tool. Also this needs to be always present in ?uiMode and currently it is not. @jirispilka

jirispilka: ok, let's do it

src/const.ts
```diff
@@ -215,8 +215,11 @@ These tools are called **Actors**. They enable you to extract structured data fr
 
 ### Tool dependencies
 - \`${HelperTools.ACTOR_CALL}\`:
-  - First call with \`step="info"\` or use \`${HelperTools.ACTOR_GET_DETAILS}\` to obtain the Actor’s schema.
+  - First call with \`step="info"\` or use \`${HelperTools.ACTOR_GET_DETAILS}\` to obtain the Actor's schema.
   - Then call with \`step="call"\` to execute the Actor.
+  - When \`step="call"\`, supports async execution via the \`async\` parameter:
+    - When \`async: false\` or not provided (default when UI mode is disabled): Waits for completion and returns results immediately.
+    - When \`async: true\` (default when UI mode is enabled): Starts the run and returns immediately with runId. Use \`${HelperTools.ACTOR_RUNS_GET}\` to check status and retrieve results.
```
```suggestion
  - When \`step="call"\`, supports async execution via the \`async\` parameter:
  - When \`async: false\` or not provided (default when UI mode is disabled): Waits for completion and returns results immediately.
  - When \`async: true\` (default when UI mode is enabled): Starts the run and returns immediately with runId. Use \`${HelperTools.ACTOR_RUNS_GET}\` to check status and retrieve results.
```

jirispilka: Just fixed indent

src/mcp/server.ts
```diff
@@ -431,43 +431,172 @@ export class ActorsMcpServer {
...
-                            mimeType: 'text/markdown',
+                resources.push({
+                    uri: 'file://readme.md',
+                    name: 'readme',
+                    description: `Apify MCP Server usage guide. Read this to understand how to use the server, especially in Skyfire mode before interacting with it.`,
+                    mimeType: 'text/markdown',
+                });
+            }
+
+            if (this.options.uiMode === 'openai') {
```
jirispilka, line 453: `openai` is repeated here several times, just note for others. We will keep it like this, we need to refactor to it across the codebase anyway, like `actor', 'actor-mcp`

src/tools/store_collection.ts
```diff
@@ -160,8 +177,61 @@ You can also try using more specific or alternative keywords related to your sea
...
+            return buildMCPResponse({
+                texts,
+                structuredContent,
+                _meta: {
+                    'openai/outputTemplate': 'ui://widget/search-actors.html',
+                    'openai/widgetAccessible': true,
+                    'openai/resultCanProduceWidget': true,
+                    'openai/widgetDescription': `Interactive actor search results showing ${actors.length} actors from Apify Store`,
+                    'openai/widgetDomain': 'https://apify.com',
+                    'openai/widgetCSP': {
```
jirispilka: the object `openai/widgetCSP` is repeated several times at different places. I believe this object is the same everywhere. It would be good to reuse it and not duplicated. 

Also other parameters in _meta might be reused

Something like:
```
// Add near the top of the file, after imports
const OPENAI_WIDGET_CSP = {
    connect_domains: ['https://api.apify.com'],
    resource_domains: [
        'https://mcp.apify.com',
        'https://images.apifyusercontent.com',
    ],
} as const;

const OPENAI_WIDGET_BASE_META = {
    'openai/widgetAccessible': true,
    'openai/resultCanProduceWidget': true,
    'openai/widgetDomain': 'https://apify.com',
    'openai/widgetCSP': OPENAI_WIDGET_CSP,
} as const;

```
Then reuse them in all three locations where they appear

src/mcp/server.ts
```diff
@@ -431,43 +431,172 @@ export class ActorsMcpServer {
...
+
+                    let widgetJsFilename = '';
+                    let widgetTitle = '';
+
+                    if (uri === 'ui://widget/search-actors.html') {
+                        widgetJsFilename = 'search-actors-widget.js';
+                        widgetTitle = 'Apify Actor Search';
+                    } else if (uri === 'ui://widget/actor-run.html') {
+                        widgetJsFilename = 'actor-run-widget.js';
+                        widgetTitle = 'Apify Actor Run';
```
jirispilka: This might be fragile .... we might try to load resource that does not exists, such as `search-actors-widget.js`. Is that correct. 
Should we have these stored together to avoid our errors?

Also widget files are in ../web/dist/ (separate build output)
No compile-time guarantee they exist
Easy to deploy server without building widgets first?

I'm not sure what would be the best way to solve this

src/tools/run.ts
jakcinmarina: oki, is there anything else I need to do or just this? not sure if @jirispilka you finished the review :)

MQ37: just modify the logic of the `loadToolsFromInput()` so when the `call-actor` tool is present or we are in the `?uiMode` the `get-actor-run` is also included automatically - and ideally integration test case for this

src/mcp/server.ts
jakcinmarina: Agree, new implementation now  moves widget configuration into a single registry and removes the remaining hardcoded paths and metadata.

All widget-related config (CSP, domain, flags) now lives in one place.

Widgets are copied to dist/ during postbuild so paths are consistent across local dev, npm installs, and Docker builds. A small build-time check script was also added so later CI can verify widget files are present. 

Adding new widgets should now be much simpler.

MQ37 (approved):
Thank you!

jirispilka (approved):
Thank you, there are some nits here and there but let's merge and fix it later. 

We need to merge it with other PR and then release it. Thank you!

PR #386, opened 2026-01-08
feat: implement migration plan from low-level Server API to high-level McpServer API

jirispilka:
**What**: 
Refactor `src/mcp/server.ts` from the deprecated low-level `Server` API to the recommended high-level `McpServer` API.

**Why**: 
- The low-level API is deprecated by the MCP SDK maintainers
- The high-level API provides automatic tool management and notifications
- Opportunity to simplify architecture and reduce code by ~500 lines

**Impact**:
- All existing features preserved (dynamic tool loading, Actor-MCP proxy, notifications)
- Cleaner, more maintainable codebase
- Future-proof against SDK changes

IMPORTANT - We don't want to start with implementation. This is only an analysis and possible impact

### DIFF
--- a/res/MCP_SERVER_REFACTOR_ANALYSIS.md
+++ b/res/MCP_SERVER_REFACTOR_ANALYSIS.md
@@ -39,7 +39,7 @@ Request → McpServer → Tool's callback (knows how to execute itself)
 ### Key Changes
 
 1. **Replace `Server` with `McpServer`** - Use high-level API as foundation
-2. **Convert JSON Schema to Zod** - Use `json-schema-to-zod` library for Actor schemas
+2. **Convert JSON Schema to Zod** - Use Zod v4 native JSON Schema conversion for Actor schemas
 3. **Move execution logic to callbacks** - Each tool type has its own registration pattern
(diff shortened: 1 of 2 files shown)


res/MCP_SERVER_REFACTOR_ANALYSIS.md
```diff
@@ -0,0 +1,967 @@
...
+
+**After (Callback-Per-Tool):**
+```
+Request → McpServer → Tool's callback (knows how to execute itself)
+```
+
+### Key Changes
+
+1. **Replace `Server` with `McpServer`** - Use high-level API as foundation
+2. **Convert JSON Schema to Zod** - Use `json-schema-to-zod` library for Actor schemas
```
MQ37: `json-schema-to-zod` is deprecated with Zod v4 as v4 supports this natively

MQ37 (commented):
I think the plans looks good but we will encounter a lot of issues when actually implementing this. I think we should make our testing setup more robust before doing changes like this and that would be with the refactor at least one full sprint effort I think. It may be worth it as it sometimes happen that the SDK update introduces breaking changes for our implementation.

Issue #365, opened 2025-12-16
Tool call-actor remove step=info and simplify the tool to support long running tasks


MQ37:
Currently we should not enable long running tasks support for `call-actor` tool because of the two step nature of the tool where it would store the `step=info` calls in Redis without actually needing too since these are instant and only the `step=call` need to be in Redis. We should remove the `step=info` logic and simplify this tool to only call Actor and enforce LLM to use the `fetch-actor-details` tool instead to get the input schema.

⚠️ Skyfire payments logic might be a painpoint in this case since there are some instructions baked in the `call-actor` tool

PR #387, opened 2026-01-08
feat: simplify call-actor tool and add long rung running task support,

MQ37:
closes https://github.com/apify/apify-mcp-server/issues/365

This is quite a large PR but core changes are not that huge and it was quite painful—a lot of **circular dependencies**, and I had to summon Opus to solve that 😆

🔧 **Changes** 🔧

- Simplified `call-actor` tool:  
  - Removed the `step` param  
  - Added long-running task support  
  - Moved the `step=info` logic to `fetch-actor-details` tool  

- Updated `fetch-actor-details` tool:  
  - Moved the `step=info` logic here via the `output` input param (supports granular output)  
  - Added `mcp-tools`, which uses the previous logic from `call-actor` (connects to MCP server Actors and lists the tools)  
  - It is backwards compatible  

- Improved workflow evals and added related test cases  

- Moved a few functions and consts to solve **circular dependencies**—it was really painful  

✅ **Testing** ✅

Also tested `fetch-actor-details` and `call-actor` manually with OpenCode, and it works well. 🚀

### DIFF
--- a/src/tools/fetch-actor-details.ts
+++ b/src/tools/fetch-actor-details.ts
@@ -57,14 +57,14 @@ const fetchActorDetailsToolArgsSchema = z.object({
         .min(1)
         .describe(`Actor ID or full name in the format "username/name", e.g., "apify/rag-web-browser".`),
     output: z.object({
-        description: z.boolean().default(true).describe('Include Actor description text only.'),
-        stats: z.boolean().default(true).describe('Include usage statistics (users, runs, success rate).'),
-        pricing: z.boolean().default(true).describe('Include pricing model and costs.'),
-        rating: z.boolean().default(true).describe('Include user rating (out of 5 stars).'),
-        metadata: z.boolean().default(true).describe('Include developer, categories, last modified date, and deprecation status.'),
-        inputSchema: z.boolean().default(true).describe('Include required input parameters schema.'),
-        readme: z.boolean().default(true).describe('Include full README documentation.'),
-        mcpTools: z.boolean().default(false).describe('List available tools (only for MCP server Actors).'),
+        description: z.boolean().default(true).optional().describe('Include Actor description text only.'),
+        stats: z.boolean().default(true).optional().describe('Include usage statistics (users, runs, success rate).'),
+        pricing: z.boolean().default(true).optional().describe('Include pricing model and costs.'),
+        rating: z.boolean().default(true).optional().describe('Include user rating (out of 5 stars).'),
+        metadata: z.boolean().default(true).optional().describe('Include developer, categories, last modified date, and deprecation status.'),
+        inputSchema: z.boolean().default(true).optional().describe('Include required input parameters schema.'),
+        readme: z.boolean().default(true).optional().describe('Include full README documentation.'),
+        mcpTools: z.boolean().default(false).optional().describe('List available tools (only for MCP server Actors).'),
     })
         .optional()
         .default({


src/tools/fetch-actor-details.ts
```diff
@@ -2,32 +2,105 @@ import { z } from 'zod';
...
+}
+
 const fetchActorDetailsToolArgsSchema = z.object({
     actor: z.string()
         .min(1)
         .describe(`Actor ID or full name in the format "username/name", e.g., "apify/rag-web-browser".`),
+    output: z.array(z.enum(['description', 'stats', 'pricing', 'rating', 'metadata', 'input-schema', 'readme', 'mcp-tools']))
+        .min(1)
+        .optional()
+        .default(['description', 'stats', 'pricing', 'rating', 'metadata', 'readme', 'input-schema'])
```
jirispilka: Instead of using an array of enums for selecting output fields, a more idiomatic and user-friendly approach in Zod is to use an object with boolean flags. This change simplifies the internal tool logic because you no longer need to check for inclusion in an array using .includes().

```
const fetchActorDetailsToolArgsSchema = z.object({
    actor: z.string()
        .min(1)
        .describe(`Actor ID or full name in the format "username/name", e.g., "apify/rag-web-browser".`),
    output: z.object({
        description: z.boolean().default(true).describe("Include Actor description text only."),
        stats: z.boolean().default(true).describe("Include usage statistics (users, runs, success rate)."),
        pricing: z.boolean().default(true).describe("Include pricing model and costs."),
....
    })
    .optional()
    .default({
        description: true,
        stats: true,
        pricing: true,
    .....
    })
    .describe("Specify which information to include in the response to save tokens."),
});
```

Another option is to use the `fields` as it is standard in rest API. We are using it when fetching dataset and it worked fine.

```
const fetchActorDetailsToolArgsSchema = z.object({
    actor: z.string()
        .min(1)
        .describe(`Actor ID or full name in the format "username/name", e.g., "apify/rag-web-browser".`),
    fields: z.string()
        .optional()
        .describe(`Comma-separated list of fields to include in the response. 
Available fields: 'description', 'stats', 'pricing', 'rating', 'metadata', 'input-schema', 'readme', 'mcp-tools'.
Default: all fields except 'mcp-tools'.`),
});
```

AI conclusion :) """ Use fields if you want to match the API convention. Use the Object/Boolean flags if you want the best performance and reliability from the LLM."""

src/tools/fetch-actor-details.ts
```diff
@@ -2,32 +2,105 @@ import { z } from 'zod';
...
+Default: ['description', 'stats', 'pricing', 'rating', 'metadata', 'readme', 'input-schema']. Use specific options to save tokens.`),
 });
 
 export const fetchActorDetailsTool: ToolEntry = {
     type: 'internal',
     name: HelperTools.ACTOR_GET_DETAILS,
     description: `Get detailed information about an Actor by its ID or full name (format: "username/name", e.g., "apify/rag-web-browser").
-This returns the Actor's title, description, URL, README (documentation), input schema, pricing/usage information, and basic stats.
-Present the information in a user-friendly Actor card.
+
```
jirispilka, line 87: This description is super long now :(

src/tools/actor.ts
```diff
@@ -504,11 +450,11 @@ You can search for available Actors using the tool: ${HelperTools.STORE_SEARCH}.
                 ? new ApifyClient({ skyfirePayId: args['skyfire-pay-id'] })
                 : new ApifyClient({ token: apifyToken });
 
-            // Step 2: Call the Actor
+            // Validate input parameter is provided
             if (!input) {
                 // Missing input is most likely an LLM error, so NOT marking as a soft-fail to track potential issues with tool description
                 return buildMCPResponse({
-                    texts: [`Input is required when step="call". Please provide the input parameter based on the Actor's input schema.`],
+                    texts: [`Input is required. Please provide the input parameter based on the Actor's input schema. Use fetch-actor-details tool with output=['input-schema'] to get the Actor's input schema first.`],
```
jirispilka: Here I would do the same logic as below, we return input-schema if the input is incorrect. Let us return input-schema if the input is empty.

I would write it something like ... [`fetch-actor-details` was called and schema retrieved, "input-schema"]. Just to tell LLM how the schema was retried (but that's not that important)

src/tools/fetch-actor-details.ts
MQ37: This is actually good idea, I like that :+1:

src/tools/fetch-actor-details.ts
MQ37: I will make it shorter :)

PR #393, opened 2026-01-19
fix: skyfire parameter injection and missing tool registration

MQ37:
## Executive Summary

Fixed critical bug where 8 out of 11 skyfire-enabled tools were missing the required \`skyfire-pay-id\` parameter in their input schemas, making them unusable in Skyfire payment mode. Refactored Skyfire integration to eliminate ~180 lines of duplicated code.

### 1. Schema Injection Bug

**Problem:** Only 3 tools were getting \`skyfire-pay-id\` parameter injected into their schemas. The other 8 tools would always fail because LLMs wouldn't see the parameter in the schema and wouldn't provide it.

**Solution:** Created \`SKYFIRE_ENABLED_TOOLS\` Set constant in \`src/const.ts\` and updated schema injection logic in \`src/mcp/server.ts\` to check this Set instead of using hardcoded tool names.

### 2. Missing Tool Registration

**Problem:** \`abortActorRun\` tool was defined but never added to the \`runs\` category, making it inaccessible.

**Solution:** Added \`abortActorRun\` to \`src/tools/categories.ts\`.

## Refactoring

Created reusable helper functions in \`src/utils/skyfire.ts\`:
- \`validateSkyfirePayId()\` - Validates presence of skyfire-pay-id parameter
- \`createApifyClientWithSkyfireSupport()\` - Creates ApifyClient with correct credentials

Refactored 13 code locations (tools and handlers) to use these helpers, eliminating duplicated validation and client creation logic.

## Infrastructure

- Added skyfire mode detection via \`?payment=skyfire\` query parameter in Actor server
- Added integration test to verify schema injection for all skyfire-enabled tools
- Updated test helpers to correctly pass payment parameter
- Updated documentation with integration test guidelines

## Impact

- **Users:** All 11 skyfire-enabled tools now work correctly in Skyfire payment mode
- **Developers:** Easier to add new skyfire-enabled tools (update Set constant only)
- **Maintainability:** Centralized logic reduces duplication and maintenance burden

### DIFF
--- a/src/tools/dataset.ts
+++ b/src/tools/dataset.ts
@@ -1,12 +1,12 @@
 import { z } from 'zod';
 
+import { createApifyClientWithSkyfireSupport } from '../apify-client.js';
 import { HelperTools, TOOL_STATUS } from '../const.js';
 import type { InternalToolArgs, ToolEntry, ToolInputSchema } from '../types.js';
 import { compileSchema } from '../utils/ajv.js';
 import { parseCommaSeparatedList } from '../utils/generic.js';
 import { buildMCPResponse } from '../utils/mcp.js';
 import { generateSchemaFromItems } from '../utils/schema-generation.js';
-import { createApifyClientWithSkyfireSupport } from '../apify-client.js';
 
 const getDatasetArgs = z.object({
     datasetId: z.string()
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -34,7 +34,7 @@ import { type ActorCallOptions } from 'apify-client';
 
 import log from '@apify/log';
 
-import { ApifyClient } from '../apify-client.js';
+import { ApifyClient, createApifyClientWithSkyfireSupport } from '../apify-client.js';
 import {
     ALLOWED_TASK_TOOL_EXECUTION_MODES,
     APIFY_MCP_URL,
@@ -72,7 +72,6 @@ import { parseBooleanFromString } from '../utils/generic.js';
 import { logHttpError } from '../utils/logging.js';
 import { buildMCPResponse } from '../utils/mcp.js';
 import { createProgressTracker } from '../utils/progress.js';
-import { createApifyClientWithSkyfireSupport } from '../apify-client.js';
 import { validateSkyfirePayId } from '../utils/skyfire.js';
 import { getToolStatusFromError } from '../utils/tool-status.js';
 import { cloneToolEntry, getToolPublicFieldOnly } from '../utils/tools.js';
--- a/src/tools/actor.ts
+++ b/src/tools/actor.ts
@@ -4,7 +4,7 @@ import { z } from 'zod';
 
 import log from '@apify/log';
 
-import { ApifyClient } from '../apify-client.js';
+import { ApifyClient, createApifyClientWithSkyfireSupport } from '../apify-client.js';
 import {
     ACTOR_MAX_MEMORY_MBYTES,
     CALL_ACTOR_MCP_MISSING_TOOL_NAME_MSG,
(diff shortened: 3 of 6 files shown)


src/utils/skyfire.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Creates ApifyClient with appropriate credentials based on Skyfire mode.
+ * In Skyfire mode, uses skyfire-pay-id from args; otherwise uses apifyToken.
+ *
+ * @param apifyMcpServer - The MCP server instance with configuration options
+ * @param args - Tool arguments that may contain skyfire-pay-id
+ * @param apifyToken - Standard Apify token for non-Skyfire mode
+ * @returns ApifyClient instance configured for the appropriate mode
+ */
+export function createApifyClientWithSkyfireSupport(
```
jirispilka: I believe a better place migth be apify-client.ts. It would just contain another function used to get apify client with skyfire support

src/tools/dataset.ts
```diff
@@ -56,17 +56,24 @@ USAGE EXAMPLES:
...
         openWorldHint: false,
     },
     call: async (toolArgs: InternalToolArgs) => {
-        const { args, apifyToken } = toolArgs;
+        const { args, apifyToken, apifyMcpServer } = toolArgs;
         const parsed = getDatasetArgs.parse(args);
-        const client = new ApifyClient({ token: apifyToken });
+
+        const skyfireError = validateSkyfirePayId(apifyMcpServer, args);
+        if (skyfireError) return skyfireError;
```
jirispilka: I somehow don't like this boiler place code in every tool. Call a function, returning a string and check the string is not null and returning errors. This seems like not a very good thing to do.

What about add this logic to server.ts? 

# Codex 😊 

## Goal
Stop every tool from doing:
```ts
const skyfireError = validateSkyfirePayId(...);
if (skyfireError) return skyfireError;
```
…and instead validate **once** in the main tool dispatcher.

## Why this fits your codebase
You already centralize a lot of per‑tool behavior in `src/mcp/server.ts` before calling the actual tool. For actor tools it already runs the skyfire guard in two places (regular tool call and async task path) before invoking the actor logic. Example:

```920:924:src/mcp/server.ts
if (tool.type === 'actor') {
    const skyfireError = validateSkyfirePayId(this, args);
    if (skyfireError) return skyfireError;

    // ...
}
```

So the natural refactor is to **move this check up** to a generic “tool requires Skyfire pay id” guard.

---

## Proposed design

### 1) Add a flag to tool definitions
Add an optional boolean flag in tool types (best place: `ToolBase` in `src/types.ts`) — something like:

- `requiresSkyfirePayId?: boolean`

`ToolBase` is shared by `HelperTool`, `ActorTool`, and `ActorMcpTool`, so this is a clean extension point.

### 2) Mark only tools that actually need it
Set `requiresSkyfirePayId: true` on tool definitions that use Apify API (all those currently calling `validateSkyfirePayId` + `createApifyClientWithSkyfireSupport`).

From your current usage list, that’s:

- `src/tools/run.ts` (all 3 tools in this file)
- `src/tools/get-actor-output.ts`
- `src/tools/dataset.ts` (3 tools)
- `src/tools/key_value_store.ts` (3 tools)
- `src/tools/actor.ts` (the internal helper tool in this file that uses Apify API)
- also **all actor tools** if you want to enforce it consistently

### 3) Enforce once in the dispatcher
In the main tool handler in `src/mcp/server.ts`, before branching by tool type, do:

```
if (tool.requiresSkyfirePayId) {
  const skyfireError = validateSkyfirePayId(this, args);
  if (skyfireError) return skyfireError;
}
```

Then delete the local guard from each tool file.

### 4) Do the same in async task execution
The long‑running task path (`executeToolAndUpdateTask`) also manually checks. It should use the same centralized guard before running the tool logic.

Currently it does this only for actor tools:

```1094:1102:src/mcp/server.ts
const skyfireError = validateSkyfirePayId(this, args);
if (skyfireError) {
    result = skyfireError;
    toolStatus = TOOL_STATUS.SOFT_FAIL;
} else {
    // ...
}
```

Move that to the top of `executeToolAndUpdateTask` and gate it by `tool.requiresSkyfirePayId`.

---

## Benefits
- Single policy for skyfire validation
- No duplicated “guard boilerplate” in tool files
- Easy to add/remove skyfire requirement per tool
- Consistent behavior across sync and async executions

---

## Edge cases / notes

- **Actor MCP tools**: probably `requiresSkyfirePayId = false` (they proxy external MCP servers, not Apify API).
- **Schemas**: you already allow additional properties for skyfire input in some tools (`additionalProperties: true`). That can remain as‑is.
- **Return behavior**: `validateSkyfirePayId` returns a response without `isError`, so your behavior stays consistent.

MQ37: @copilot do it bro

PR #398, opened 2026-01-21
feat: add internal tools for call-actor tool calls

jakcinmarina:
Introduced 2 internal tools that get called by `call-actor` tool. 
It wasn't possible to implement conditional rendering based on the renderWidget tool, since existing openai metadata in tool definitions pre-render the widget, even if empty - renders iframe with border. 

So since if user for example writes the prompt as "scrape google maps for coffee places" the flow was: 
- search actors to find suitable actor (rendered widget, so implemented an internal tool that returns minimal output)
- get actor details (rendered widget, so implemented an internal tool that returns schema only)
- finally call-actor with all needed info

### DIFF
--- a/src/tools/actor.ts
+++ b/src/tools/actor.ts
@@ -19,7 +19,7 @@ import { connectMCPClient } from '../mcp/client.js';
 import { getMCPServerTools } from '../mcp/proxy.js';
 import { getWidgetConfig, WIDGET_URIS } from '../resources/widgets.js';
 import { actorDefinitionPrunedCache } from '../state.js';
-import type { ActorDefinitionStorage, ActorInfo, ApifyToken, DatasetItem, InternalToolArgs, ToolEntry, ToolInputSchema } from '../types.js';
+import type { ActorDefinitionStorage, ActorInfo, ApifyToken, DatasetItem, InternalToolArgs, ToolEntry, ToolInputSchema, UiMode } from '../types.js';
 import { ensureOutputWithinCharLimit, getActorDefinitionStorageFieldNames, getActorMcpUrlCached } from '../utils/actor.js';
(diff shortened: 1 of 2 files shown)


src/tools/search-actors-internal.ts
```diff
@@ -0,0 +1,76 @@
...
+        return buildMCPResponse({
+            texts: [
+                `Found ${minimalActors.length} Actors for "${parsed.keywords}".`,
+            ],
+            structuredContent: {
+                actors: minimalActors,
+                query: parsed.keywords,
+                count: minimalActors.length,
+            },
+        });
```
MQ37, line 84: change: I would be defensive here and would put the found Actors both to the structured output and the texts since some clients may support only texts output - I know VSCode has (maybe it was updated) logic that if structured is available they use only structured and when only text is available they use the text, but still I would add that to text as not all clients support this.

src/tools/search-actors-internal.ts
```diff
@@ -0,0 +1,76 @@
...
+        const { args, apifyToken, userRentedActorIds, apifyMcpServer } = toolArgs;
+        const parsed = searchActorsInternalArgsSchema.parse(args);
+        let actors = await searchActorsByKeywords(
+            parsed.keywords,
+            apifyToken,
+            parsed.limit + ACTOR_SEARCH_ABOVE_LIMIT,
+            parsed.offset,
+            apifyMcpServer.options.skyfireMode ? true : undefined,
+        );
+        actors = filterRentalActors(actors || [], userRentedActorIds || []).slice(0, parsed.limit);
```
MQ37: nit: I would extract the shared logic with the original search actors tool to a func so it can be reused.

PR #399, opened 2026-01-21
fix: prevent get-actor-run loop after call-actor tool

jakcinmarina:
While testing I found that certain actors would still trigger the `get-actor-run` after `call-actor` tool call so I added a bit more tighter instructions around `get-actor-run` invocation.


MQ37 (changes requested):
Thank you for the PR. Based on my testing with `gpt-5-mini` the rules are too strict and the agent refuses to get me the results (the results is visible in the widget but the agent refuses to retrieve them for additional processing for example) :sweat_smile: 

Image

PR #411, opened 2026-01-28
feat: add hot reload to widgets [internal]

baldasseva:
**Why**

I need to work on the widgets, and it's annoying to have to rebuild at every step. 

**How**

I asked AI, and it spat out this script. It works. MCPJam is picking up the widgets correctly and it updates them when I update them <3

**Question**

- I've edited the `start:standby` script, but let me know if you prefer to create a new command for this.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -343,7 +343,7 @@ We use **4 spaces** for indentation (configured in `.editorconfig`).
 ### Development commands
 
 - `npm run start` - Start standby server (uses `tsx` for direct TypeScript execution)
-- `npm run dev` - Alias for `npm run start`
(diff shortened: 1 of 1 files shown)


MQ37 (approved):
LGTM 👍

package.json
```diff
@@ -81,7 +81,7 @@
   },
   "scripts": {
     "start": "npm run start:standby",
-    "dev": "npm run start:standby",
+    "dev": "node scripts/dev-standby.js",
```
jirispilka, line 84: Thank you, can you please update AGENTS.md 
This line, it should describe that it runs server with hot-reload
```
- `npm run dev` - Alias for `npm run start`
```

PR #414, opened 2026-01-28
feat(widget): setup shared Apify libraries

baldasseva:
The changes that I did here look something like this:

image

That's our shared card component, just to check that it works.
From now on, we can work to update the UI with our components.
The Search results text should be using our typography. Not sure that it does... But we'll check that out later
Also it doesn't go into dark mode. Probably we need to tweak something to let it notice that we are going into dark mode. Will look into it in the next days, I guess.

I've added too many people for review, but it's FYI.

### DIFF
--- a/src/web/src/utils/init-widget.tsx
+++ b/src/web/src/utils/init-widget.tsx
@@ -1,5 +1,9 @@
 import "../index.css";
 import React from "react";
+import { UiDependencyProvider } from "@apify/ui-library";
+import { tokens as lightCssVariables } from "@apify/ui-library/dist/src/design_system/colors/generated/css_variables.light.js";
+import { tokens as darkCssVariables } from "@apify/ui-library/dist/src/design_system/colors/generated/css_variables.dark.js";
+import { ThemeProvider } from "styled-components";
 import { createRoot } from "react-dom/client";
 
 function resolveTheme(): "light" | "dark" {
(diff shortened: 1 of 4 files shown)


src/web/src/utils/init-widget.tsx
```diff
@@ -1,5 +1,9 @@
 import "../index.css";
 import React from "react";
+import { UiDependencyProvider } from "@apify/ui-library";
+import { tokens as lightCssVariables } from "@apify/ui-library/dist/src/design_system/colors/generated/css_variables.light.js";
```
baldasseva, line 4: this is ugly. No other way to do this for now. 

If we have time and we need to modify ui-library for this job, we could export this with buckets properly.

src/web/src/utils/init-widget.tsx
```diff
@@ -42,9 +46,81 @@ export const renderWidget = (Component: React.FC) => {
...
+
+        // Inject fonts and CSS variables for proper styling
+        const head = document.head || document.getElementsByTagName("head")[0];
+        if (head) {
+            // Fonts preconnect + stylesheet
+            if (!document.getElementById("apify-fonts-preconnect-1")) {
+                const link1 = document.createElement("link");
+                link1.id = "apify-fonts-preconnect-1";
+                link1.rel = "preconnect";
+                link1.href = "https://fonts.googleapis.com";
```
baldasseva, line 58: Not sure the fonts are loading because the heading looks ugly XD

src/web/src/utils/init-widget.tsx
jmikitova: @baldasseva I'm not sure about this; I tried with normal
```
import { UiDependencyProvider, cssColorsVariablesLight, cssColorsVariablesDark } from "@apify/ui-library";
```
and I see the variables when checking `http://localhost:3000/index.html`
image

is this what we're trying to do here (?) 😅

PR #423, opened 2026-02-04
feat(vibe-code): add mcp config and docs for design system usage

jmikitova:
- adds Figma and Storybook mcps (run storybook locally from `apify-core` repo - `npm run storybook`)
- adds guidelines for developing front-end using design system components (`ui-library`)

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -82,7 +82,7 @@ What happens:
 - Editing files under `src/web/src/widgets/*.tsx` triggers a rebuild; the next widget render will use the updated code without restarting the server.
 
 Notes:
-- You can get your `APIFY_TOKEN` from https://console.apify.com/settings/integrations
+- You can get your `APIFY_TOKEN` from [Apify Console](https://console.apify.com/settings/integrations)
 - Widget discovery happens when the server connects. Changing widget code is hot-reloaded; adding brand-new widget filenames typically requires reconnecting the MCP client (or restarting the server) to expose the new resource.
 - You can preview widgets quickly via the local esbuild dev server at `http://localhost:3000/index.html`.
 
@@ -94,7 +94,7 @@ This repository includes a `.mcp.json` configuration file that allows you to use
 
 To use the Storybook MCP server (or any other MCP server that requires authentication), you need to configure your Apify API token in Claude Code's settings:
 
-1. Get your Apify API token from https://console.apify.com/settings/integrations
+1. Get your Apify API token from [Apify Console](https://console.apify.com/settings/integrations)
 2. Create or edit `.claude/settings.local.json` file
 3. Add the following environment variable configuration:
 


.mcp.json
```diff
@@ -0,0 +1,12 @@
+{
+    "mcpServers": {
+        "storybook": {
+            "type": "http",
+            "url": "http://localhost:6007/mcp"
```
baldasseva: https://baldasseva--storybook-mcp.apify.actor/mcp?token=APIFY_API_TOKEN (haha I do not want to force people to use this, but since this is open source, it could be better than a random localhost). Also, not everyone has core setup on their PC

DESIGN_SYSTEM_AGENT_INSTRUCTIONS.md
```diff
@@ -0,0 +1,265 @@
...
+**CRITICAL**: These instructions are MANDATORY for all UI/frontend work. Follow exactly as written.
+
+## Pre-Work Checklist (BLOCKING REQUIREMENT)
+
+Before ANY UI/component work:
+
+1. **Check MCP availability**:
+   - Search for `mcp__storybook__*` tools
+   - Search for `mcp__figma__*` tools
+   - If EITHER is missing: **STOP** and alert user to start required MCP server
```
baldasseva: Since this is an open-source repo and anybody could access it, fork it, etc., I wouldn't put these points as blocking requirements, but as suggestions.

DESIGN_SYSTEM_AGENT_INSTRUCTIONS.md
```diff
@@ -0,0 +1,265 @@
...
+color: '#1976d2'
+padding: '8px'
+border-radius: '4px'
+font-size: '14px'
+
+// ✅ REQUIRED
+color: ${theme.color.primary.action}
+padding: ${theme.space.space8}
+border-radius: ${theme.radius.radius8}
+${theme.typography.shared.tablet.bodyMMedium}
```
baldasseva: All instructions about typography tokens are not false, but also not entirely true XD
It's rare to use them directly; typically we use Text or Heading components from ui-library

DESIGN_SYSTEM_AGENT_INSTRUCTIONS.md
jmikitova: good point, I forgot about that

DESIGN_SYSTEM_AGENT_INSTRUCTIONS.md
jmikitova: I will add using `<Text>` and `<Heading>` components as "mandatory", so it doesn't hallucinate the tokens, and if there is some specific reason why it cannot use it, we can always correct it afterwards

.mcp.json
jmikitova: I already made a mistake and commited the token locally, so I think we need to figure out how to load the token from env vars :D :D

.mcp.json
```diff
@@ -0,0 +1,12 @@
+{
+    "mcpServers": {
+        "storybook": {
+            "type": "http",
+            "url": "https://baldasseva--storybook-mcp.apify.actor/mcp?token=APIFY_API_TOKEN"
```
katacek: nit: do we want to add a comment how to get / use the token, smt like? 

"_comment": "get APIFY_API_TOKEN from [console](https://console.apify.com/settings/integrations)"

and +1 for reading the token from envs, I think I will commit it with my first PR as well :D

DESIGN_SYSTEM_AGENT_INSTRUCTIONS.md
```diff
@@ -0,0 +1,261 @@
...
+
+1. **Get design context**:
+   ```
+   mcp__figma__get_design_context (with Figma URL)
+   ```
+
+2. **Extract design tokens**:
+   - Colors → Map to `theme.color.*`
+   - Spacing → Map to `theme.space.*`
+   - Typography → Map to `theme.typography.*`
```
katacek: nit: I would adjust typography it in favor of `Text` and `Heading` as elsewhere

Issue #397, opened 2026-01-21
Add more context to logged errors


MQ37:
Currently we are logging errors without additional context and it is hard to understand why it is happening without the "story". We should at least add MCP session ID so we can find what the client did before the error.

PR #427, opened 2026-02-06
chore: add mcpSessionId and sanitizeParams to all log statements

MQ37:
related to https://github.com/apify/apify-mcp-server/issues/397

## Summary
- Add `mcpSessionId` to all log statements that have access to MCP session context for better traceability
- Add `sanitizeParams()` utility to redact `skyfire-pay-id` from logged tool call parameters
- Refactor `callActorGetDataset` from positional params to a single options object
- Rename all log property keys from `sessionId` to `mcpSessionId` for consistency

## Testing
- ✅ `npm run type-check` passes
- ✅ `npm run test:unit` passes (213 tests)
- ✅ Lint passes

## Follow-up: AsyncLocalStorage
The manual `mcpSessionId` threading through function signatures and log calls is verbose. A follow-up PR should use Node.js `AsyncLocalStorage` to auto-inject context into all logs:

```typescript
import { AsyncLocalStorage } from 'node:async_hooks';
export const mcpContext = new AsyncLocalStorage<{ mcpSessionId?: string }>();

// Wrap at request entry point:
mcpContext.run({ mcpSessionId }, () => handleToolCall(request));

// Logger auto-injects context — no manual passing needed:
log.info('Calling Actor', { actorName }); // mcpSessionId added automatically
```

This would eliminate the need to pass `mcpSessionId` through every function signature and log call.


src/utils/logging.ts
```diff
@@ -52,3 +52,24 @@ export function logHttpError<T extends object>(error: unknown, message: string,
...
 }
+
+/**
+ * Sanitizes tool call parameters by redacting the skyfire-pay-id.
+ * Used for logging to avoid exposing the Skyfire payment token.
+ *
+ * @param params - The parameters object to sanitize
+ * @returns A new object with skyfire-pay-id replaced with '[REDACTED]'
+ */
+export function sanitizeParams(params: unknown): unknown {
```
jirispilka: I know you gonna not like, but the name of this function is not really clear, could be improved, what about `redactSkyfirePayId`

jirispilka: On top of that, this function is pretty ugly, codex suggests:
IMO this is nicer, but I'm not sure it is gonna work :) 

```
const SKYFIRE_PAY_ID_KEY = 'skyfire-pay-id';
const REDACTED_VALUE = '[REDACTED]';

const isPlainRecord = (value: unknown): value is Record<string, unknown> => {
    return typeof value === 'object' && value !== null && !Array.isArray(value);
};

/**
 * Sanitizes tool call parameters by redacting the skyfire-pay-id.
 * Used for logging to avoid exposing the Skyfire payment token.
 *
 * @param params - The parameters object to sanitize
 * @returns A new object with skyfire-pay-id replaced with '[REDACTED]'
 */
export function sanitizeParams(params: unknown): unknown {
    if (!isPlainRecord(params) || !(SKYFIRE_PAY_ID_KEY in params)) {
        return params;
    }

    if (params[SKYFIRE_PAY_ID_KEY] === REDACTED_VALUE) {
        return params;
    }

    return { ...params, [SKYFIRE_PAY_ID_KEY]: REDACTED_VALUE };
}
```

MQ37: @copilot do it pls bro. no mistakes or you go to jail

PR #452, opened 2026-02-11
feat(stdio): add Sentry error tracking for stdio transport

MQ37:
## Summary

- Adds Sentry SDK integration for the **stdio transport only** — no other transports are affected.
- Uploads source maps to Sentry during the release workflow so stack traces in the Sentry dashboard show original TypeScript source.
- Respects the existing `--telemetry-enabled` / `TELEMETRY_ENABLED` opt-out — Sentry is automatically disabled when telemetry is off.

## Changes

- **`src/instrument.ts`** (new): Sentry initialization with DSN, release version from `package.json`, and telemetry opt-out check.
- **`src/stdio.ts`**: Imports `instrument.js` as the first import; captures unhandled errors with `Sentry.captureException()` and flushes before exit.
- **`.sentryclirc`** (new): Sentry CLI config (org: `apify`, project: `apify-mcp`).
- **`.github/workflows/release.yaml`**: Adds source map inject + upload step in the `publish_to_npm` job using `SENTRY_AUTH_TOKEN` secret.
- **`package.json`**: Adds `@sentry/cli` devDependency.
- **`README.md`**: Documents Sentry usage in the Telemetry section.

## Notes

- Requires `SENTRY_AUTH_TOKEN` secret to be set in the GitHub repo (already configured).
- Source maps are uploaded only during the release workflow, tagged with the release version.

### DIFF
--- a/src/instrument.ts
+++ b/src/instrument.ts
@@ -7,12 +7,9 @@
  * Respects the --telemetry-enabled flag and TELEMETRY_ENABLED env var.
  * Sentry is disabled when telemetry is explicitly disabled.
  */
-import { createRequire } from 'node:module';
-
 import * as Sentry from '@sentry/node';
 
-const require = createRequire(import.meta.url);
-const packageJson = require('../package.json');
+import { getPackageVersion } from './utils/version.js';
 
 // Check if telemetry is disabled via CLI arg or env var before yargs parses them.
 // This mirrors the --telemetry-enabled / TELEMETRY_ENABLED option from stdio.ts.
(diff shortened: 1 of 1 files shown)


src/instrument.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * Respects the --telemetry-enabled flag and TELEMETRY_ENABLED env var.
+ * Sentry is disabled when telemetry is explicitly disabled.
+ */
+import { createRequire } from 'node:module';
+
+import * as Sentry from '@sentry/node';
+
+const require = createRequire(import.meta.url);
+const packageJson = require('../package.json');
```
jirispilka: You know, we are using `readFileSync` everywhere. This will break on non-linux os
```
  const authPath = join(homedir(), '.apify', 'auth.json');
  const content = readFileSync(authPath, 'utf-8');
```

jirispilka (approved):
Let's change it to `readFileSync`

src/instrument.ts
MQ37: ahh yes, lets use the get package version helper func, I totally forgot it exists :+1: Good catch bro :sunglasses:

PR #454, opened 2026-02-11
fix: Use readmeSummary instead of full readme to avoid context bloat

### LINKED ISSUE #422: Use Actor readme summary instead of full Actor readme
Actor readmes are typically huge and fill up context

Use the AI-generated summary (stored in Mongo) instead of the README.
Expose this summary via the public API and use it in the MCP server.

FYI @Jkuzz in kitchen we discussed that we can use the same summary that you’re using for Actor quality. 
I'm tagging you here to reconsider :)

janbuchar:
- closes #422
- [x] Resolve code TODO
- [x] Release API client and use stable release instead of beta here

### DIFF
--- a/src/tools/build.ts
+++ b/src/tools/build.ts
@@ -72,6 +72,7 @@ function pruneActorDefinition(response: ActorDefinitionWithDesc): ActorDefinitio
         actorFullName: response.actorFullName || '',
         buildTag: response?.buildTag || '',
         readme: response?.readme || '',
+        readmeSummary: response.readmeSummary,
         input: response?.input && 'type' in response.input && 'properties' in response.input
             ? {
                 ...response.input,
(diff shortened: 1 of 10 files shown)


src/tools/build.ts
```diff
@@ -36,7 +36,7 @@ export async function getActorDefinition(
             const actorDefinitions = buildDetails?.actorDefinition as ActorDefinitionWithDesc;
             // We set actorDefinition ID to Actor ID
             actorDefinitions.id = actor.id;
-            actorDefinitions.readme = truncateActorReadme(actorDefinitions.readme || '', limit);
+            actorDefinitions.readme = actor.readmeSummary ?? truncateActorReadme(actorDefinitions.readme || '', limit);
```
jirispilka: @janbuchar we might need to use full Actor readme in `ActorSearchDetail` in `src/web/src/pages/ActorSearch/ActorSearchDetail.tsx`

I think we should add a new field, we might need to use the full readme at some point.

PR #467, opened 2026-02-19
ci: Improve code checks

janbuchar:
- code checks are in a reusable workflow


MQ37 (approved):
LGTM, thank you! But I would honestly remove the on master push logic as the same logic is also called on the PR to master, right? So that just duplicates the workflow runs - I get that it can prevent issues in scenarios where someone merges before the CICD finishes on a PR but I would not duplicate it and keep just the on PR logic.

PR #468, opened 2026-02-19
feat: add ActorExecutor pattern for mode-aware direct actor tools

MQ37:
## Summary
- Introduce `ActorExecutor` abstraction that makes direct actor tools (`type: 'actor'`) UI-mode-aware
- **Default mode**: sync execution via `callActorGetDataset()` → same behavior as before
- **OpenAI mode**: async `start()` → return immediately with widget metadata + tracking

## Why
Direct actor tools were completely UI-mode-unaware — they always ran synchronously without widget metadata, regardless of `uiMode`. This was the primary motivating bug for the tool mode separation effort.

### New files
| File | Content |
|------|---------|
| `src/tools/default/actor-executor.ts` | `DefaultActorExecutor` — sync, dataset results |
| `src/tools/openai/actor-executor.ts` | `OpenAIActorExecutor` — async start, widget metadata |

### Modified files
| File | Change |
|------|--------|
| `src/types.ts` | Added `ActorExecutor`, `ActorExecutionParams`, `ActorExecutionResult` types |
| `src/mcp/server.ts` | Added `actorExecutor` field, constructor wiring, replaced both dispatch paths |

## Behavioral change
In OpenAI mode, direct actor tools now run **asynchronously** with widget tracking instead of synchronously. This is **intentional** — it matches how `call-actor` already behaves in OpenAI mode.

Default mode is **unchanged**.

## Review focus
- Is the executor interface correct for both dispatch paths?
- Does the default executor preserve existing behavior exactly?
- Is the `try/finally` addition in the task path safe? (Previously `progressTracker.stop()` wasn't called on exceptions)

## Verification
```
✅ npm run type-check — passes
✅ npm run lint — passes
✅ npm run test:unit — 248 tests pass
```

Part of [tool mode separation plan](https://github.com/apify/apify-mcp-server/pull/465). PR #3a of 7.


src/tools/openai/actor_executor.ts
```diff
@@ -0,0 +1,54 @@
...
+import { HelperTools } from '../../const.js';
+import { getWidgetConfig, WIDGET_URIS } from '../../resources/widgets.js';
+import type { ActorExecutionParams, ActorExecutionResult, ActorExecutor } from '../../types.js';
+
+/**
+ * OpenAI actor executor for UI mode.
+ * Runs actors asynchronously — starts the run and returns immediately with widget metadata.
+ * The widget automatically tracks progress and updates the UI.
+ */
+export const openaiActorExecutor: ActorExecutor = {
```
jirispilka, line 12: `openaiActorExecutor` this is really weird name

src/tools/default/actor_executor.ts
jirispilka, line 1: why to nest this in default? 

In the tools we are creating nested structure with core, default .... which is hard to understand its purpose

openai is clear, that's something I get

PR #484, opened 2026-02-23
fix(web): add cache-busting for OpenAI widget resources

MQ37:
## Summary

OpenAI's MCP client caches widget content keyed by resource URI. Since our widget URIs were static (`ui://widget/search-actors.html`), deploying new widget code had no effect — users kept seeing stale cached versions. This was [confirmed by the community](https://community.openai.com/t/issues-refreshing-widgets-keep-getting-cached-versions/1120690).

**Fix**: Compute a SHA-256 content hash of each widget JS file at server init and append it as a query parameter (`?v=<hash>`). When widget code changes → hash changes → URI changes → OpenAI fetches fresh content.

## Technical Changes

- Compute per-widget content hash (SHA-256, 16 hex chars) during `resolveAvailableWidgets()` at server init
- Emit versioned URIs in `listResources()` and tool response `_meta` (`openai/outputTemplate`)
- Strip `?v=...` in `readResource()` before map lookup for backward compatibility with old URIs
- Add `getVersionedWidgetMeta()` helper used by all tool files for dynamic responses
- Expose `getAvailableWidgets()` getter on `ActorsMcpServer` so tools can access versioned metadata
- Updated unit tests with `readFileSync` mock and assertions for `versionHash`/`versionedUri`

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -101,8 +101,13 @@ export class ActorsMcpServer {
     private telemetryEnabled: boolean | null = null;
     private telemetryEnv: TelemetryEnv = DEFAULT_TELEMETRY_ENV;
 
-    // List of widgets that are ready to be served
+    // List of widgets that are ready to be served (with version hashes for cache-busting)
     private availableWidgets: Map<string, AvailableWidget> = new Map();
+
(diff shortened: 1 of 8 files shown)


jirispilka (approved):
I added small fixes in: https://github.com/apify/apify-mcp-server/pull/486

PR #493, opened 2026-02-24
fix: Use USD cost from toolResponseMetadata instead of computeUnits

MQ37:
## Summary
This PR updates the ActorRun component to display the actual USD cost from `toolResponseMetadata.usageTotalUsd` instead of the raw `computeUnits` value from the actor stats.

## Key Changes
- Added `useOpenAiGlobal` hook to retrieve `toolResponseMetadata` containing the USD cost information
- Updated initial run data setup to extract `usageTotalUsd` from `toolResponseMetadata` and use it for the `cost` field
- Updated polling response handler to extract `usageTotalUsd` from the response's `_meta` object and use it for the `cost` field
- Updated mock OpenAI setup to support `toolResponseMetadata` configuration
- Added mock `toolResponseMetadata` with realistic USD cost values for testing
- Updated mock polling responses to include `_meta.usageTotalUsd` with varying costs based on completion status

## Implementation Details
- Both initial load and polling updates now follow the same pattern: extract `usageTotalUsd` as a number from their respective metadata sources, with fallback to `undefined` if not available
- The mock environment now provides realistic USD cost data (0.0456 for completed runs, 0.0123 for in-progress)
- Type safety is maintained with explicit type checking before casting `usageTotalUsd` to number

https://claude.ai/code/session_015gXLi3mUpVM2N4qMNTTTs8

### DIFF
--- a/src/web/src/pages/ActorRun/ActorRun.tsx
+++ b/src/web/src/pages/ActorRun/ActorRun.tsx
@@ -299,7 +299,7 @@ export const ActorRun: React.FC = () => {
             const developerUsername = extractDeveloperUsername(fullActorName);
 
             const usageTotalUsd = typeof toolResponseMetadata?.usageTotalUsd === 'number'
-                ? toolResponseMetadata.usageTotalUsd as number
+                ? toolResponseMetadata.usageTotalUsd
                 : undefined;
 
             setRunData({
@@ -317,7 +317,7 @@ export const ActorRun: React.FC = () => {
                 dataset: toolOutput.dataset,
             });
         }
-    }, [toolOutput, runData]);
+    }, [toolOutput, runData, toolResponseMetadata]);
 
     // Fetch actor details to get pictureUrl
     useEffect(() => {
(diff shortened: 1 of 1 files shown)


src/web/src/pages/ActorRun/ActorRun.tsx
katacek: suggestion: add the `toolResponseMetadata` to the dependency array

src/web/src/pages/ActorRun/ActorRun.tsx
```diff
@@ -296,6 +298,10 @@ export const ActorRun: React.FC = () => {
             const humanizedName = humanizeActorName(actorNameOnly);
             const developerUsername = extractDeveloperUsername(fullActorName);
 
+            const usageTotalUsd = typeof toolResponseMetadata?.usageTotalUsd === 'number'
+                ? toolResponseMetadata.usageTotalUsd as number
```
katacek: nit: the cast `as number` is not necessary

Issue #432, opened 2026-02-09
Simplify release process for cross-repo changes


jirispilka:
**Description:**
Deploying `apify-mcp-server` currently requires manually specifying a new version via the `custom` version input in the GitHub Action run. This is error-prone and slows down releases.

**Proposed solution:** Derive the version automatically (e.g., from git tag, package.json, or release workflow) so the deployment action does not require manual version input.

- Version is derived automatically where possible
- Release workflow is documented
- Manual override remains available when needed

**Code hints:** `.github/workflows/release.yaml` – `custom_version` input (line 18–21), `release_type: custom` path; `apify/workflows/git-cliff-release` usage; consider deriving from `package.json` version or git tag.

PR #499, opened 2026-02-25
ci: Refactor release workflows

janbuchar:
- Related to #432 
 - Refactor CI/CD workflows into reusable components: extract release metadata (`_update_release_metadata.yaml`), pkg.pr.new publishing (`_publish_to_pkg_pr_new.yaml`), and NPM publishing (`manual_publish_to_npm.yaml`) into standalone reusable/dispatchable workflows
 - Rename `release.yaml` → `manual_release_stable.yaml` and `pre_release.yaml` → `on_pull_request_label.yaml`, composing them from the new reusable workflows
 - Extend `on_master.yaml` to auto-publish beta prereleases to NPM and pkg.pr.new on every push to master (skipping ci, docs, and test commits)
     - the `before-beta-release.js` script to compute the next beta version suffix from the npm registry is copied from our other repos so it should be correct
- After merging, we will need to update the OIDC settings in npmjs


.github/workflows/on_pull_request_label.yaml
janbuchar, line 1: The contents of `evaluations.yaml` should probably also go here - thoughts?

jirispilka: Yes, with the new structure this seems to be a right place

PR #504, opened 2026-02-26
fix: Simplify AGENTS.md by moving coding standards to CONTRIBUTING.md 

jirispilka:
AGENTS.md was quite long, and the coding guidelines were scattered across both AGENTS.md and CONTRIBUTING.md.

I’ve moved everything into CONTRIBUTING.md and kept only a reference in AGENTS.md.

I’m wondering how effective Claude will be, I’m a bit concerned it will read the entire CONTRIBUTING.md every time anyway.


CONTRIBUTING.md
```diff
@@ -178,15 +197,53 @@ Use comments to guide reviewers:
...
+- ✅ Call `mcp__storybook__get-ui-building-instructions` first
+- ✅ Read 1-3 similar components for patterns (max 3 files)
+- ✅ Verify zero hardcoded values before submitting
+
+**Figma Integration**: Call `mcp__figma__get_design_context` when working from designs.
+
 ---
 
-## 6. Development Setup
+## 7. Development Setup
```
MQ37: change: please let's not use numbers so we do not have to reorder that like this when we insert a new section.

jirispilka: Fixed

jirispilka:
> > I've learned that Claude doesn't read AGENTS.md by default, but only CLAUDE.md (https://www.reddit.com/r/ClaudeAI/comments/1ngu3og/how_to_make_claude_understand_the_agentsmd_and/), so maybe you could introduce the `CLAUDE.md` file as well?
> 
> Yeah, that's correct. I'm solving it using symlink locally.
> 
> Hmm, should we have CLAUDE.md isntead of AGENTS.md then? @jmikitova @MQ37

I went ahead and moved CLAUDE.md to AGENTS.md

PR #508, opened 2026-02-26
feat(web): change web dev server port [internal]

baldasseva:
Hello, AI team

I am wiring my local apify-core repo to the local dev mcp server. Apify console frontend and the widgets in dev mode are both served by port 3000 and this is creating me problems. Since in the near future I will want to use the widgets in the console, I would like to start them both locally and this is the smallest change I can make to fix this problem

### DIFF
--- a/src/web/build.js
+++ b/src/web/build.js
@@ -118,7 +118,7 @@ async function buildAll() {
 
         const { host, port } = await ctx.serve({
             servedir: dirName,
-            port: 3000,
+            port: 3226,
         });
 
         console.log(`\n✓ Dev server running at http://${host}:${port}/`);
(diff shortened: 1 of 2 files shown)


src/web/build.js
```diff
@@ -118,7 +118,7 @@ async function buildAll() {
 
         const { host, port } = await ctx.serve({
             servedir: dirName,
-            port: 3000,
+            port: 3226,
```
baldasseva, line 121: 2/26 like today's date lol

jirispilka (approved):
Sounds good to me

src/web/build.js
MQ37: lol

MQ37 (approved):

PR #519, opened 2026-03-02
fix: Actor run show number of fields for nested objects

katacek:
solving issue with nested object in actor run response -> reported [here](https://www.notion.so/apify/ChatGPT-apps-311f39950a2280f4970de7a49c579630?source=copy_link#311f39950a228067acd1c0cd7d9c7b7a)

-> after the fix, insted of [object object] we show `XZ fields` (number of keys in the object) the same way as in console, the detail can be viewed after clicking on `View on Apify` (it would be overkill implementing the whole views logic to the widget)
-> for empty columns, we show '-' : I believe it is better solution than hiding the column entirely

Screenshot 2026-03-02 at 14 39 30

### DIFF
--- a/src/web/src/widgets/actor-run-widget.tsx
+++ b/src/web/src/widgets/actor-run-widget.tsx
@@ -60,6 +60,7 @@ setupMockOpenAi({
                                   {
                                       title: "Example Page 1",
                                       url: "https://example.com/page-1",
+                                      crawl: { loadedUrl: "https://example.com/page-1", httpStatusCode: 200, depth: 0, contentType: "text/html" },
                                       description: "This is a long description that should test text overflow and ellipsis in table cells",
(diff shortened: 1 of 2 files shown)


src/web/src/widgets/actor-run-widget.tsx
```diff
@@ -60,6 +60,7 @@ setupMockOpenAi({
                                   {
                                       title: "Example Page 1",
                                       url: "https://example.com/page-1",
+                                      crawl: { loadedUrl: "https://example.com/page-1", httpStatusCode: 200, depth: 0, contentType: "text/html" },
```
katacek, line 63: this is just for local development to see how it looks

jirispilka (approved):
Looks nice. Thanks

Image

PR #537, opened 2026-03-06
feat: Migrate widget client code to MCP Apps SDK

### LINKED ISSUE #535: Client-side changes for MCP Apps

jirispilka:
Widgets now use `@modelcontextprotocol/ext-apps` instead of `window.openai`. JSON-RPC over postMessage replaces the proprietary OpenAI Apps protocol.

> **Note**: 1300+ of the diff lines are `src/web/package-lock.json` — the only new dependency is `@modelcontextprotocol/ext-apps`. Real code change: **29 files, +403/−411**.

- **Server**: removed `openai/outputTemplate` from tool definition `_meta` (only `ui.resourceUri` stays). This matters because MCP Jam defaults to the legacy renderer when it sees both — and our widgets are pure MCP Apps now. The `openai/toolInvocation/*` UX hints are kept (they don't affect renderer detection).
- **Client**: new `McpAppProvider` React context wrapping the ext-apps `useApp()` hook. All `window.openai.*` calls replaced with `app.callServerTool()`, `app.openLink()`, `app.requestDisplayMode()`, `app.updateModelContext()`. Deleted the old `use-open-ai-global` hook, `use-widget-state` hook, and `window.openai` type declarations.
- **ChatGPT quirk**: first widget in a conversation doesn't receive `tool-result` via the bridge (iframe loads too late). Workaround reads `window.openai.toolOutput` as initial data — see comment in `mcp-app-context.tsx`.
- **No React.StrictMode** — double-mount breaks the ext-apps transport.
- **Run completion**: when an Actor run finishes, `app.updateModelContext()` notifies the model with status, dataset ID and item count so it can follow up.

Tested in both ChatGPT and MCP Jam.

`search-actors`
image

`run-actor`

image

`call-actor`

It works from time to time. I also checked the master branch and it has the same behaviour.
I believe this is chatgpt issue only

image

Closes #535


src/resources/resource_service.ts
```diff
@@ -90,6 +90,7 @@ export function createResourceService(options: ResourceServiceOptions): Resource
                 log.debug('Reading widget file', { uri, jsPath: widget.jsPath });
                 const fs = await import('node:fs');
                 const widgetJs = fs.readFileSync(widget.jsPath, 'utf-8');
+                const safeWidgetJs = widgetJs.replace(/<\/script/gi, '<\\/script');
```
MQ37: question: is this actually needed? Why would the widget JS file containe the script tag?

src/tools/openai/call_actor.ts
```diff
@@ -64,10 +64,6 @@ export const openaiCallActor: ToolEntry = Object.freeze({
         idempotentHint: false,
         openWorldHint: true,
     },
-    execution: {
```
MQ37, line 67: nice catch!

src/web/src/context/mcp-app-context.tsx
```diff
@@ -0,0 +1,76 @@
...
+    useEffect(() => {
+        if (!app) return;
+        setHostContext(app.getHostContext());
+
+        // ChatGPT sets window.openai.toolOutput synchronously as an Apps SDK
+        // compatibility layer. When the tool completes before the iframe loads
+        // (first widget in a conversation), the MCP Apps bridge never sends
+        // ui/notifications/tool-result. Read the sync value as initial data.
+        if (!receivedViaBridge.current) {
+            const openai = (window as any).openai;
```
MQ37: nit: I do not like the `as any` here

src/web/src/utils/mock-openai.ts
```diff
@@ -8,12 +7,17 @@ interface MockOpenAiConfig {
...
 }
 
+/**
+ * Sets up a mock `window.openai` for local development.
+ * The MCP Apps SDK auto-detects `window.openai` and uses its OpenAI transport,
+ * so this mock makes dev mode work transparently with `useApp()`.
+ */
 export const setupMockOpenAi = (config: MockOpenAiConfig = {}) => {
-    if (typeof window === "undefined" || window.openai) return;
+    if (typeof window === "undefined" || (window as any).openai) return;
```
MQ37: nit: please no any

src/web/src/widgets/actor-detail-widget.tsx
```diff
@@ -2,9 +2,10 @@ import { ActorSearchDetail } from "../pages/ActorSearch/ActorSearchDetail";
 import { setupMockOpenAi } from "../utils/mock-openai";
 import { MOCK_ACTOR_DETAILS_RESPONSE } from "../utils/mock-actor-details";
 import { renderWidget } from "../utils/init-widget";
+import { useWidgetProps } from "../hooks/use-widget-props";
 import type { ActorDetails } from "../types";
 
-const shouldEnableMocks = typeof window !== "undefined" && !window.openai;
+const shouldEnableMocks = typeof window !== "undefined" && !(window as any).openai;
```
MQ37: nit: same in this file

src/web/src/widgets/search-actors-widget.tsx
```diff
@@ -90,7 +90,7 @@ const mockActors = [
 ];
 
 // Set up mock window.openai for local development (no-ops when window.openai already exists)
-const isMockEnvironment = typeof window !== "undefined" && !window.openai;
+const isMockEnvironment = typeof window !== "undefined" && !(window as any).openai;
```
MQ37: nit: same here with the any

src/web/src/utils/mock-openai.ts
```diff
@@ -1,4 +1,3 @@
-import { OpenAiGlobals } from "../types";
```
katacek, line 1: - locally, Actor run widget was not rendered (get stuck at loading state) with the error Method not found, claude explanation

> The ext-apps SDK's PostMessageTransport sends ui/initialize to window.parent. When the widget loads       
  directly (not in a host iframe), window.parent === window — so the message echoes back to itself. The widget's own Protocol receives it as an        
  incoming request, can't handle it, and replies "Method not found". 
> The mock in mock-openai.ts needs to act as a postMessage host. Since we can't prevent the echo in same-window mode, the mock needs to:             
>  1. Handle widget→host messages (ui/initialize, tools/call, etc.)                                                                                     
 > 2. Ignore echoes of its own host→widget messages                                                                               


this is the change by claude after which it started to work

```
import { MOCK_ACTOR_DETAILS_RESPONSE } from "./mock-actor-details";

interface MockOpenAiConfig {
    toolOutput?: any;
    toolResponseMetadata?: any;
    callTool?: (name: string, args: any) => Promise<any>;
    initialWidgetState?: any;
}

/**
 * Methods that flow from Host → App (sent by the mock host to the widget).
 * When the widget is loaded directly (window.parent === window), these messages
 * echo back to the same window. We ignore them in the mock listener to avoid loops.
 */
const HOST_TO_APP_METHODS = new Set([
    "ui/notifications/tool-result",
    "ui/notifications/tool-cancelled",
    "ui/notifications/host-context-changed",
    "ui/notifications/sandbox-resource-ready",
    "ui/notifications/sandbox-proxy-ready",
    "ui/resource-teardown",
]);

/**
 * Sets up a mock MCP Apps host for local development.
 *
 * The ext-apps SDK uses JSON-RPC over postMessage (window.parent.postMessage).
 * When the widget is loaded directly in a browser (not in a host iframe),
 * window.parent === window, so the widget's outgoing messages echo back.
 *
 * This listener acts as the MCP Apps host side: it handles app→host requests
 * (ui/initialize, tools/call, ui/open-link, etc.) and ignores the echoes of
 * its own host→app notifications.
 *
 * Also sets window.openai.toolOutput for the ChatGPT race-condition fallback
 * in mcp-app-context.tsx.
 */
export const setupMockOpenAi = (config: MockOpenAiConfig = {}) => {
    if (typeof window === "undefined" || (window as any).openai) return;

    console.log("[mock-host] Setting up MCP Apps mock host");

    // Keep window.openai for the ChatGPT toolOutput fallback in mcp-app-context.tsx.
    // ChatGPT sets this synchronously; the fallback reads it when tool-result arrives
    // late over the bridge.
    (window as any).openai = {
        toolOutput: config.toolOutput || {},
        toolResponseMetadata: config.toolResponseMetadata || null,
    };

    const send = (message: object) => window.postMessage(message, "*");

    const sendToolResult = (toolOutput: any) => {
        send({
            jsonrpc: "2.0",
            method: "ui/notifications/tool-result",
            params: {
                content: [],
                structuredContent: toolOutput,
            },
        });
    };

    window.addEventListener("message", async (event: MessageEvent) => {
        const data = event.data;
        if (!data || data.jsonrpc !== "2.0") return;
        // Ignore JSON-RPC responses (id + result/error, no method) — these are
        // echoes of the mock's own responses bouncing back in same-window mode.
        if (!data.method) return;
        // Ignore host→app notifications echoing back.
        if (HOST_TO_APP_METHODS.has(data.method)) return;

        const { id, method, params } = data;

        switch (method) {
            case "ui/initialize":
                console.log("[mock-host] ui/initialize", params);
                send({
                    jsonrpc: "2.0",
                    id,
                    result: {
                        protocolVersion: "2026-01-26",
                        hostInfo: { name: "Dev Mock Host", version: "1.0.0" },
                        hostCapabilities: {
                            updateModelContext: { text: {}, structuredContent: {} },
                            message: { text: {} },
                            openLinks: {},
                        },
                        hostContext: {
                            theme: "light",
                            displayMode: "inline",
                            platform: "web",
                            locale: "en-US",
                        },
                    },
                });
                break;

            case "ui/notifications/initialized":
                // App finished initialization — send the initial tool result.
                console.log("[mock-host] app initialized, sending tool result");
                setTimeout(() => sendToolResult(config.toolOutput || {}), 0);
                break;

            case "tools/call": {
                const toolName = params?.name;
                const toolArgs = params?.arguments || {};
                console.log(`[mock-host] tools/call: ${toolName}`, toolArgs);
                let result: any;
                if (config.callTool) {
                    result = await config.callTool(toolName, toolArgs);
                } else if (toolName === "fetch-actor-details") {
                    result = {
                        structuredContent: MOCK_ACTOR_DETAILS_RESPONSE.structuredContent,
                    };
                } else {
                    console.warn(`[mock-host] No mock handler for tool: ${toolName}`);
                    result = {};
                }
                send({
                    jsonrpc: "2.0",
                    id,
                    result: {
                        content: [],
                        structuredContent: result.structuredContent ?? result,
                        ...(result._meta ? { _meta: result._meta } : {}),
                    },
                });
                break;
            }

            case "ui/update-model-context":
                console.log("[mock-host] updateModelContext", params);
                send({ jsonrpc: "2.0", id, result: {} });
                break;

            case "ui/open-link":
                console.log("[mock-host] openLink", params);
                if (params?.url) window.open(params.url, "_blank");
                send({ jsonrpc: "2.0", id, result: { isError: false } });
                break;

            case "ui/request-display-mode":
                console.log("[mock-host] requestDisplayMode", params);
                send({ jsonrpc: "2.0", id, result: { mode: params?.mode } });
                break;

            case "ui/message":
                console.log("[mock-host] message", params);
                send({ jsonrpc: "2.0", id, result: {} });
                break;

            default:
                console.log(`[mock-host] unhandled method: ${method}`);
        }
    });
};

export const updateMockOpenAiState = (updates: Record<string, unknown>) => {
    if (typeof window === "undefined") return;

    // Update window.openai for the ChatGPT fallback.
    if ((window as any).openai) {
        Object.assign((window as any).openai, updates);
    }

    // Send ui/notifications/tool-result via postMessage so the ext-apps SDK
    // picks up the updated data (replaces the old openai:set_globals event).
    if (updates.toolOutput !== undefined) {
        window.postMessage(
            {
                jsonrpc: "2.0",
                method: "ui/notifications/tool-result",
                params: {
                    content: [],
                    structuredContent: updates.toolOutput,
                },
            },
            "*"
        );
    }
};
```

src/web/src/pages/ActorRun/ActorRun.tsx
```diff
@@ -402,15 +428,15 @@ export const ActorRun: React.FC = () => {
...
-                            await setWidgetState({
-                                ...widgetStateRef.current,
-                                runStatus: newStatus,
-                                runId: runData.runId,
-                                datasetId: newData.dataset?.datasetId,
-                                itemCount: newData.dataset?.itemCount,
-                                lastUpdateTime: Date.now(),
-                            });
+                            // Notify the model that the run completed so it can follow up.
+                            if (app) {
```
katacek: `app` is always `true` here, as by the line 374
`if (!app || !runData?.runId) return;`

src/web/src/pages/ActorRun/ActorRun.tsx
```diff
@@ -97,6 +88,22 @@ const extractDeveloperUsername = (fullActorName: string): string => {
...
+ * Used when the host overwrites toolResult with a different tool call (e.g. search-actors),
+ * so the widget can still show the correct run when opened for a call-actor response.
+ */
+function getRunIdFromStableSource(hostContext: Record<string, unknown> | undefined): string | null {
+    if (typeof window !== "undefined") {
+        const params = new URLSearchParams(window.location.search);
+        const fromUrl = params.get("runId");
+        if (fromUrl?.trim()) return fromUrl.trim();
+    }
+    const fromHost = hostContext?.runId;
```
katacek: adding comment from my review by claude, it seems valid if I do understand it correctly (and claude does not halucinate when explaning it to me :D ), please, double-check and I am ok if it is ok :)

> non-standard hostContext.runId

  ```
const fromHost = hostContext?.runId;
  if (typeof fromHost === "string" && fromHost.trim()) return fromHost.trim();
```

>   McpUiHostContext from @modelcontextprotocol/ext-apps doesn't document a runId field — this reads a non-standard extension property. If the spec changes or the host doesn't set it, this silently returns null. The comment doesn't explain who/what populates hostContext.runId — worth documenting where this is expected to come from.

src/resources/resource_service.ts
jirispilka: Oh, thanks. I missed that. It is claude hallucinating

src/web/src/context/mcp-app-context.tsx
jirispilka: Fixed, thanks!

src/resources/resource_service.ts
jirispilka: Make sense

src/web/src/utils/mock-openai.ts
jirispilka: I'll fix it everywhere

src/web/src/pages/ActorRun/ActorRun.tsx
jirispilka: Good catch!

src/web/src/pages/ActorRun/ActorRun.tsx
jirispilka: Make sense, thanks! This seems to be hallucination only 🤦🏻 . I removed it

src/web/src/utils/mock-openai.ts
jirispilka: Oh, I completely missed `mocks` 

@katacek I think your code would work but it re-implements MCP JSON RPC calls
There is `AppBridge` in mcp-apps that can be used for this purpose so I've tried to use it.

Can you please check it again?
Or should we skip it since it is dev only?

Speaking about mock data, I'm not really happy that the mock data are part of the prod code. 
Looks confusing at the first glance 

Claude has created this issue for a discussion: https://github.com/apify/apify-mcp-server/issues/541

katacek: aah, ok ok - tested, it works now but I see your point about mock data being part of prod code 

I can check the issue and try to come up with better way (if next sprint would be ok?)

katacek (approved):

PR #542, opened 2026-03-09
feat: Remove mock data from prod build

### LINKED ISSUE #541: Separate dev/mock code from production widget entry points
`Claude` generated issue

Widget entry files (`actor-run-widget.tsx`, `search-actors-widget.tsx`, `actor-detail-widget.tsx`) mix dev-only mock setup code with prod

1. **Bloated bundles** mock code and its dependencies (e.g. `AppBridge` from `@modelcontextprotocol/ext-apps/app-bridge`) are bundled into production widgets (~2.1MB per widget)
2. **Unclear separation** — mock data (200+ lines of fake dataset items), mock host setup, and `renderWidget()` all live in the same file
3. **Inconsistent guards** — `actor-run-widget.tsx` checks `isLocalDev &&` while others don't; `setupMockOpenAi` has its own internal guard

## Current structure

```
widgets/
  actor-run-widget.tsx      # ~300 lines: mock data + mock setup + renderWidget()
  search-actors-widget.tsx  # ~130 lines: mock data + mock setup + renderWidget()
  actor-detail-widget.tsx   # ~20 lines: mock setup + renderWidget()
utils/
  mock-openai.ts            # mock host (AppBridge + window.openai fallback)
  mock-actor-details.ts     # mock fixture data
```

## Proposed 

Production entry points should be thin — just `renderWidget(Component)`. Dev code should be completely separate and only loaded in dev mode.

```
widgets/
  actor-run-widget.tsx       # just: renderWidget(ActorRun)
  search-actors-widget.tsx   # just: renderWidget(ActorSearch)
  actor-detail-widget.tsx    # just: renderWidget(ActorSearchDetail)
dev/
  mock-host.ts               # AppBridge dev host setup
  fixtures/                  # mock data for each widget
  dev-entry.ts               # conditional dev-only imports (dynamic import or Vite env guard)
```

This would:
- Eliminate mock code from production bundles entirely
- Drop bundle sizes significantly (no `AppBridge` in prod)
- Make widget entry points trivially simple

jirispilka:
The main motivation is to decrease bundle size. 
claude.ai complains that the bundles are larger than 5 MB.

The mock data were moved into separate files, hence the diff is a bit larger

Added an esbuild-defined compile-time flag IS_DEV_BUILD, moved widget mock setup into dev-only *.dev modules, and kept production widget entrypoints focused on rendering. This removes mock host code from shipped bundles while preserving local widget preview behavior.

closes #541


src/web/build.js
```diff
@@ -66,6 +66,7 @@ async function buildWidget(widget, watch = false) {
         outfile: resolve(dirName, 'dist', `${widget.name}.js`),
         plugins: [inlineCssPlugin],
         define: {
+            IS_DEV_BUILD: watch ? 'true' : 'false',
```
MQ37, line 69: what about we use env var instead of the `watch`?

MQ37 (approved):
Pre-approving 👍 Only one coment using env var instead of the global.

src/web/build.js
jirispilka: Yeah, I was considering different approaches and then went with the `watch`

I don't really have strong preference.

Let me pull @katacek for her opinion :)

katacek: I am fine with this approach, I understand it means less set up and it would only be 'an issue' if we would want to run dev mode without --watch (I cannot think of a case for it now)

jirispilka: @MQ37 I'll keep it this way. Since it is for dev-server only, we can change it later.

PR #552, opened 2026-03-12
fix: Fix CSP config for ChatGPT

jirispilka:


MQ37 (approved):
For me the widgets are loading properly on the desktop ChatGPT web ui - quite clunky but it loads eventually. On the mobile only the search Actors widget is loading, the call-actor is endlessly loading for some reason - more context in slack dms. Pre-approving 👍

PR #571, opened 2026-03-19
fix: Show error state in ActorRun widget instead of loading

jirispilka:
When call-actor returns an isError tool result, the widget now displays a "Failed" badge with a cleaned error message instead of showing "Loading Actor run data ..." forever. 

It can happen that Actor is called with a wrong input.

Non-existent Actor should not be called at all.

image

image

### DIFF
--- a/src/web/src/utils/actor-run.ts
+++ b/src/web/src/utils/actor-run.ts
@@ -1,7 +1,4 @@
-export type ToolResultLike = {
-    isError?: boolean;
-    content?: unknown[];
-};
+import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
 
 /**
  * Strip noise that is meant for the model, not the end-user:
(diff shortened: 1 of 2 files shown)


src/web/src/utils/actor-run.ts
```diff
@@ -0,0 +1,36 @@
...
+ */
+function cleanErrorText(raw: string): string {
+    // Take only the first meaningful line (the rest is model guidance)
+    const firstLine = raw.split("\n")[0].trim();
+    // Strip trailing JSON blob (e.g. `{"statusCode":404}`)
+    return firstLine.replace(/\s*\{[^}]*\}\s*\.?$/, "").trim();
+}
+
+export function extractActorRunErrorMessage(toolResult: ToolResultLike | null | undefined): string | null {
+    if (!toolResult?.isError || !Array.isArray(toolResult.content)) {
```
katacek: question: can there be a case of recieving  `{ isError: true }` but no content? as then, there would be error but we this would return null so we will end up in the same loading state 

(or we can trust the mcp type that the content will be there always)

src/web/src/utils/actor-run.ts
```diff
@@ -0,0 +1,36 @@
+export type ToolResultLike = {
```
katacek: suggestion: can we use `CallToolResult` type (as in `mcp-app-context.tsx`, line 8)

katacek (approved):
Ups, how could I forget about the error state? Thanks for fixing this! 👌 

I have just 2 suggestions (1 below, 1 in code), but no blockers - pre-approving

Suggestion 1: cn be done later, just we might double-check w design, the badge in the middle seems a bit weird (to me) 
Image

src/web/src/utils/actor-run.ts
jirispilka: Good point. Thanks!

src/web/src/utils/actor-run.ts
jirispilka: Nice catch! Thanks

PR #590, opened 2026-03-22
feat: introduce PaymentProvider interface and migrate Skyfire to provider pattern, add x402 payment provider

MQ37:
closes https://github.com/apify/ai-team/issues/150

## 🔗 PR Chain (1/3)
1. **👉 #590 (This PR) - Payment provider architecture & Skyfire migration**
2. ⏳ #591 - Base x402 payment provider support
3. ⏳ #608 - Finalize x402 payments (tool result propagation & interceptors)

---

## Summary

Replace hardcoded Skyfire payment logic with a generic `PaymentProvider` interface that supports multiple payment schemes. This introduces the architecture and migrates Skyfire; **[PR2 (#591)](https://github.com/apify/apify-mcp-server/pull/591) adds the x402 provider**.

### Changes

- **New `src/payments/` module** with `PaymentProvider` type, `SkyfirePaymentProvider`, `resolvePaymentProvider()` factory, and `preparePayment()` helper
- **`skyfireMode` boolean → `paymentProvider` instance** on `ActorsMcpServerOptions` — centralizes all payment logic in the provider
- **`requiresSkyfirePayId` → `paymentRequired`** on tool definitions — generic field usable by any payment scheme
- **New `preparePayment()` helper** — centralizes payment processing (validation, field stripping, client creation) in one call before AJV validation
- **`ApifyClient` creation moved to server handler** — passed via `InternalToolArgs.apifyClient` instead of per-tool creation
- **Removed `additionalProperties: true` hack** from common tools — no longer needed since payment fields are stripped before AJV validation
- **`skyfirePayId` → generic `paymentHeaders`** on `ApifyClient` — forwards arbitrary payment headers via request interceptors
- **Backward compatibility** preserved in `index_internals.ts` exports (deprecated re-exports)

### Architecture

```
PaymentProvider interface
├── decorateToolSchema() — add payment fields to tool schemas
├── validatePayment()    — check payment credentials in tool args
├── getPaymentHeaders()  — extract headers for Apify API requests
├── removePaymentFields()— clean payment fields from Actor input
├── allowsUnauthenticated— whether provider replaces Apify token
├── getUsageGuide()      — optional readme resource content
└── redactForLogging()   — redact sensitive fields for logging
```

### Verification

- ✅ `npm run type-check` — zero errors
- ✅ `npm run lint` — zero errors
- ✅ `npm run build` — passes
- ✅ `npm run test:unit` — 393 tests pass
- ✅ `npm run test:integration` — passes

[...]

PR #591, opened 2026-03-23
feat: implement X402PaymentProvider for x402 protocol support

MQ37:
## 🔗 PR Chain (2/3)
1. ✅ #590 - Payment provider architecture & Skyfire migration
2. **👉 #591 (This PR) - Base x402 payment provider support**
3. ⏳ #608 - Finalize x402 payments (tool result propagation & interceptors)

---

## Summary

Implements the `X402PaymentProvider` for x402 protocol support. Stacked on top of PR1 (#590) which introduced the PaymentProvider architecture.

### How x402 works

1. Client reads `_meta.x402` from tool definitions to know payment is required
2. Client signs an EIP-3009 TransferWithAuthorization and includes it in `params._meta["x402/payment"]`
3. Provider extracts the payment, base64-encodes it, and forwards as `PAYMENT-SIGNATURE` header
4. The Apify API verifies and settles the payment

### Changes

- **New `src/payments/x402.ts`** — Full `PaymentProvider` implementation that reads from `_meta["x402/payment"]` and forwards as `PAYMENT-SIGNATURE` header
- **`PaymentProvider` interface updated** — Added `PaymentMeta` type and optional `meta?` parameter to `validatePayment()` and `getPaymentHeaders()` so providers can access `_meta`
- **`InternalToolArgs` updated** — Added `meta?` field so internal tools can thread request metadata to payment providers
- **`createApifyClientWithPaymentSupport` updated** — Accepts optional `meta` 4th parameter
- **14 tool files updated** — Thread `meta` to `createApifyClientWithPaymentSupport`
- **`resolvePaymentProvider` updated** — x402 now creates `X402PaymentProvider` (replaces placeholder error)

### Verification

- ✅ `npm run type-check` — zero errors
- ✅ `npm run lint` — zero errors
- ✅ `npm run test:unit` — 393 tests pass

### Depends on

- #590 (PaymentProvider architecture + Skyfire migration)

### DIFF
--- a/src/payments/skyfire.ts
+++ b/src/payments/skyfire.ts
@@ -51,7 +51,7 @@ export class SkyfirePaymentProvider implements PaymentProvider {
 
     validatePayment(args: Record<string, unknown>): string | null {
         if (args[SKYFIRE_PAY_ID_KEY] === undefined) {
-            return 'Missing required "skyfire-pay-id" field. Obtain a Skyfire PAY JWT token via the create-pay-token tool and pass it as "skyfire-pay-id".';
+            return `Missing required "${SKYFIRE_PAY_ID_KEY}" field. Obtain a Skyfire PAY JWT token via the create-pay-token tool and pass it as "${SKYFIRE_PAY_ID_KEY}".`;
         }
         return null;
     }
--- a/src/tools/core/actor_tools_factory.ts
+++ b/src/tools/core/actor_tools_factory.ts
@@ -158,7 +158,7 @@ export async function getMCPServersAsTools(
     mcpSessionId?: string,
 ): Promise<ToolEntry[]> {
     /**
-     * This is case for the Skyfire request without any Apify token, we do not support
+     * This is case for the payment provider request without any Apify token, we do not support
      * standby Actors in this case, so we can skip MCP servers since they would fail anyway (they are standby Actors).
     */
     if (apifyToken === null || apifyToken === undefined) {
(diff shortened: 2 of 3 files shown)


tests/integration/suite.ts
```diff
@@ -2580,7 +2580,7 @@ export function createIntegrationTestsSuite(
             'should inject skyfire-pay-id parameter into all SKYFIRE_ENABLED_TOOLS when skyfireMode is enabled',
             async () => {
                 client = await createClientFn({
-                    skyfireMode: true,
+                    payment: 'skyfire',
```
stepskop, line 2583: Shouldn't this be needed in the previous PR that migrates the Skyfire integration structure?

src/payments/x402.ts
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        // Fall back to HTTP PAYMENT-SIGNATURE header (already base64-encoded by the client)
+        const headerPayment = getPaymentSignatureFromHeader(requestHeaders);
+        if (headerPayment) {
+            return { [PAYMENT_SIGNATURE_HEADER]: headerPayment };
+        }
+
+        return {};
+    }
```
stepskop, line 203: To be able to pay with x402, you need to include `x-apify-payment-protocol` header with `x402` as a value to pick which protocol will be used to authenticate and pay on the API.

tests/integration/suite.ts
MQ37: we will merge this PR into the other one :+1: It is not really needed but it is more maintainable this way

src/payments/x402.ts
MQ37: Yeah, good catch! :+1:

PR #633, opened 2026-04-01
feat: Normalize actor IDs before API lookup, unify tool call telemetry

jirispilka:
- Fix actor name input before API lookup — strip backticks, quotes, smart quotes, slash spacing. Avoids 404 SOFT_FAIL from LLM-garbled input.
- `logToolCallAndTelemetry()` — unified log + telemetry in one method,
- Added `toolName` to all log.info call lines.

For context, these issues were revelead by mezmo log analysis


src/tools/core/actor_tools_factory.ts
```diff
@@ -236,23 +289,27 @@ export async function getActorsAsTools(
             }
 
             try {
-                const actorDefinitionWithInfo = await getActorDefinition(actorIdOrName, apifyClient);
+                const actorDefinitionWithInfo = await getActorDefinition(normalized, apifyClient);
```
jirispilka: normalized what? This should be normalizedActorIdOrName` otherwise it is unclear

res/index.md
```diff
@@ -45,6 +45,10 @@ MCP task lifecycle, SDK features, and protocol reference for the Apify MCP serve
 - Related issues: #582 (async call-actor), #587 (roadmap), #588 (tool naming)
 - **Use case**: Reference when working on MCP task integration, resource links, or protocol-level features
 
+### [mezmo_soft_fail_fix_plan.md](./mezmo_soft_fail_fix_plan.md)
+Staff-level implementation plan from production Mezmo SOFT_FAIL log analysis: AJV / actor normalization / tool aliases / 4xx messaging / 402 copy / observability. Includes **log provenance** split between this package and `apify-mcp-server-internal`, Phase 0 triage, **two-repo execution order** (package release then internal bump + `streamable.ts` / `logging.ts` alignment), and a checklist synced from `.cursor/plans/`.
+- **Use case**: Prioritize error-quality and tooling work; avoid fixing gateway-only issues in the wrong repo
+
```
jirispilka:
```suggestion

```

jirispilka: Let's remove this, not needed

src/tools/core/actor_tools_factory.ts
```diff
@@ -215,6 +215,58 @@ export async function getMCPServersAsTools(
...
+/**
+ * Normalizes an Actor id and logs at INFO when client input differed.
+ * Single entry point for normalize+log — avoids duplicating the pattern at every call site.
+ */
+export function normalizeAndLogActorId(raw: string, extra?: Record<string, unknown>): string {
+    const normalized = normalizeActorIdForLookup(raw);
+    if (normalized !== raw) {
+        logActorIdNormalized(raw, normalized, extra);
+    }
+    return normalized;
```
jirispilka: Naming of this function is weird! As well as for the `normalizeActorIdForLookup`?
It is not clear what the function does!
We should name it like `cleanActorIdOrName` or  `cleanActorIdOrName` or any better name

Again, we are using `normalized`, which does not says what it is
Also raw is not a good input argument!

src/tools/core/actor_tools_factory.ts
```diff
@@ -225,7 +277,8 @@ export async function getActorsAsTools(
 
     const actorsInfo: (ActorInfo | null)[] = await Promise.all(
         actorIdsOrNames.map(async (actorIdOrName) => {
-            const actorDefinitionWithInfoCached = actorDefinitionPrunedCache.get(actorIdOrName);
+            const normalized = normalizeAndLogActorId(actorIdOrName, { mcpSessionId });
```
jirispilka: again, normalized used here, should be normalizedIdOrName

src/tools/core/actor_tools_factory.ts
```diff
@@ -236,23 +289,27 @@ export async function getActorsAsTools(
             }
 
             try {
-                const actorDefinitionWithInfo = await getActorDefinition(actorIdOrName, apifyClient);
+                const actorDefinitionWithInfo = await getActorDefinition(normalized, apifyClient);
                 if (!actorDefinitionWithInfo) {
                     log.softFail('Actor not found or definition is not available', {
-                        actorName: actorIdOrName, mcpSessionId, statusCode: 404, failureCategory: 'INVALID_INPUT',
+                        actorName: normalized,
+                        ...(normalized !== actorIdOrName && { actorNameInput: actorIdOrName }),
```
jirispilka: again: `normalized` -> replace

src/tools/core/call_actor_common.ts
```diff
@@ -279,7 +283,9 @@ export async function callActorPreExecute(toolArgs: InternalToolArgs): Promise<
     }
 > {
     const { args, apifyToken, apifyMcpServer, mcpSessionId } = toolArgs;
-    const parsed = callActorArgs.parse(args);
+    const parsedRaw = callActorArgs.parse(args);
+    const normalizedActor = normalizeAndLogActorId(parsedRaw.actor, { mcpSessionId, route: 'call-actor' });
```
jirispilka: normalizedActor? should be normalizedActorIdOrName

src/tools/core/call_actor_common.ts
```diff
@@ -279,7 +283,9 @@ export async function callActorPreExecute(toolArgs: InternalToolArgs): Promise<
     }
 > {
     const { args, apifyToken, apifyMcpServer, mcpSessionId } = toolArgs;
-    const parsed = callActorArgs.parse(args);
+    const parsedRaw = callActorArgs.parse(args);
```
jirispilka: parsedRaw? should be parsedArgs

src/tools/default/fetch_actor_details.ts
```diff
@@ -22,25 +23,26 @@ export const defaultFetchActorDetails: ToolEntry = Object.freeze({
     call: async (toolArgs: InternalToolArgs) => {
         const { args, apifyToken, apifyMcpServer, mcpSessionId } = toolArgs;
         const parsed = fetchActorDetailsToolArgsSchema.parse(args);
```
jirispilka: parsed should be parsedArgs for clarity

jirispilka (commented):
Naming is weird at many places, especially overly used is `normalized`, which is wrong and does not suggest what it means

src/tools/core/actor_tools_factory.ts
```diff
@@ -235,26 +235,26 @@ export function logActorIdNormalized(
...
- * Normalizes an Actor id and logs at INFO when client input differed.
- * Single entry point for normalize+log — avoids duplicating the pattern at every call site.
+ * Cleans an Actor id or name and logs at INFO when client input differed.
+ * Single entry point for clean+log — avoids duplicating the pattern at every call site.
  */
-export function normalizeAndLogActorId(raw: string, extra?: Record<string, unknown>): string {
-    const normalized = normalizeActorIdForLookup(raw);
-    if (normalized !== raw) {
-        logActorIdNormalized(raw, normalized, extra);
+export function normalizeAndLogActorId(actorIdOrName: string, extra?: Record<string, unknown>): string {
```
jirispilka: We need to also rename this function `normalizeAndLogActorId` to `cleanActorIdOrName`

src/utils/ajv.ts
```diff
@@ -32,10 +32,17 @@ export function fixZodSchemaRequired(schema: Record<string, unknown>): Record<st
...
+ * at the root. MCP / LLM clients regularly send extra top-level keys (client metadata, duplicated
+ * hints, transport leftovers). AJV then rejects the whole payload even when the declared fields
+ * are valid — production Mezmo logs showed validation SOFT_FAIL noise (e.g. `search-actors`) for
+ * otherwise acceptable calls. We only relax the **root** object; nested objects keep their own
+ * `additionalProperties` rules. Payment/session fields are stripped in `preparePayment()` before
+ * validation where applicable.
  */
 export function compileSchema(schema: Record<string, unknown>): ValidateFunction {
-    return ajv.compile(fixZodSchemaRequired(schema));
+    return ajv.compile(fixZodSchemaRequired({ ...schema, additionalProperties: true }));
```
Jkuzz: Could we use `removeAdditional: true` here instead to strip the properties instead of failing?

Jkuzz (commented):
LGTM, just one small question

src/utils/ajv.ts
jirispilka: Good catch! `removeAdditional: true` would actually work here. For the payment fields we are using `additionalProperties: true` but they are already stripped before validation (tests were missing, added). 

Fixed. Thanks

FYI it has forced me to fix stale comments and think a bit more about proper naming for the payments
So the fix commit is a bit larger

Jkuzz (approved):
Nice changes, LGTM (didn't read the tests)

PR #665, opened 2026-04-10
fix: Sanitize all env values in evals to prevent ERR_INVALID_CHAR

jirispilka:
## Summary
- Sanitize `OPENROUTER_CONFIG` (apiKey + baseURL) at capture time in `shared/config.ts` so all consumers get clean values
- Sanitize `APIFY_TOKEN` in workflow evals before passing to MCP subprocess
- Sanitize both OpenRouter clients in `evaluation_utils.ts` — `createClassifierEvaluator()` was completely unsanitized, causing 73/73 eval failures in CI
- Rename `sanitizeHeaderValue` → `sanitizeEnvValue` since it's used for URLs and tokens, not just headers

## Test plan
- [ ] CI evals pass without `ERR_INVALID_CHAR` errors
- [ ] Type-check and lint pass (verified locally)
- [ ] Unit tests pass (verified locally, 450/450)

### DIFF
--- a/evals/config.ts
+++ b/evals/config.ts
@@ -9,7 +9,7 @@ import { fileURLToPath } from 'node:url';
 import log from '@apify/log';
 
 // Re-export shared config
-export { OPENROUTER_CONFIG, sanitizeHeaderValue, validateEnvVars, getRequiredEnvVars } from './shared/config.js';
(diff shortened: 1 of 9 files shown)


jirispilka:
@copilot please fix all the review comments

Issue #638, opened 2026-04-02
tasks/list shows stale statusMessage for completed tasks


jirispilka:
## Bug

When listing tasks via `tasks/list`, completed tasks show the last intermediate progress message (e.g., "Starting the crawler.") instead of a meaningful final status message.

**Reproduction:**
```bash
$ mcpc @apify tools-call apify--rag-web-browser query:="MCP CLI" --task --detach
$ mcpc @apify tasks-list
```

All completed tasks show:
```
✔ completed - apify/rag-web-browser: Starting the crawler.
```

Instead of something like `apify/rag-web-browser: SUCCEEDED` or a summary of results.

## Root cause

In `src/mcp/server.ts:1133`, `storeTaskResult()` transitions the task to `completed` but does **not** update the `statusMessage`. The `TaskStore.storeTaskResult()` interface (from `@modelcontextprotocol/sdk`) doesn't accept a `statusMessage` parameter.

During execution, `statusMessage` is updated via `updateTaskStatus(taskId, 'working', message)` on a 5-second polling interval (`ProgressTracker.startActorRunUpdates()`). But the polling stops when the Actor reaches a terminal state, and no final `updateTaskStatus()` call is made before `storeTaskResult()`.

**Flow:**
1. Task created → status: `idle`, no statusMessage
2. Execution starts → `updateTaskStatus(taskId, 'working')` — no message
3. Progress polling → `updateTaskStatus(taskId, 'working', "apify/rag-web-browser: Starting the crawler.")` 
4. Actor finishes → polling stops, but **no final status update**
5. `storeTaskResult(taskId, 'completed', result)` → status becomes `completed`, but statusMessage is frozen at step 3

## Fix

Call `updateTaskStatus(taskId, 'working', finalMessage)` with the Actor's final status (e.g., `"apify/rag-web-browser: SUCCEEDED"`) **before** calling `storeTaskResult()`. This could be done either:
- In `ProgressTracker.startActorRunUpdates()` — emit one last update when terminal state is detected
- In `executeToolAndUpdateTask()` — call `updateTaskStatus` with a final message right before `storeTaskResult`

## Files involved

- `src/mcp/server.ts` — `executeToolAndUpdateTask()` (line ~1133)
- `src/utils/progress.ts` — `startActorRunUpdates()` polling loop

jirispilka:
Closing as fixed by #780.

[...]

PR #677, opened 2026-04-15
fix: set final statusMessage before storeTaskResult (#638)

jirispilka:
## Summary
- `storeTaskResult()` does not accept a `statusMessage` parameter, so it left the previous intermediate progress message on the task after completion
- Calls `updateTaskStatus(taskId, 'working', '<tool>: completed|failed|payment required')` immediately before each `storeTaskResult()` so the final message is in place before the terminal transition
- Verified end-to-end via mcpc: `tasks/list` now shows `"apify/rag-web-browser: completed"` instead of the stale `"Starting the crawler."`

## Test plan
- [x] Unit tests: `npm run test:unit` (2 regression cases in `task.statusMessage.test.ts`)
- [x] mcpc smoke test: call `apify/rag-web-browser`, check `tasks/list` → `statusMessage` shows `"<tool>: completed"`

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,7 +1,7 @@
 name: Bug report
 description: Report a bug or issue with the Apify MCP server
 title: "[Bug]: "
-labels: ["bug"]
+labels: ["bug","t-ai"]
 body:
   - type: dropdown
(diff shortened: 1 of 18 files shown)


jirispilka:
Closing in favor of #680 — same fix on a clean branch rebased on current master (the old branch had 7 stale commits from an older master state that muddled the diff).

PR #687, opened 2026-04-16
fix: Add actor id data to actor info

baldasseva:
The chat reponse in console has links that point to apify.com. If I don't receive ids in any way, I'm afraid I can't do anything.

So I want to receive actor ids in the structured data, then prompt the console AI chat to use that id to generate the links and ignore the apify.com url that is sent. I hope this will work!

### DIFF
--- a/src/tools/structured_output_schemas.ts
+++ b/src/tools/structured_output_schemas.ts
@@ -101,6 +101,7 @@ export const actorInfoSchema = {
     properties: {
         title: { type: 'string', description: 'Actor title' },
         url: { type: 'string', description: 'Actor URL' },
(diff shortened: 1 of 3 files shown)


jirispilka (approved):
LGTM, thanks!

I would fix the suggestion from Claude.

As the claude puts it :) 
"While this predates the PR, it was a natural opportunity to fix since the required array was already modified to add 'id'."

Pre-approving

jmikitova (approved):

jirispilka:
@baldasseva please merge it before this refactor: https://github.com/apify/apify-mcp-server/pull/694

I don't think there will be conflicts but it is better to merge this pr and deal with any possible conflict in the #694

PR #720, opened 2026-04-19
refactor: Rename 'openai' mode to 'apps' for MCP Apps compatibility

jirispilka:
## 🔗 PR chain — part of #577

This is **step 1 of 6** in the [#577 umbrella rollout](https://github.com/apify/apify-mcp-server/issues/577) of the MCP Apps decoupled pattern.

| # | Issue | PR | Status |
|---|-------|-----|--------|
| **1** | **#714 — Rename `openai` → `apps`**  | **#720 (this PR)** | **approved** |
| 2 | #715 — Capability auto-detect + gating | #721 **merged to #720** | merged |
| 3 | #716 — `fetch-actor-details` split (pilot) | #722 | in review |
| 4 | #717 — `search-actors` split | #723 | in review |
| 5 | #718 — `call-actor` split | #724 | in review |
| 6 | #719 — `get-actor-run` split | #734 | in review |

Closes #714.

## Release coordination

- **Blocks release of**: #721. This PR is an API-surface prep step; releasing it standalone would force `apify-mcp-server-internal` to migrate twice (once to `serverMode`, then again to `uiOverride` when #721 lands). Ship #720 + #721 together.
- **Downstream migration**: https://github.com/apify/apify-mcp-server-internal/issues/454 — the internal-repo audit is scoped to the `#720 + #721`.

---

## Summary

Renames the `'openai'` server mode to `'apps'` throughout the codebase to align with the MCP Apps naming convention. The `'openai'` string is retained as a deprecated alias in the external API (`UiMode`) for backward compatibility, with automatic normalization to `'apps'` and a deprecation warning at ingestion.

Preparatory refactor — **no behavior change**. Later PRs in the chain build the decoupled-pattern architecture on top.

### Type system & API surface
- `ServerMode`: `'default' | 'openai'` → `'default' | 'apps'`
- `UiMode` (external): `'apps' | 'openai'` (both accepted; `'openai'` normalized with a warning)
- New `resolveServerMode(uiMode)` in `src/types.ts` normalizes `UiMode` → canonical `ServerMode`
- `parseUiMode()` accepts `'apps'`, `'true'` (alias for `'apps'`), `'openai'` (deprecated alias)

### Directory & file renames
- `src/tools/openai/` → `src/tools/apps/` (all files, imports; `git mv` preserves history)
- `src/utils/server-instructions/openai.ts` → `.../apps.ts`
[...]

Issue #719, opened 2026-04-18
feat: Split get-actor-run into data + -widget tools


jirispilka:
## 🔗 PR chain — step 6 of 6 in #577

| # | Issue | PR | Status |
|---|-------|-----|--------|
| 1 | #714 — Rename `openai` → `apps` | #720 | in review |
| 2 | #715 — Capability auto-detect + gating | #721 | in review |
| 3 | #716 — `fetch-actor-details` split (pilot) | #722 | in review |
| 4 | #717 — `search-actors` split | #723 | in review |
| 5 | #718 — `call-actor` split | #724 | in review |
| **6** | **#719 (this issue)** | **#734** | **in review** |

Parent: #577. Final child. Completes the umbrella.

---

## Context and motivation

Final child of #577. Mirrors #716, #717, #718. Depends on #714, #715.

After this PR:
- `get-actor-run` = data-only, mode-independent (same ToolEntry returned in both modes)
- `get-actor-run-widget` = widget-only, additive in `apps` mode
- No `-internal` equivalent existed, so nothing to delete

### Why the widget variant now (not later)

Under today's `call-actor-widget` auto-polling model, `WORKFLOW_RULES` forbids calling `get-actor-run` after a widget render — so the apps-mode widget branch was effectively dead. Forward direction: `call-actor` becomes always-async (returns runId immediately, no UI), and the LLM picks `get-actor-run` (silent status) or `get-actor-run-widget` (UI progress) based on whether the user wants a visual. Shipping the widget variant now keeps the six-tool split symmetric with #722 / #723 / #724 and prepares the infrastructure before the async-execution policy change (tracked separately per #582).

### Shipped (Approach A — same as peers)
- `get-actor-run` across modes = data-only (current `default` behavior). Dropped tool-level widget `_meta`; trimmed stale "NEVER call in UI mode" block from base description (that rule lives in `WORKFLOW_RULES` now).
- Created `src/tools/apps/get_actor_run_widget.ts`:
  - Input: `{ runId }.strict()`
  - Output: widget-shaped run status (reuses shared `buildGetActorRunSuccessResponse({ widget: true })`)
  - `_meta.ui.resourceUri = ui://widget/actor-run.html`
  - `_meta.ui.visibility = ["model","app"]`, non-empty `_meta.ui.csp`
  - Response-level `openai/widgetDescription`
- Added `ACTOR_RUNS_GET_WIDGET` to `HelperTools`
[...]

Issue #718, opened 2026-04-18
feat: Split call-actor into data + -widget tools


jirispilka:
## 🔗 PR chain — step 5 of 6 in #577

| # | Issue | PR | Status |
|---|-------|-----|--------|
| 1 | #714 — Rename `openai` → `apps` | #720 | in review |
| 2 | #715 — Capability auto-detect + gating | #721 | in review |
| 3 | #716 — `fetch-actor-details` split (pilot) | #722 | in review |
| 4 | #717 — `search-actors` split | #723 | in review |
| **5** | **#718 (this issue)** | **#724** | **in review** |
| 6 | #719 — `get-actor-run` split | #734 | in review |

Parent: #577. Depends on: #714, #715, #716, #717. Stacked on #723's branch.

---

## Resolved — Approach A

Approach A landed in PR #724 (decoupled split, same pattern as peers). Real-session validation confirmed the LLM can legitimately call `call-actor` without wanting a progress widget (e.g., background job starts, scripted workflows where the user just wants a runId). Approach B would have forced every apps-mode caller into the widget surface, losing that flexibility. A-vs-B rationale is recorded in PR #724's body.

## Context and motivation

Child of #577. Depends on #714, #715. Best landed after #716, #717 validate the pattern.

`call-actor` is the special case:
- Always async in `apps` mode (returns run ID; progress polled by widget)
- Widget renders live run progress
- A data-only variant is less obviously useful — running an actor IS the user-facing moment

The right approach was deferred to PR time and is recorded in PR #724's description.

### Shipped (Approach A)
- `call-actor` = data tool in both modes — returns `{runId, actorName, status, startedAt, input}`, no widget `_meta`. Apps-mode variant still runs asynchronously.
- `call-actor-widget` = apps-only, same data + widget `_meta` for live progress rendering. Strict input: `{ actor, input, callOptions? }` — rejects stray `async` / `previewOutput` keys.
- Workflow rule "NEVER poll `get-actor-run` after `call-actor-widget`" scoped to the widget variant only; polling after silent `call-actor` is fine.
- Third disambiguation bullet in `src/utils/server-instructions/apps.ts` pairing `call-actor` with `call-actor-widget`.

### Out of scope
- Async execution policy decoupling from `apps` mode (#577 keeps them bundled)
- Other tools (child 6 — handled by #719 / PR #734)
- Widget UI code

## Files changed (PR #724)

| File | Change |
|------|--------|
[...]

Issue #717, opened 2026-04-18
feat: Split search-actors into data + -widget tools


jirispilka:
## 🔗 PR chain — step 4 of 6 in #577

| # | Issue | PR | Status |
|---|-------|-----|--------|
| 1 | #714 — Rename `openai` → `apps` | #720 | in review |
| 2 | #715 — Capability auto-detect + gating | #721 | in review |
| 3 | #716 — `fetch-actor-details` split (pilot) | #722 | in review |
| **4** | **#717 (this issue)** | **#723** | **in review** |
| 5 | #718 — `call-actor` split | #724 | in review |
| 6 | #719 — `get-actor-run` split | #734 | in review |

Parent: #577. Depends on: #714, #715, #716.

---

## Context and motivation

Child of #577. Mirrors the `fetch-actor-details` split (#716). Depends on #714, #715, and ideally #716 (pattern validated first).

After this PR:
- `search-actors` = data-only in both modes (replaces the `search-actors-internal` role)
- `search-actors-widget` = widget-only, additive in `apps` mode
- `search-actors-internal` = **removed**

### In scope
- Align `search-actors` across modes to pure data (current `default` behavior)
- Create `src/tools/apps/search_actors_widget.ts`:
  - Input: `keywords`, `limit`, `offset` (reuse `searchActorsBaseArgsSchema` per #700)
  - Output: widget-shaped result list
  - `_meta.ui.resourceUri`, `_meta.ui.visibility = ["model","app"]`, `_meta.ui.csp`
- Delete `src/tools/apps/search_actors.ts` (base is now mode-independent) and `src/tools/apps/search_actors_internal.ts`
- Add `STORE_SEARCH_WIDGET` to `HelperTools`; remove `STORE_SEARCH_INTERNAL`
- Update `src/tools/categories.ts`, `src/utils/server-instructions/apps.ts`
- Update references in `src/tools/apps/call_actor.ts`: `STORE_SEARCH_INTERNAL` → `STORE_SEARCH`

### Out of scope
- Other tools (children 5–6)
- Widget UI code

## Design notes

- Widget refetches server-side; input matches data tool.
- Existing `searchActorsBaseArgsSchema` (#700) reused for the widget input.
- Widget output schema: rename/adapt `actorSearchInternalOutputSchema` → `actorSearchWidgetOutputSchema`.

## Files to modify / delete / create

| File | Change |
|------|--------|
| `src/tools/apps/search_actors_widget.ts` | **create** |
[...]

PR #721, opened 2026-04-19
refactor: Auto-detect MCP Apps capability and gate server mode

### LINKED ISSUE #715: feat: Auto-detect MCP Apps capability, gate -widget tools
## 🔗 PR chain — step 2 of 6 in #577

| # | Issue | PR | Status |
|---|-------|-----|--------|
| 1 | #714 — Rename `openai` → `apps` | #720 | in review |
| **2** | **#715 (this issue)** | **#721** | **in review** |
| 3 | #716 — `fetch-actor-details` split (pilot) | #722 | in review |
| 4 | #717 — `search-actors` split | #723 | in review |
| 5 | #718 — `call-actor` split | #724 | in review |
| 6 | #719 — `get-actor-run` split | #734 | in review |

Parent: #577. Depends on: #714. Blocks: #716–#719.

---

## Context and motivation

Part of #577 rollout. Depends on #714 (rename). Blocks children that produce `-widget` tools.

The server already detects `clientSupportsUi` from `initialize.capabilities.extensions["io.modelcontextprotocol/ui"]` at `src/mcp/server.ts:115–121` and does **nothing** with it today. Mode is fixed at construction via `--ui`/`UI_MODE`.

After this PR, mode resolves per-connection from the client capability, with `--ui` as a bidirectional override. This matches the MCP Apps spec recommendation ("Servers SHOULD check client capabilities before registering UI-enabled tools").

No user-visible change yet — no `-widget` tools exist. Subsequent children produce them.

### In scope
- Consume `clientSupportsUi` to select tool set per connection (move tool-list resolution from construction to per-`initialize`)
- `--ui`/`UI_MODE` becomes bidirectional override:
  - unset + capability advertised → `apps` mode
  - unset + no capability → `default` mode
  - `--ui=true` → force `apps` regardless of capability
  - `--ui=false` → force `default` regardless of capability
- Tests cover the full matrix (override × capability → mode)

### Out of scope
- Creating `-widget` tools (children 3–6)
- Removing `-internal` tools (die naturally in children 3–6)
- Changes to widget `_meta` shapes

## Design notes

- **Per-connection mode**: `ActorsMcpServer` currently normalizes mode at construction (`src/mcp/server.ts:135`). After this PR, mode resolves at `initialize` time.
- **Stdio vs streamable-http**: stdio is 1:1 server:connection so existing model still works. Streamable-http may multiplex — confirm per-session tool lists.
- **`clientSupportsUi` detection**: existing check at `src/mcp/server.ts:115–121` stays. Just start using it.
- **`UiOverride` type**: internal representation `'on' | 'off' | 'auto'` — cleaner than boolean-or-undefined.

## Files to modify (high-level)

| Area | Change |
|------|--------|
| `src/mcp/server.ts` | Per-init mode resolution; consume `clientSupportsUi` |
| `src/stdio.ts`, `src/dev_server.ts` | `parseUiMode()` → `UiOverride = 'on' \| 'off' \| 'auto'` |
| `src/utils/tools_loader.ts` | Accept mode at load time if not already |
| Tests | Add mode-resolution matrix |

## Internal repo impact

`apify-mcp-server-internal` constructs `ActorsMcpServer` with a fixed `uiMode`. Needs to either become capability-aware or keep an explicit override. Audit required.

## Verification checklist
- [x] `npm run type-check` passes
- [x] `npm run lint` passes
- [x] `npm run test:unit` passes (new matrix test)
- [x] mcpc probe with simulated initialize payloads for both `clientSupportsUi` states
- [x] Behavior unchanged for existing callers
- [ ] Internal-repo audit complete

jirispilka:
## 🔗 PR chain — part of #577

Step **2 of 6** in the [#577 umbrella rollout](https://github.com/apify/apify-mcp-server/issues/577). **Depends on #720** (stacked on `claude/refactor-ui-handling-amyRL`; rebase to `master` after #720 merges).

| # | Issue | PR | Status |
|---|-------|-----|--------|
| 1 | #714 — Rename `openai` → `apps` | #720 | in review |
| **2** | **#715 — Capability auto-detect + gating** | **this PR** | **in review** |
| 3 | #716 — `fetch-actor-details` split (pilot) | #722 | in review |
| 4 | #717 — `search-actors` split | #723 | in review |
| 5 | #718 — `call-actor` split | #724 | in review |
| 6 | #719 — `get-actor-run` split | #734 | in review |

Closes #715.

Related PR in the internal repo: https://github.com/apify/apify-mcp-server-internal/pull/458

---

## Summary

Single hosted URL (`mcp.apify.com`) serves both UI-capable and non-UI clients. The server resolves mode per-connection from the client's `initialize` capabilities via a single `serverMode` option:

- `serverMode: 'default' | 'apps'` — force the mode (bypasses detection).
- `serverMode: 'auto'` (the default) — pick `'apps'` when the client advertises the `io.modelcontextprotocol/ui` extension with the widget MIME type, otherwise `'default'`.

Mode resolution + tool registration happen inside an override of the SDK's `initialize` **request** handler. The SDK awaits the handler's promise before sending `InitializeResult`, so tool loading naturally blocks the response — the first `tools/list` after initialize is guaranteed to see the correct mode-specific variants, without any transport-level orchestration.

No user-visible tool-shape change yet — children #722–#734 build on top.

---

### Final design: override the SDK's `initialize` request handler

```ts
// src/mcp/server.ts#setupInitializeHandler
this.server.setRequestHandler(InitializeRequestSchema, async (request) => {
    this.clientSupportsUi = detectClientSupportsUi(request);
    if (this.serverModeOption === 'auto') {
        this.serverMode = resolveServerMode('auto', this.clientSupportsUi);
        this.actorExecutor = actorExecutorsByMode[this.serverMode];
    }
    if (this.deferredToolsLoader) {
        const tools = await this.deferredToolsLoader();
        if (tools.length > 0) this.upsertTools(tools);
    }
    await this.resolveWidgets();
    return sdkInitHandler(request);  // delegate to SDK for protocolVersion + InitializeResult shape
});
```

The SDK's request dispatch awaits our handler's returned promise before sending the response (see `protocol.js#_onrequest`). Async work — including network fetches for Actor metadata — naturally blocks `InitializeResult`. Handler rejection is surfaced as a JSON-RPC error frame rather than a hung client.

### Two-phase tool loader

`src/utils/tools_loader.ts` is split so the expensive network fetch can run upfront while the mode-dependent composition happens synchronously inside the initialize handler:

- `fetchToolSources(input, client)` — **async, mode-agnostic**. Fetches Actor metadata from the Apify API; returns a reusable bundle.
- `buildToolsForMode(sources, mode)` — **sync, mode-dependent**. Composes the final tool list (mode-specific `-widget` variants, apps-only UI tools, auto-injected `get-actor-run` / `get-actor-output`, dedup).
- `loadToolsFromInput(input, client, mode)` — convenience wrapper running both phases.

Shared `normalizeInput(input)` helper returns `{ selectors, addActorEnabled, actorsExplicitlyEmpty }` — the only non-derivable flags both phases need.

### Unified instructions

Previously split into `apps.ts` and `default.ts` (via a shared `common.ts` template). The apps-specific content (widget workflow rules, `-widget` tool disambiguation) is **inert** for default-mode clients because those clients never see `-widget` tool names in `tools/list` — the rules cannot match any tool the model can call. Collapsed into a single `getServerInstructions()` in `src/utils/server-instructions/index.ts`. No `_instructions` mutation needed.

### ext-apps SDK integration

`src/mcp/server.ts` uses `getUiCapability` + `RESOURCE_MIME_TYPE` from `@modelcontextprotocol/ext-apps/server` — the canonical helpers per the MCP Apps spec. The hardcoded `'io.modelcontextprotocol/ui'` constant is gone. `tsconfig` switched to `moduleResolution: \"Bundler\"` so TypeScript honors the ext-apps `./server` conditional export.

---

### ❌ Drop auto-detect entirely (multiple URLs)

Two published URLs: `mcp.apify.com?ui=apps` vs `mcp.apify.com?ui=default`. Each client uses the right one.

**Rejected.** A single hosted URL is an important product property. Users connecting Claude Desktop / ChatGPT / Cursor shouldn't have to know or pick. The operational cost of auto-detect — ~50 lines of override code + fixture-level tests — is modest compared to the documentation/UX cost of explaining two URLs.

### ❌ Register tools in `server.oninitialized` (MCP reference pattern)

The ext-apps SDK docstring for `getUiCapability` shows the canonical pattern: read capabilities in `server.oninitialized`, register conditional tools there.

```ts
// Not used — see below
server.server.oninitialized = () => {
    const uiCap = getUiCapability(server.server.getClientCapabilities());
    if (uiCap?.mimeTypes?.includes(RESOURCE_MIME_TYPE)) { /* register apps tools */ }
    else { /* register default tools */ }
};
```

**Rejected**  The MCP base lifecycle requires clients to send `notifications/initialized` before any other request, but the SDK does **not** gate subsequent request handlers on that notification. Notification handlers are dispatched fire-and-forget via `Promise.resolve().then(handler)` — see `@modelcontextprotocol/sdk/shared/protocol.js#_onnotification`. A follow-up `tools/list` can race past `oninitialized` even if the callback body is synchronous, because tool registration and request dispatch run as separate microtasks.

The initialize-request-handler override is the only approach with a behavioral guarantee: it lives inside the request/response cycle the client is already waiting on.

References:
- MCP Apps spec: https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx
- MCP base lifecycle: https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle

---

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint`
- [x] `npm run test:unit` (591 total)
- [x] `npm run build`
- [x] mcpc probe matrix — verified tools-list on first connection, widget `_meta` present/absent as expected:
  - [x] **A** `auto` + no UI caps (mcpc default) → `default` mode, 8 tools, no `_meta.ui` on any tool
  - [x] **B** `--ui apps` explicit → `apps` mode, 10 tools, `search-actors` / `fetch-actor-details` / `call-actor` all return `_meta.ui.resourceUri` pointing to correct widget URIs
  - [x] **C** `--ui openai` (legacy alias) → identical to `--ui apps`
  - [x] **D** `--ui false` / `--ui default` explicit → `default` mode regardless, no widget metadata
  - [x] `UI_MODE=apps` env var → `apps` mode
  - [x] Full async flow: `call-actor(async=true)` → `get-actor-run(waitSecs)` → `get-actor-output` all succeed
  - [x] Sync `call-actor` returns full dataset output inline
  - [x] Widget `_meta` confirmed absent in default mode, present in apps mode (`ui://widget/search-actors.html`, `ui://widget/actor-run.html`)
- [x] Integration tests (human-run — needs `APIFY_TOKEN`)
- [ ] `apify-mcp-server-internal` audit — `options.serverMode: 'default' | 'apps'` still works as before. Callers that omit `serverMode` now get `'auto'` (was effectively `'default'`). Coordinated migration tracked in apify-mcp-server-internal#454.

## Migration note for library consumers

- `options.serverMode: 'default' | 'apps'` — unchanged behavior, forces the mode, bypasses detection.
- `options.serverMode: 'auto'` (new) or omitted — resolve from client capabilities.
- `options.uiMode` — still accepted as a deprecated alias.
- Unknown string input at ingestion boundaries (`UI_MODE=bogus`, `?ui=bogus`) now maps to `'auto'` (was `undefined` + DEFAULT fallback). Typos opt into auto-detect rather than silently forcing default.
- **`loadToolsByName` / `loadToolsFromUrl` / `loadToolsFromInput` require an explicit `serverMode`** when called outside the MCP connection lifecycle (no `initialize` handler). With `'auto'`, internal tools are queued until `initialize` fires and never applied if it doesn't. Pass `serverMode: ServerMode.DEFAULT` (or `'apps'`) to the constructor when using these methods directly — e.g. programmatic usage, integration tests, or the internal repo's restore path.


MQ37 (approved):
Solid architecture — the initialize-request-handler override is well-reasoned (the `oninitialized` race condition is real), and the two-phase tool loading split is clean. A few minor concerns:

---

**1. Naming nit: `detectClientSupportsUi` → `doesClientSupportUi`**
`src/mcp/server.ts:110`

For a function returning `boolean`, a question-style name reads more naturally:
```ts
// Current
function detectClientSupportsUi(request: InitializeRequest | undefined): boolean {

// Suggested
function doesClientSupportUi(request: InitializeRequest | undefined): boolean {
```
`detect` implies a side-effect or discovery process; `does`/`clientSupportsUi` makes the boolean contract obvious at the call site.

---

**2. Unified instructions may cause hallucinated tool calls in default mode**
`src/utils/server-instructions/index.ts:55-58`

The instructions explicitly name apps-only tools (`search-actors-internal`, `fetch-actor-details-internal`) and recommend them in disambiguation guidance. Default-mode clients will never see these in `tools/list`, but models aren't great at negative constraints — seeing a tool name in instructions can trigger attempted calls.

jirispilka:
> Solid architecture — the initialize-request-handler override is well-reasoned (the `oninitialized` race condition is real), and the two-phase tool loading split is clean. A few minor concerns:
> 
> **1. Naming nit: `detectClientSupportsUi` → `doesClientSupportUi`** `src/mcp/server.ts:110`
> 
> For a function returning `boolean`, a question-style name reads more naturally:
> 
> ```ts
> // Current
> function detectClientSupportsUi(request: InitializeRequest | undefined): boolean {
> 
> // Suggested
> function doesClientSupportUi(request: InitializeRequest | undefined): boolean {
> ```
> 
> `detect` implies a side-effect or discovery process; `does`/`clientSupportsUi` makes the boolean contract obvious at the call site.
> 
> **2. Unified instructions may cause hallucinated tool calls in default mode** `src/utils/server-instructions/index.ts:55-58`
> 
> The instructions explicitly name apps-only tools (`search-actors-internal`, `fetch-actor-details-internal`) and recommend them in disambiguation guidance. Default-mode clients will never see these in `tools/list`, but models aren't great at negative constraints — seeing a tool name in instructions can trigger attempted calls.


Good points. Thanks
1. detectClientSupportsUi -> `isUiSupportedByClient` (is prefix is used heavily at apify-core)
2. good catch: in the earlier version it was complicated to set the instructions per mode. But once I refactor it was actually quite easy. So I went for it.

Thanks for the review!

PR #737, opened 2026-04-21
feat: Surface compact input schema on search-actors results

### LINKED ISSUE #791: feat: Cap search-actors limit at 10 so includeInputSchema is always returned
## Context

apify-core#27466 added `includeInputSchema=true` to `GET /v2/store`, returning a compact JSON Schema per Actor. The endpoint hard-caps `limit <= 10` when the flag is set (returns 400 above that).

#737 enabled this enrichment in `searchAndFilterActors`, but only when the caller's `limit <= STORE_INPUT_SCHEMA_PAGE_LIMIT` (10). For `limit > 10`, the call drops `includeInputSchema=true` and falls back to the un-enriched search so the API does not 400.

## Problem

When `limit > 10`, search results no longer include `inputSchema` per Actor. Consumers that asked for >10 results lose schema enrichment entirely and have to fall back to per-Actor `fetch-actor-details` round-trips — exactly the fan-out we removed.

Larger result sets also clutter the LLM context window with low-signal Actor cards; in practice an LLM rarely needs more than a handful of candidates to pick from.

## Proposed fix

Cap the public `searchActorsBaseArgsSchema.limit` at `STORE_INPUT_SCHEMA_PAGE_LIMIT` (10) so every search is uniformly enriched.

`limit > 10` is rarely useful for an LLM consumer and the cost (no schema, mandatory fan-out, larger context footprint) outweighs the upside. Narrowing the public range matches what the upstream API enforces and removes the conditional schema-on/off branch in `searchAndFilterActors`.

For agents that genuinely need bulk Store browsing, a planned **code mode** (agent runs code on the server that calls the Apify API directly) is a better fit than a fat MCP tool response — it keeps the LLM context lean and lets the agent paginate, filter, and aggregate on its own. We should design for that path rather than scaling the search-actors response shape.

## Pointers

- `src/tools/core/search_actors_common.ts` — `searchActorsBaseArgsSchema.limit.max(100)` → `max(STORE_INPUT_SCHEMA_PAGE_LIMIT)`; update the `.describe()` text accordingly.
- `src/utils/actor_search.ts` — `searchAndFilterActors` can drop the `limit <= STORE_INPUT_SCHEMA_PAGE_LIMIT` branch once the cap is in place; always pass `includeInputSchema: true`.
- `tests/unit/utils.actor_search.test.ts` — drop the \"caller limit > cap\" case; assert the schema-on path is the only path.

References: apify-core#27466, #737

MQ37:
## Context
LLMs hitting `search-actors` previously had to make an N+1 round-trip (`fetch-actor-details` per hit) to learn each Actor's input shape. apify-core#27466 added `includeInputSchema=true` on `GET /v2/store` to project a compact schema server-side.

## Solution
Pass `includeInputSchema=true` and surface the returned shape as `inputFields` on both the structured card and the text card.

## Worth your attention
- **Public `limit` capped at 10** — `searchActorsBaseArgsSchema.limit.max(MAX_LIMIT_WITH_INPUT_SCHEMA)` so every search is uniformly enriched. Closes #791.
- **`searchActorsByKeywords` throws fail-fast** on `includeInputSchema=true` with `limit > MAX_LIMIT_WITH_INPUT_SCHEMA` rather than silently clamping or letting apify-core 400. Public path is Zod-capped, so this is a guard against future internal-caller misuse.
- **Rental filtering is upstream's job** — apify-core's `AGENT_SAFE_PRICING_MODELS` already filters rentals. The legacy `+50` over-fetch, `filterRentalActors`, and the `log.error` observability check are gone.
- **"Input schema" terminology dropped from LLM-facing text** — what we emit is a flat typed field list, not a real JSON Schema. Field name (`inputFields`), card label, helper (`inputFieldsToString`), and tool description all use "input fields" now; `actorInfoSchema.inputFields` encodes the wire shape so strict consumers can validate it.

## Follow-up
- apify-mcp-server-internal#500 — drop the dead Mongo lookup that populated `_meta.userRentedActorIds`
- #805 — drop the now-dead `userRentedActorIds` plumbing in this repo (`InternalToolArgs`, `_meta`, `mcp/server.ts`)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -74,7 +74,6 @@ Breaking changes must be coordinated; check whether updates are needed in `apify
 - **Follow [CONTRIBUTING.md](./CONTRIBUTING.md) for all naming and coding standards.** It is the single source of truth for naming rules (function verbs, boolean prefixes, type suffixes, enumerations, file names, etc.), string formatting, parameters, error handling, and anti-patterns. Read it before writing code.
 - **Validate tool inputs with Zod.** No ad-hoc shape checks.
 - **Reference tool names via the `HelperTools` enum**, not hardcoded strings (exception: integration tests).
-- **Put constants in `src/const.ts`**, not as local variables in individual files. This includes cache sizes/TTLs, concurrency limits, magic numbers, and string enums.
 - Always follow the latest [MCP spec](https://modelcontextprotocol.io/specification/2025-11-25) and [MCP Apps spec](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx).
 
 ## Further reading
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -74,6 +74,7 @@ Breaking changes must be coordinated; check whether updates are needed in `apify
 - **Follow [CONTRIBUTING.md](./CONTRIBUTING.md) for all naming and coding standards.** It is the single source of truth for naming rules (function verbs, boolean prefixes, type suffixes, enumerations, file names, etc.), string formatting, parameters, error handling, and anti-patterns. Read it before writing code.
 - **Validate tool inputs with Zod.** No ad-hoc shape checks.
 - **Reference tool names via the `HelperTools` enum**, not hardcoded strings (exception: integration tests).
+- **Put constants in `src/const.ts`**, not as local variables in individual files. This includes cache sizes/TTLs, concurrency limits, magic numbers, and string enums.
 - Always follow the latest [MCP spec](https://modelcontextprotocol.io/specification/2025-11-25) and [MCP Apps spec](https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx).
 
 ## Further reading
--- a/src/const.ts
+++ b/src/const.ts
@@ -110,7 +110,7 @@ export const USER_CACHE_TTL_SECS = 60 * 60; // 1 hour
 /** API rejects `includeInputSchema=true` above this; mirrors apify-core `MAX_LIMIT_WITH_INPUT_SCHEMA`. */
 export const MAX_LIMIT_WITH_INPUT_SCHEMA = 10;
 /** Max input fields shown inline in the text Actor card; structured output keeps the full schema. */
-export const MAX_INPUT_SCHEMA_TEXT_FIELDS = 20;
+export const MAX_INPUT_FIELDS_IN_TEXT_CARD = 20;
 
 export const ACTOR_PRICING_MODEL = {
     /** Rental Actors */
--- a/src/const.ts
+++ b/src/const.ts
@@ -109,6 +109,8 @@ export const USER_CACHE_MAX_SIZE = 200;
 export const USER_CACHE_TTL_SECS = 60 * 60; // 1 hour
 /** API rejects `includeInputSchema=true` above this; mirrors apify-core `MAX_LIMIT_WITH_INPUT_SCHEMA`. */
 export const STORE_INPUT_SCHEMA_PAGE_LIMIT = 10;
+/** Max input properties shown inline in the text Actor card; structured output keeps the full schema. */
+export const STORE_INPUT_SCHEMA_TEXT_FIELD_LIMIT = 20;
 
 export const ACTOR_PRICING_MODEL = {
     /** Rental Actors */
--- a/src/const.ts
+++ b/src/const.ts
@@ -107,6 +107,13 @@ export const MCP_SERVER_CACHE_MAX_SIZE = 500;
 export const MCP_SERVER_CACHE_TTL_SECS = 30 * 60; // 30 minutes
 export const USER_CACHE_MAX_SIZE = 200;
 export const USER_CACHE_TTL_SECS = 60 * 60; // 1 hour
+/**
+ * Hard cap on `limit` for `GET /v2/store?includeInputSchema=true`. The API
+ * returns 400 above this; matches `MAX_LIMIT_WITH_INPUT_SCHEMA` in apify-core.
+ */
+export const STORE_INPUT_SCHEMA_PAGE_LIMIT = 10;
+/** Max input properties rendered inline in the text Actor card; structured output keeps the full schema. */
+export const STORE_INPUT_SCHEMA_TEXT_FIELD_LIMIT = 10;
 
 export const ACTOR_PRICING_MODEL = {
     /** Rental Actors */
--- a/src/const.ts
+++ b/src/const.ts
@@ -108,9 +108,9 @@ export const MCP_SERVER_CACHE_TTL_SECS = 30 * 60; // 30 minutes
 export const USER_CACHE_MAX_SIZE = 200;
 export const USER_CACHE_TTL_SECS = 60 * 60; // 1 hour
 /** API rejects `includeInputSchema=true` above this; mirrors apify-core `MAX_LIMIT_WITH_INPUT_SCHEMA`. */
-export const STORE_INPUT_SCHEMA_PAGE_LIMIT = 10;
+export const MAX_LIMIT_WITH_INPUT_SCHEMA = 10;
 /** Max input properties shown inline in the text Actor card; structured output keeps the full schema. */
-export const STORE_INPUT_SCHEMA_TEXT_FIELD_LIMIT = 20;
+export const MAX_INPUT_SCHEMA_TEXT_FIELDS = 20;
 
 export const ACTOR_PRICING_MODEL = {
     /** Rental Actors */
--- a/src/const.ts
+++ b/src/const.ts
@@ -112,8 +112,6 @@ export const USER_CACHE_TTL_SECS = 60 * 60; // 1 hour
  * returns 400 above this; matches `MAX_LIMIT_WITH_INPUT_SCHEMA` in apify-core.
  */
 export const STORE_INPUT_SCHEMA_PAGE_LIMIT = 10;
-/** Max input properties rendered inline in the text Actor card; structured output keeps the full schema. */
-export const STORE_INPUT_SCHEMA_TEXT_FIELD_LIMIT = 10;
 
 export const ACTOR_PRICING_MODEL = {
     /** Rental Actors */
--- a/src/const.ts
+++ b/src/const.ts
@@ -107,10 +107,7 @@ export const MCP_SERVER_CACHE_MAX_SIZE = 500;
 export const MCP_SERVER_CACHE_TTL_SECS = 30 * 60; // 30 minutes
 export const USER_CACHE_MAX_SIZE = 200;
 export const USER_CACHE_TTL_SECS = 60 * 60; // 1 hour
-/**
- * Hard cap on `limit` for `GET /v2/store?includeInputSchema=true`. The API
- * returns 400 above this; matches `MAX_LIMIT_WITH_INPUT_SCHEMA` in apify-core.
- */
+/** API rejects `includeInputSchema=true` above this; mirrors apify-core `MAX_LIMIT_WITH_INPUT_SCHEMA`. */
 export const STORE_INPUT_SCHEMA_PAGE_LIMIT = 10;
 
 export const ACTOR_PRICING_MODEL = {
--- a/src/const.ts
+++ b/src/const.ts
@@ -109,7 +109,7 @@ export const USER_CACHE_MAX_SIZE = 200;
 export const USER_CACHE_TTL_SECS = 60 * 60; // 1 hour
 /** API rejects `includeInputSchema=true` above this; mirrors apify-core `MAX_LIMIT_WITH_INPUT_SCHEMA`. */
 export const MAX_LIMIT_WITH_INPUT_SCHEMA = 10;
-/** Max input properties shown inline in the text Actor card; structured output keeps the full schema. */
+/** Max input fields shown inline in the text Actor card; structured output keeps the full schema. */
 export const MAX_INPUT_SCHEMA_TEXT_FIELDS = 20;
 
 export const ACTOR_PRICING_MODEL = {
--- a/src/tools/core/search_actors_common.ts
+++ b/src/tools/core/search_actors_common.ts
@@ -22,7 +22,7 @@ export const searchActorsBaseArgsSchema = z.object({
         .min(1)
         .max(MAX_LIMIT_WITH_INPUT_SCHEMA)
         .default(5)
-        .describe(`The maximum number of Actors to return (1–${MAX_LIMIT_WITH_INPUT_SCHEMA}, default = 5).`),
+        .describe(`The maximum number of Actors to return (max = ${MAX_LIMIT_WITH_INPUT_SCHEMA}, default = 5).`),
     offset: z.number()
         .int()
         .min(0)
--- a/src/tools/core/search_actors_common.ts
+++ b/src/tools/core/search_actors_common.ts
@@ -78,8 +78,8 @@ Usage:
 - You MUST always do at least two searches: first with broad keywords, then optionally with more specific terms if needed.
 
 Important limitations: This tool does not return full Actor documentation or detailed usage instructions - only summary information.
-Each result includes a compact \`inputSchema\` (types only) so you can construct Actor input directly without a separate ${HelperTools.ACTOR_GET_DETAILS} call.
-For complete Actor details (descriptions per field, defaults, README), use the ${HelperTools.ACTOR_GET_DETAILS} tool.
+Each result lists the Actor's input fields with their types (e.g. \`url: string, maxResults?: number\`) so you can construct an Actor call directly without a separate ${HelperTools.ACTOR_GET_DETAILS} round-trip.
+For complete Actor details (per-field descriptions, defaults, README), use the ${HelperTools.ACTOR_GET_DETAILS} tool.
 The search is limited to publicly available Actors and may not include private, rental, or restricted Actors depending on the user's access level.
 
 Returns list of Actor cards with the following info:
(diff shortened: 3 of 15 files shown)


jirispilka:
I'm afraid this PR adds a lot of complexity. It was supposed to be a simple addition to the existing logic. 

The branch adds: a new cache, a generic semaphore utility, a parallel fetch pipeline that duplicates defaultBuild()+schema, and a third argument.

I can see three approaches:

1. Provide minimal code and reuse existing things as much as possible. Do not fetch inputs for all search results but for top 10. Use Promise.allSettled. Remove the semaphore. Extract helper for the defaultBuild()+schema and reuse it at the two places.

2. Add this to `apify-core /store/search` endpoint with `responseFormat=agent`. It already uses an `$in` join in fetchAgentSafetyContext (Acts2 + UsersActs2 + UserBackgroundChecks in parallel). Adding one more $in against ActorBuilds could be fine imo.

3.  Store a pre-computed input info into the Algolia index. We would not need to hit Mongo at all as the schema would be stored there. I'm not sure how often we sync the data. So there is a risk we might get a stale input fields.

My order of preference would be 3 -> 2 -> 1 but let me pull @Jkuzz for his opinion here.

jirispilka (changes requested):
See my comment here: https://github.com/apify/apify-mcp-server/pull/737

MQ37:
Makes sense, it is large and inefficient  change as I did not want to touch the API or MongoDB - let's touch the API then.

I would be for the option 2 so we do the input fields lookup from MongoDB for the default build - no need to handle the Algolia. Right? @Jkuzz

CLAUDE.md
```diff
@@ -72,6 +72,7 @@ Breaking changes must be coordinated; check whether updates are needed in `apify
 - **Follow [CONTRIBUTING.md](./CONTRIBUTING.md) for all naming and coding standards.** It is the single source of truth for naming rules (function verbs, boolean prefixes, type suffixes, enumerations, file names, etc.), string formatting, parameters, error handling, and anti-patterns. Read it before writing code.
 - **Validate tool inputs with Zod.** No ad-hoc shape checks.
 - **Reference tool names via the `HelperTools` enum**, not hardcoded strings (exception: integration tests).
+- **Put constants in `src/const.ts`**, not as local variables in individual files. This includes cache sizes/TTLs, concurrency limits, magic numbers, and string enums.
```
jirispilka: I've created this issue that is going against this rule: https://github.com/apify/apify-mcp-server/issues/643

TL;DR #643 :  What's wrong: Two root files have grown too large and unfocused:
  - src/const.ts (217 lines) — mixes unrelated domains; many constants used by only 1–2 files
  - src/types.ts (698 lines) — has dead exports and types that belong closer to where they're used

Why it matters: Over time this makes the codebase harder to navigate — you import from a massive root file rather than finding types/constants co-located with the feature that owns them.

src/tools/core/search_actors_common.ts
```diff
@@ -77,8 +77,9 @@ Usage:
 - Prefer broad, generic keywords - use just the platform name (e.g. "Instagram" instead of "Instagram scraper").
 - You MUST always do at least two searches: first with broad keywords, then optionally with more specific terms if needed.
 
-Important limitations: This tool does not return full Actor documentation, input schemas, or detailed usage instructions - only summary information.
-For complete Actor details, use the ${HelperTools.ACTOR_GET_DETAILS} tool.
+Important limitations: This tool does not return full Actor documentation or detailed usage instructions - only summary information.
+Each result includes a compact \`inputSchema\` (types only) so you can construct Actor input directly without a separate ${HelperTools.ACTOR_GET_DETAILS} call.
```
jirispilka: Compact input schema, the model might not know what you are reffering too. I would rephrase and add an example, e.g. arg1: str, arg2?: int.

src/tools/core/search_actors_common.ts
```diff
@@ -91,6 +92,7 @@ Returns list of Actor cards with the following info:
 - **Pricing:** Details with pricing link
 - **Stats:** Usage, success rate, bookmarks
 - **Rating:** Out of 5 (if available)
+- **Input schema:** Compact JSON Schema of input properties (types only)
```
jirispilka: I would not use "input schema" in the tool descriptions, it's a misleading name for a list of input arguments. The model will expect the full JSON Schema (with descriptions, defaults etc.), not a flattened arg list. 

All LLM-facing texts must not use "input schema". Use "input arguments" or "input fields"

src/utils/actor_card.ts
```diff
@@ -11,6 +11,25 @@ import {
     type StructuredPricingInfo,
 } from './pricing_info.js';
 
+function getInputSchema(actor: Actor | ActorStoreList): ActorStoreInputSchema | undefined {
+    return 'inputSchema' in actor ? actor.inputSchema : undefined;
+}
+
+function formatInputSchemaForText(inputSchema: ActorStoreInputSchema): string | null {
```
jirispilka: formatInputSchemaForText sounds really weird. Could be inputSchemaToText? inputSchemaToString sounds better, matches pricingInfoToString

src/utils/actor_search.ts
```diff
@@ -1,87 +1,80 @@
...
 
-    return filterRentalActors(actors || [], userRentedActorIds || []).slice(0, limit) as ActorStoreList[];
-}
+    // Observability: apify-core's `AGENT_SAFE_PRICING_MODELS` filter should mean we never see rentals here.
+    const rentalActors = actors
+        .filter((actor) => actor.currentPricingInfo?.pricingModel === ACTOR_PRICING_MODEL.FLAT_PRICE_PER_MONTH)
+        .map((actor) => `${actor.username}/${actor.name}`);
+    if (rentalActors.length > 0) {
+        log.error('Unexpected rental Actors in store search results', { keywords, rentalActors });
+    }
```
jirispilka: Do we really need this check? The filter is upstream's job, not ours.
This log.error begs for problems in mezmo. It iwll log eror that we can't control

src/const.ts
```diff
@@ -104,6 +104,10 @@ export const MCP_SERVER_CACHE_MAX_SIZE = 500;
 export const MCP_SERVER_CACHE_TTL_SECS = 30 * 60; // 30 minutes
 export const USER_CACHE_MAX_SIZE = 200;
 export const USER_CACHE_TTL_SECS = 60 * 60; // 1 hour
+/** API rejects `includeInputSchema=true` above this; mirrors apify-core `MAX_LIMIT_WITH_INPUT_SCHEMA`. */
+export const STORE_INPUT_SCHEMA_PAGE_LIMIT = 10;
```
jirispilka: The name STORE_INPUT_SCHEMA_PAGE_LIMIT is weird. I remember we decided to drop STORE_COLLECTION etc., and here you're bringing it back. 

MAX_LIMIT_WITH_INPUT_SCHEMA reads cleaner. Same goes for STORE_INPUT_SCHEMA_TEXT_FIELD_LIMIT two lines below.

src/utils/actor_card.ts
```diff
@@ -11,6 +11,25 @@ import {
...
+    const entries = Object.entries(inputSchema.properties);
+    if (entries.length === 0) return null;
+
+    const requiredSet = new Set(inputSchema.required ?? []);
+    const shown = entries.slice(0, STORE_INPUT_SCHEMA_TEXT_FIELD_LIMIT);
+    const fields = shown
```
jirispilka, line 24: Nit: I like the way mcpc formats input arguments: search-actors (keywords?:str, limit?:int, offset?:int). We emit string/integer/boolean, str/int/bool is shorter. The differene is negligible, but still,  it can save some tokens :)

Perhaps not worth to add a function to convert string -> str, just a note here

Let's drop it because it would mismatch the structuredContent

src/utils/actor_search.ts
```diff
@@ -1,87 +1,80 @@
...
+    const { search, apifyToken, limit, offset, allowsAgenticUsers, includeInputSchema } = options;
     const client = new ApifyClient({ token: apifyToken });
     const storeClient = client.store();
     if (allowsAgenticUsers !== undefined) storeClient.params = { ...storeClient.params, allowsAgenticUsers };
+    if (includeInputSchema !== undefined) storeClient.params = { ...storeClient.params, includeInputSchema };
 
-    const results = await storeClient.list({ search, limit, offset });
+    const effectiveLimit = includeInputSchema && limit !== undefined
+        ? Math.min(limit, STORE_INPUT_SCHEMA_PAGE_LIMIT)
+        : limit;
```
jirispilka: I know we discussed limit. I wasn't aware of this consequence: this Math.min silently clamps when a caller passes includeInputSchema=true with a high limit, and even through searchAndFilterActors the description promises a compact schema per result that we don't return when limit > STORE_INPUT_SCHEMA_PAGE_LIMIT. 

We should change the max limit to STORE_INPUT_SCHEMA_PAGE_LIMIT — also lets us drop the conditional and the Math.min.
```
const effectiveLimit = includeInputSchema && limit !== undefined
     ? Math.min(limit, STORE_INPUT_SCHEMA_PAGE_LIMIT)
     : limit;
```

tests/unit/utils.actor_search.test.ts
```diff
@@ -0,0 +1,119 @@
...
+}));
+
+function makeActor(idSuffix: number): ActorStoreList {
+    return {
+        id: `id-${idSuffix}`,
+        name: `actor-${idSuffix}`,
+        username: 'user',
+        currentPricingInfo: { pricingModel: 'FREE' },
+        stats: {},
+    } as unknown as ActorStoreList;
```
jirispilka: `as unknown as ActorStoreList` -  double-cast defeats the type system. The repo already has the better pattern: makeActorTool in tests/unit/tools.skyfire.test.ts:43 uses (overrides: Partial<X> = {}) => ({ ...defaults, ...overrides }).

tests/unit/utils.actor_card.test.ts
```diff
@@ -573,4 +573,71 @@ describe('formatActorToStructuredCard', () => {
...
+    });
+
+    it('renders every property without truncation so structured output and text stay in sync', () => {
+        const properties: Record<string, { type: string }> = {};
+        for (let i = 0; i < 15; i++) properties[`field${i}`] = { type: 'string' };
+        const actor = { ...mockActorStoreList, inputSchema: { type: 'object' as const, properties } } as ActorStoreList;
+        const result = formatActorToActorCard(actor);
+        expect(result).toContain('field0?: string');
+        expect(result).toContain('field14?: string');
+        expect(result).not.toMatch(/Input schema \(\d+ of \d+\)/);
```
jirispilka: Claude is saying this: Negative assertion on a string format the current code never produces

tests/unit/utils.actor_card.test.ts
```diff
@@ -573,4 +573,71 @@ describe('formatActorToStructuredCard', () => {
...
+        required: ['url'],
+    };
+
+    it('renders input schema as a TypeScript-like inline list with required marker', () => {
+        const actor = { ...mockActorStoreList, inputSchema } as ActorStoreList;
+        const result = formatActorToActorCard(actor);
+        expect(result).toContain('- **Input schema:** url: string, maxResults?: number');
+    });
+
+    it('renders every property without truncation so structured output and text stay in sync', () => {
```
jirispilka: name says "renders every property without truncation", but formatInputSchemaForText does truncate at 20. Test uses 15/20, so it just sits inside the
  no-truncation window. Name is misleading and coverage is missing

jirispilka (changes requested):
I like the direction! 

Glad rental actors are gone and the code is simpler. But  the tool description's "input schema" will confuse an LLM.

IMO the naming is off in a few places. I know you hate nitpick comments but I think we should approach a similar review bar as in apify-core :). We keep coming back to this code, and we've got the bandwidth now, so let's decrease the YOLO part a bit :)

CLAUDE.md
MQ37: Makes sense, let's split the file.

src/utils/actor_search.ts
MQ37: I though it is a good idea as we might want to know when accidentally serving rentals - then we know the API is wrong. But let's not complicate that - removing.

src/const.ts
MQ37: `MAX_LIMIT_WITH_INPUT_SCHEMA` - I do not honestly like this name as it does not specify what is the max limit for. But let's not waste calories here - renaming based on your suggestion as I do not have strong opinion about naming of this var.

src/utils/actor_card.ts
MQ37: renamed

src/tools/core/search_actors_common.ts
MQ37: Makes sense - renaming to input fields

tests/unit/utils.actor_card.test.ts
MQ37: great catch! :+1: This was bad and I agree we should be more pedantic/strict as we have more manpower

src/utils/actor_search.ts
MQ37: I agree, I wanted to keep it so the guys can have it as their first PR but then it introduced this complexity - does not make sense to do it and let's keep it simple and solve what we can in this PR.

tests/unit/utils.actor_card.test.ts
MQ37: great catch :facepalm:

src/tools/core/search_actors_common.ts
MQ37: Makes sense

src/utils/actor_card.ts
MQ37: I like the shorter format but for LLMs it currently it does not make sense, from my testing it can turn 388 token result in to 387, which I think is not worth it for additional logic we would need to write. So I agree - let's drop it now.

MQ37:
@jirispilka I agree, I should have checked the changes more - there was quite a lot of issues as I wanted to finally ship that and move fast. Let's slow down and do things properly.

I did put more time and effort into this PR and it should now be finalized.

src/tools/core/search_actors_common.ts
```diff
@@ -20,9 +20,9 @@ export const searchActorsBaseArgsSchema = z.object({
     limit: z.number()
         .int()
         .min(1)
-        .max(100)
+        .max(MAX_LIMIT_WITH_INPUT_SCHEMA)
         .default(5)
-        .describe('The maximum number of Actors to return (default = 5)'),
+        .describe(`The maximum number of Actors to return (1–${MAX_LIMIT_WITH_INPUT_SCHEMA}, default = 5).`),
```
jirispilka:
```suggestion
        .describe(`The maximum number of Actors to return (max = ${MAX_LIMIT_WITH_INPUT_SCHEMA}, default = 5).`),
```
I would drop the `1-`

src/const.ts
```diff
@@ -107,6 +107,10 @@ export const MCP_SERVER_CACHE_MAX_SIZE = 500;
 export const MCP_SERVER_CACHE_TTL_SECS = 30 * 60; // 30 minutes
 export const USER_CACHE_MAX_SIZE = 200;
 export const USER_CACHE_TTL_SECS = 60 * 60; // 1 hour
+/** API rejects `includeInputSchema=true` above this; mirrors apify-core `MAX_LIMIT_WITH_INPUT_SCHEMA`. */
+export const MAX_LIMIT_WITH_INPUT_SCHEMA = 10;
+/** Max input properties shown inline in the text Actor card; structured output keeps the full schema. */
+export const MAX_INPUT_SCHEMA_TEXT_FIELDS = 20;
```
jirispilka: MAX_INPUT_SCHEMA_TEXT_FIELDS is misleading. I know, you have changed that already.

This should be something like MAX_INPUT_FIELDS_RETURNED (or something better, it should contain fields)

src/const.ts
```diff
@@ -107,6 +107,10 @@ export const MCP_SERVER_CACHE_MAX_SIZE = 500;
 export const MCP_SERVER_CACHE_TTL_SECS = 30 * 60; // 30 minutes
 export const USER_CACHE_MAX_SIZE = 200;
 export const USER_CACHE_TTL_SECS = 60 * 60; // 1 hour
+/** API rejects `includeInputSchema=true` above this; mirrors apify-core `MAX_LIMIT_WITH_INPUT_SCHEMA`. */
+export const MAX_LIMIT_WITH_INPUT_SCHEMA = 10;
+/** Max input properties shown inline in the text Actor card; structured output keeps the full schema. */
```
```suggestion
/** Max input fields shown inline in the text Actor card; structured output keeps the full schema. */
```

src/utils/actor_search.ts
```diff
@@ -1,87 +1,70 @@
...
 export async function searchActorsByKeywords(
-    search: string,
-    apifyToken: string,
-    limit: number | undefined = undefined,
-    offset: number | undefined = undefined,
-    allowsAgenticUsers: boolean | undefined = undefined,
+    options: SearchActorsByKeywordsOptions,
 ): Promise<ActorStoreList[]> {
+    const { search, apifyToken, limit, offset, allowsAgenticUsers, includeInputSchema } = options;
+    if (includeInputSchema === true && limit !== undefined && limit > MAX_LIMIT_WITH_INPUT_SCHEMA) {
```
jirispilka: limit is never undefined in this context, it is called like this
```
            searchActorsByKeywords({
                search: actorSlug,
                apifyToken: apifyClient.token || '',
                limit: ACTOR_DETAILS_PICTURE_SEARCH_LIMIT,
            }).catch(() => []),
```
and in the actor_search, where zod will fill in default value

I think the best fix is to make it required and then you don't need to play with the condition

But I would not throw in this context!

jirispilka (changes requested):
I like it. 

I'm requesting the changes as I believe the throw inside search should not be there at all. The both callers fill a limit. And if we start to call this function somewhere else we might get unexpected error.

src/utils/actor_search.ts
MQ37: let's make the `limit` required and drop the throw - let's make the API throw instead.

MQ37:
@jirispilka makes sense, the throw was too defensive - let's make the API throw instead.

jirispilka (approved):
Thanks 💪🏻

PR #743, opened 2026-04-23
Support custom task IDs in MCP task store

jirispilka:
## Summary
Implements support for custom task IDs in the MCP task store by introducing `ApifyTaskStore`, which extends the SDK's `InMemoryTaskStore` to allow overriding auto-generated task IDs via the `context.taskId` parameter.

## Key Changes
- **New `ApifyTaskStore` class** (`src/mcp/task_store.ts`): Extends `InMemoryTaskStore` to support custom task IDs passed through `CreateTaskOptions.context.taskId`. Uses a clever workaround to shadow the private `generateTaskId()` method for a single call.
- **Updated server initialization** (`src/mcp/server.ts`): Replaced `InMemoryTaskStore` with `ApifyTaskStore` for stdio transport mode.
- **Improved task creation** (`src/mcp/server.ts`): Modified task creation to generate deterministic task IDs using the pattern `{tool.name}-{random-base64}` and pass them via context instead of as the requestId parameter.
- **Documentation** (`DEVELOPMENT.md`): Added comprehensive testing guide for task mode, including examples for `--task`, `--detach`, `tasks-list`, and `tasks-get` commands.

## Implementation Details
The `ApifyTaskStore` works around the SDK's private `generateTaskId()` method by temporarily replacing it on the instance with a custom function that returns the desired ID, then deletes itself after one invocation. This allows seamless integration with the existing SDK without requiring modifications to the upstream library.

https://claude.ai/code/session_01DCP8NmF4VioXoDRx8esbqG

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -140,6 +140,25 @@ mcpc --json @stdio tools-call search-actors keywords:="scraper" | jq ‘.content
 - `get-actor-output` — test field filtering with dot notation, non-existent dataset
 - `search-apify-docs` / `fetch-apify-docs` — test relevant and non-existent queries
 
+**Testing task mode (long-running calls):**
+
+Use `--task` to request task execution. Use `--detach` to start a task and return immediately with the task ID.
+
+```bash
+# Reject: tool does not support task mode
+mcpc @stdio tools-call search-apify-docs --task query:="web scraping"
+
+# Start call-actor as a task and wait for result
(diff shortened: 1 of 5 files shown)


jirispilka:
Superseded by #809. The runtime monkey-patch on SDK's private `generateTaskId()` worked for stdio but stored tasks under empty-string `taskId` on the hosted Redis path.

PR #755, opened 2026-04-26
docs: Audit MCP integration test coverage vs 2025-11-25 spec

jirispilka:
https://claude.ai/code/session_01K8GodA6iMy8e1xrMQzMnnF


MQ37 (approved):
Makes sense but let's first rework the integration tests to make them more scalable and maintainable - I would definitely not add to the single huge file we already have

Issue #766, opened 2026-04-28
Add automated integration tests for agentic payment headers


MQ37:
We should have automated integration tests verifying that agentic payment headers (Skyfire `skyfire-pay-id`, x402 challenge) are correctly forwarded from MCP clients to the backend — e.g. against a dummy/mock payment endpoint. Currently this is only validated manually, so payment regressions on tools (like missing `paymentRequired: true`) can land unnoticed.

Bortlesboat:
A compact regression matrix that would make this much harder to break:

1. **Free tool baseline**: `paymentRequired: false` tool call forwards normal MCP/tool headers but never forwards `x-payment`, `payment-required`, or `skyfire-pay-id`.
2. **x402 paid tool**: a mock backend returns `402` with `payment-required`; the MCP layer marks the tool as payment-required and preserves the challenge back to the client. On retry, the client's `x-payment` header is forwarded byte-for-byte to the backend.
3. **Skyfire paid tool**: same shape, but assert `skyfire-pay-id` is forwarded to the backend and not dropped by the MCP adapter.
4. **Misconfigured paid backend**: backend returns `402`, but the tool definition is missing `paymentRequired: true`; test should fail loudly so paid tools cannot regress into silent manual-only validation.
5. **Logging/privacy guard**: tests should assert only header presence/rail/failure stage are logged, not raw payment credentials or signed payloads.

I'd keep the main fixture local/mock so CI never spends money. If a live fixture is useful for an optional smoke later, a public x402 MCP target can help, but the regression suite should not depend on an external paid service.

PR #768, opened 2026-04-28
test: Add integration test for x402 _meta advertising on paid tools

MQ37:
Stacked on #734.

## Summary

Adds one Streamable-HTTP integration test asserting that tools with \`paymentRequired: true\` advertise \`_meta.x402\` (with \`scheme\`, \`network\`, \`asset\`, \`payTo\`, \`amount\`) when the server runs with \`?payment=x402\`, and that free tools do not.

The expected paid set is hardcoded in the test rather than imported from \`SKYFIRE_ENABLED_TOOLS\` so that silent drift between \`paymentRequired\` flags and the constant is caught here. This locks in the regression flagged in #734 (where \`get-actor-run\` lost \`paymentRequired: true\`).

Tracks #766 (broader plan for automated agentic-payment integration tests).

## Tests

- \`npm run type-check\` — passes
- \`npm run lint\` — passes
- \`npm run test:unit\` — 615 passed
- \`npm run test:integration\` — human-run per CLAUDE.md

### DIFF
--- a/tests/integration/suite.ts
+++ b/tests/integration/suite.ts
@@ -2614,6 +2614,61 @@ export function createIntegrationTestsSuite(
 
         // x402 payment mode only works with Streamable-HTTP transport (requires HTTP headers).
         it.runIf(options.transport === 'streamable-http')(
+            'should advertise x402 metadata on all paymentRequired tools when x402 payment is enabled',
+            async () => {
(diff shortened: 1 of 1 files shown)


jirispilka (approved):
Thanks 💪🏻

PR #780, opened 2026-05-01
fix: Report initial status and terminal status message (if saved)

jirispilka:
## Summary

Make actor-run progress and abort notifications reliable end-to-end.

- **Abort actually aborts.** `callActorGetDataset` awaits the abort API call before returning on `notifications/cancelled`, so the run is aborted rather than fire-and-forget. The polling tracker is also stopped on the cancel path.
- **First progress arrives immediately.** Initial actor-run status is reported before `waitForFinish`, so clients don't wait for the polling interval to see the first update.
- **Last progress is guaranteed and accurate.** After `waitForFinish` resolves, the tracker is stopped and the run is re-fetched once (in parallel with dataset/build) to capture the actor's final `statusMessage` — the platform may write it just after the status flip. A terminal notification is then emitted explicitly so the client always sees `SUCCEEDED`/`FAILED`/etc., even if polling never ticked at terminal.
- **Status format is terminal-aware.** `formatRunStatusMessage` always leads with the run status (`apify/foo: SUCCEEDED — Done`). At terminal status, the actor's `statusMessage` is only appended when explicitly marked terminal (`isStatusMessageTerminal === true`); otherwise it's omitted to avoid showing stale in-progress text.
- **No duplicate emissions.** Three sources of duplicates closed:
  - Polling dedup state is seeded with the initial run, so the first poll tick doesn't re-emit the same state already sent at start.
  - A `stopped` flag in `ProgressTracker` prevents in-flight polling ticks from emitting after `stop()` is called.
  - `updateProgress` dedups consecutive identical messages, so a polling tick that catches the terminal state doesn't double up with the explicit final emit.
- `PROGRESS_NOTIFICATION_INTERVAL_MS` lowered 5s → 2s for snappier updates.

All tests are passing now

image

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -160,7 +160,7 @@ export const ALLOWED_DOC_DOMAINS = [
     'https://crawlee.dev',
 ] as const;
 
-export const PROGRESS_NOTIFICATION_INTERVAL_MS = 5_000; // 5 seconds
+export const PROGRESS_NOTIFICATION_INTERVAL_MS = 2_000; // 2 seconds
 
 export const APIFY_STORE_URL = 'https://apify.com';
 export const APIFY_FAVICON_URL = `${APIFY_STORE_URL}/favicon.ico`;
--- a/src/const.ts
+++ b/src/const.ts
@@ -160,7 +160,8 @@ export const ALLOWED_DOC_DOMAINS = [
     'https://crawlee.dev',
 ] as const;
 
-export const PROGRESS_NOTIFICATION_INTERVAL_MS = 2_000; // 2 seconds
+// The console uses const MIN_OBSERVER_INTERVAL_MILLIS = 3000 so it should be fine.`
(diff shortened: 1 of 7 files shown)


src/const.ts
```diff
@@ -160,7 +160,7 @@ export const ALLOWED_DOC_DOMAINS = [
     'https://crawlee.dev',
 ] as const;
 
-export const PROGRESS_NOTIFICATION_INTERVAL_MS = 5_000; // 5 seconds
+export const PROGRESS_NOTIFICATION_INTERVAL_MS = 2_000; // 2 seconds
```
MQ37: nit: not sure if this is too aggresive, but the console uses `const MIN_OBSERVER_INTERVAL_MILLIS = 3000` so it should be fine.

MQ37 (approved):

src/const.ts
jirispilka: Ok, changed it to 3 secs

PR #781, opened 2026-05-01
fix: Validate fetch-apify-docs URL by hostname, not prefix

MQ37:
GHSA-jwp7-wg77-3w9v: `url.startsWith(domain)` against allowlist entries
like `https://docs.apify.com` is bypassable by URLs such as
`https://docs.apify.com.evil.com/`, `https://docs.apify.com@evil.com/`,
or `https://docs.apify.com.evil.com:8080/`. The bypass lets the tool
fetch attacker-controlled HTML and return it to the LLM, enabling
prompt injection.

Replace the prefix check with a parsed-URL comparison: derive allowed
hostnames from `ALLOWED_DOC_DOMAINS` once at module load and require
`new URL(url)` to have `protocol === 'https:'` and a hostname in that
set. Add unit coverage for the bypass cases and other rejections.

Fixes https://github.com/apify/apify-mcp-server/security/advisories/GHSA-jwp7-wg77-3w9v

### DIFF
--- a/src/tools/common/fetch_apify_docs.ts
+++ b/src/tools/common/fetch_apify_docs.ts
@@ -32,6 +32,23 @@ export function buildMarkdownUrl(url: string): string {
     return parsed.toString();
 }
 
+const ALLOWED_DOC_HOSTS: ReadonlySet<string> = new Set(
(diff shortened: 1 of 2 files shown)


jirispilka (approved):
Thanks 💪🏻

PR #800, opened 2026-05-06
feat: Add switch to enable/disable auto-mode

jirispilka:
Re-enables the capability-driven server mode auto-detection feature that was previously disabled,

- Added `SERVER_MODE_AUTO_DETECTION_ENABLED` 
- The kill switch is a simple boolean constant that can be flipped to `false` to disable auto-detection without reverting code changes
- Kept disabled (`false`) for now, pending further testing before rollout

@MQ37 I'm sorry for a bigger diff than required, but the server-mode helpers were present in the types.ts 🤯 . My bad. Fixed

### DIFF
--- a/src/utils/server_mode.ts
+++ b/src/utils/server_mode.ts
@@ -1,0 +1,33 @@
+import { SERVER_MODE_AUTO_DETECTION_ENABLED } from '../const.js';
+import { ServerMode, type ServerModeOption } from '../types.js';
+
+/**
+ * Parse an untrusted raw mode string (from CLI flag, env var, or URL param) into a {@link ServerModeOption}.
+ *
+ * Accepts:
+ * - `'default'` / `'apps'` — canonical values
+ * - `'true'` / `'on'` / `'false'` / `'off'` — CLI shorthand
+ * - `'auto'` — resolve from client capabilities (default for missing/unknown input)
+ * - `'openai'` — deprecated alias for `'apps'` (pre-MCP-Apps naming); silently normalized
+ *
+ * Missing or unrecognized input returns `'auto'`, so a typo in an env var becomes
+ * capability-driven resolution instead of silently forcing default mode.
+ */
+export function parseServerMode(rawMode: string | null | undefined): ServerModeOption {
+    if (!rawMode) return 'auto';
+    if (rawMode === 'true' || rawMode === 'on' || rawMode === ServerMode.APPS || rawMode === 'openai') return ServerMode.APPS;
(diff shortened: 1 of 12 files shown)


src/utils/server_mode.ts
```diff
@@ -0,0 +1,33 @@
...
+ * - `'openai'` — deprecated alias for `'apps'` (pre-MCP-Apps naming); silently normalized
+ *
+ * Missing or unrecognized input returns `'auto'`, so a typo in an env var becomes
+ * capability-driven resolution instead of silently forcing default mode.
+ */
+export function parseServerMode(rawMode: string | null | undefined): ServerModeOption {
+    if (!rawMode) return 'auto';
+    if (rawMode === 'true' || rawMode === 'on' || rawMode === ServerMode.APPS || rawMode === 'openai') return ServerMode.APPS;
+    if (rawMode === 'false' || rawMode === 'off' || rawMode === ServerMode.DEFAULT) return ServerMode.DEFAULT;
+    if (rawMode === 'auto') return 'auto';
```
MQ37: change: this condition is not needed when the final case is also `auto`
```suggestion
```

tests/integration/suite.ts
```diff
@@ -5,13 +5,20 @@ import Ajv from 'ajv';
...
+    SERVER_MODE_AUTO_DETECTION_ENABLED,
+    SKYFIRE_ENABLED_TOOLS,
+} from '../../src/const.js';
 import { RESOURCE_MIME_TYPE } from '../../src/resources/widgets.js';
 // Import tools from getCategoryTools instead of directly to avoid circular dependency during module initialization
 import { getCategoryTools, getDefaultTools } from '../../src/tools/index.js';
 import { callActorOutputSchema } from '../../src/tools/structured_output_schemas.js';
 import { actorNameToToolName } from '../../src/tools/utils.js';
-import type { ServerMode, ToolCategory, ToolEntry } from '../../src/types.js';
+import { type ServerMode, type ToolCategory, type ToolEntry } from '../../src/types.js';
```
MQ37: question: why this change? I would be for introducing formatter soon to enforce some code styling standards as these random changes can be then more easily prevented

MQ37 (approved):
Cool, nice and clean PR 👍

src/utils/server_mode.ts
jirispilka: Correct. Fixed

tests/integration/suite.ts
jirispilka: You are right! Fixed. Thanks!

PR #803, opened 2026-05-07
Removed getApifyAPIBaseUrl

### LINKED ISSUE #797: refactor: Pass Apify API base URL as a server option instead of a global env var
## Problem

`getApifyAPIBaseUrl()` reads `process.env.APIFY_API_BASE_URL` as global state. It can't simply be deleted (#803) — the hosted server sets that env var from its config, so removal silently falls back to production.

## Proposed solution

Add `apiBaseUrl?` to `ActorsMcpServer` options. `getApifyAPIBaseUrl()` demotes to its default resolver:

```
apiBaseUrl ?? process.env.APIFY_API_BASE_URL ?? 'https://api.apify.com'
```

Additive — the hosted server keeps working unchanged and migrates on its own schedule. Keep the `APIFY_IS_AT_HOME` branch.

Only the egress consumer needs this option (the apify-client `baseUrl`). The four client-facing consumers are split onto a separately derived public origin by #1176 — **do #1176 first**.

## Out of scope

- Removing `getApifyAPIBaseUrl()` or the env var. Deleting a mechanism the hosted server consumes forces a cross-repo lockstep for no gain; the fallback is one line.
- The hosted-server migration — tracked in apify/apify-mcp-server-internal#711.

saksham365:
fixes #797

### DIFF
--- a/src/apify_client.ts
+++ b/src/apify_client.ts
@@ -23,12 +23,6 @@ function addUserAgent(config: AxiosRequestConfig): AxiosRequestConfig {
     return updatedConfig;
 }
 
-export function getApifyAPIBaseUrl(): string {
-    // Workaround for Actor server where the platform APIFY_API_BASE_URL did not work with getActorDefinition from actors.ts
-    if (process.env.APIFY_IS_AT_HOME) return 'https://api.apify.com';
-    return process.env.APIFY_API_BASE_URL || 'https://api.apify.com';
-}
-
 export class ApifyClient extends _ApifyClient {
     constructor(options: ExtendedApifyClientOptions) {
         /**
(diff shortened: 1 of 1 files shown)


jirispilka:
Hey @saksham365, thank you for the PR,  the implementation is correct. 
When I reviewed it, I realized that the spec is incomplete :( 

getApifyAPIBaseUrl() is still needed because our hosted server sets process.env.APIFY_API_BASE_URL (staging URL) and relies on this function to  pass it into the ApifyClient constructor. Removing it silently breaks staging — all API calls would fall back to the hardcoded production URL in apify-client.
  
The proper fix requires a coordinated change across both repos:
  1. Add baseUrl? to ActorsMcpServer constructor options (public repo)
  2. Thread it through to every new ApifyClient() call
  3. Update the internal repo to pass it explicitly instead of setting the env var
  4. Then remove getApifyAPIBaseUrl()
  
 We'll handle steps 1–3 on our side first. We can close this PR. I'll update the issue. Again, I'm sorry for the bad spec.

jirispilka:
Issue was updated: https://github.com/apify/apify-mcp-server/issues/797 
Let me close this PR and, again, apologizes for a bad spec.

saksham365:
no worries @jirispilka

PR #811, opened 2026-05-08
fix(tools): Omit undefined limit from get-dataset-items output

### LINKED ISSUE #731: Minor: `get-dataset-items` structuredContent may include `limit: undefined`
## Problem

`src/tools/common/get_dataset_items.ts:91-100`:

```ts
const structuredContent = {
    datasetId: parsed.datasetId,
    items: v.items,
    itemCount: v.items.length,
    totalItemCount: v.total,
    offset: parsed.offset ?? 0,
    limit: parsed.limit,   // optional — can be undefined
};
```

`parsed.limit` is optional (no default). When not provided it is `undefined`. JSON serialization omits `undefined` values, so `limit` is absent in the serialized output. The declared `datasetItemsOutputSchema` defines `limit` as `{ type: 'number' }` — a client that reads `structuredContent.limit` gets `undefined` and the schema implies a number.

This is a minor inconsistency between the declared `outputSchema` and the actual structured content.

## Proposed solution

Apply a consistent fallback:

```ts
limit: parsed.limit ?? null,  // or omit entirely when undefined
```

Or align the `datasetItemsOutputSchema` to mark `limit` as optional (which it effectively is).

mvanhorn:
## What

Closes #731

`get-dataset-items` returned `structuredContent` with `limit: undefined` when the caller did not supply a `limit` argument. JSON serialization drops `undefined`, so the wire payload silently omitted the key while `datasetItemsOutputSchema` declared `limit: { type: 'number' }` — a quiet schema/payload mismatch.

## Why

A client reading `structuredContent.limit` against the declared schema would see `undefined` where the schema implies a number is present. The schema already does not list `limit` in `required`, so the well-formed response is one that omits the key when the input was unset, not one that names it with `undefined`.

## How

Conditionally spread `limit` only when defined:

```ts
...(parsed.limit !== undefined ? { limit: parsed.limit } : {}),
```

This keeps `structuredContent` consistent with both the declared schema and the post-serialization payload.

## Tests

Three new vitest cases in `tests/unit/tools.get_dataset_items.test.ts`:

1. `omits limit ... when caller did not provide one` — fails on master, passes here
2. `includes limit ... when caller provided one` — sanity that the present-key path still works
3. `survives a JSON serialization round-trip without exposing undefined` — what wire clients actually see

`npm run type-check`, `npm run lint`, and `npm run test:unit` all pass (616 passed, 4 skipped).

## Risks

Minimal. The schema already declared `limit` as optional (not in `required`); this change only stops emitting the key when its value would be `undefined`. Existing callers that rely on `'limit' in structuredContent` to detect "limit was passed" now have a correct signal.

### DIFF
--- a/src/tools/common/get_dataset_items.ts
+++ b/src/tools/common/get_dataset_items.ts
@@ -88,13 +88,7 @@ export const getDatasetItems: ToolEntry = Object.freeze({
             });
         }
 
-        // Conditionally spread `limit` so the structured content matches the
-        // declared `datasetItemsOutputSchema`. The schema declares `limit` as
-        // `{ type: 'number' }` and does not list it as required, so omitting
-        // the key when the caller did not supply one keeps the response
-        // well-formed instead of carrying `limit: undefined` (which JSON
-        // serialization drops anyway, leaving a silent mismatch between
-        // schema and payload). See issue #731.
+        // omit limit when unset; undefined silently drops from JSON (#731)
         const structuredContent = {
             datasetId: parsed.datasetId,
             items: v.items,
--- a/src/tools/common/get_dataset_items.ts
+++ b/src/tools/common/get_dataset_items.ts
@@ -88,13 +88,20 @@ export const getDatasetItems: ToolEntry = Object.freeze({
             });
         }
 
+        // Conditionally spread `limit` so the structured content matches the
+        // declared `datasetItemsOutputSchema`. The schema declares `limit` as
+        // `{ type: 'number' }` and does not list it as required, so omitting
+        // the key when the caller did not supply one keeps the response
+        // well-formed instead of carrying `limit: undefined` (which JSON
+        // serialization drops anyway, leaving a silent mismatch between
+        // schema and payload). See issue #731.
         const structuredContent = {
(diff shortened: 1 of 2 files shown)


jirispilka (changes requested):
Good fix, the one-liner is exactly right.

Thanks!

Two small things to clean up per our conventions:

- 7-line comment above the fix — explains what the code already shows. One line max, e.g.: // omit limit when unset; undefined silently drops from JSON (#731)
- Test 3 (survives a JSON serialization round-trip) — redundant. Test 1 already confirms limit is absent from the object; serialization can't add it back. Drop it.
- Docstring at top of test file — multi-line comment blocks aren't our style. Remove it.

mvanhorn:
Cleaned up in d0efcfa:

- Trimmed the 7-line comment to one line: `// omit limit when unset; undefined silently drops from JSON (#731)`
- Dropped the JSON round-trip test (test 1 already confirms the key is absent on the object)
- Removed the test-file docstring

Type-check, lint, and 615 unit tests green.

jirispilka (approved):

PR #812, opened 2026-05-08
fix: Abort in-flight tool handler on tasks/cancel (#763)

### LINKED ISSUE #763: fix: tasks/cancel must abort the running tool handler
**Base branch:** `claude/implement-get-actor-run-MUTXI` (or master once merged).

Sub-issue of #582 — standalone bug, useful regardless of whether #582 lands. The task-mode `call-actor` design in #582 assumes this chain works.

## Bug

`tasks/cancel` marks the task `cancelled` in `taskStore`, but the in-flight tool handler keeps running.

`src/mcp/server.ts:executeToolAndUpdateTask` checks `isTaskCancelled` once before calling `tool.call(...)`. After that the handler runs to completion with the original `extra.signal`, which is **not** wired to the taskStore cancel state. For an Actor-running tool this means the Apify run keeps consuming compute until natural completion even after the client cancels.

## Scope of the bug

Affects every Actor-running tool, not just `call-actor`:

- Default `call-actor` and apps `call-actor` / `call-actor-widget` — call `actorClient.start()` + `waitForFinish()`
- Direct actor tools (RAG Web Browser et al., built via `src/tools/core/actor_tools_factory.ts`) — handlers call `callActorGetDataset` which already has the same `Promise.race(CLIENT_ABORT)` receiver
- Quick tools (`search-actors`, `fetch-actor-details`, `get-actor-output`, docs tools) don't have a sustained wait loop, so cancel is moot mid-execution

The fix is in the **transmitter** side — `executeToolAndUpdateTask` doesn't propagate taskStore-cancel into the handler's `AbortSignal`. The receiver side (`callActorGetDataset`'s abort race) already exists and works for both call-actor and direct actor tools. Wire the transmitter once, all paths start working.

## Test first

Add failing integration cases in `tests/integration/suite.ts` (gated to streamable-http where tasks are exposed). Each spins up RAG Web Browser long enough that the Actor cannot finish before cancel:

1. **call-actor path:** call `call-actor` for `apify/rag-web-browser` with `{ query: "long arbitrary search", maxResults: 50 }` under `tools/call` + `task` params. Once the task transitions to `working`, call `tasks/cancel`. Assert the Actor run reaches `ABORTED` within ~5 s (poll `apifyClient.run(runId).get()`).
2. **direct actor tool path:** call the registered `apify/rag-web-browser` actor tool with the same input under `tools/call` + `task` params. Same cancel + abort assertion.

Confirm both fail on `claude/implement-get-actor-run-MUTXI` before fixing.

## Fix

`src/mcp/server.ts:executeToolAndUpdateTask`:

- Create a derived `AbortController` chained from `extra.signal`.
- Spin up an interval (e.g. 500 ms) that calls `isTaskCancelled(taskStore, taskId)`; on `true`, call `derivedController.abort()` and clear the interval.
- Pass `{ ...extra, signal: derivedController.signal }` into `tool.call`.
- Clear the interval in `finally`.

Reuse `isTaskCancelled` from `src/mcp/utils.ts`.

The handler-side abort handling already exists for both call-actor and direct actor tools via the `CLIENT_ABORT` race in `callActorGetDataset`, so once the signal fires the Apify run will be aborted.

## Plan

- [ ] Failing integration tests reproducing the bug (call-actor + direct actor tool)
- [ ] AbortController watcher in `executeToolAndUpdateTask`
- [ ] Confirm both tests pass
- [ ] Unit test: cancel during a fake long handler aborts its signal

## Files touched

- `src/mcp/server.ts`
- `tests/integration/suite.ts`
- `tests/unit/...` (optional, if cheap)

## Acceptance

- [ ] `npm run type-check`, `npm run lint`, `npm run test:unit` clean
- [ ] Both new integration tests pass; both fail on the base branch
- [ ] No public tool input or output schema change
- [ ] `apify-mcp-server-internal` unaffected

## Independence

Independent of #__A__ (refactor) and #__C__ (feat). Can ship today.

## Estimate

~2 h (mostly the integration tests).

jirispilka:
fixes: https://github.com/apify/apify-mcp-server/issues/763

tasks/cancel marks the task `cancelled` in the store but the in-flight
handler keeps running, leaving the underlying Apify run consuming compute
until natural completion. The MCP SDK's tasks/cancel handler only writes
status to the store — it does not abort the request's AbortController.

### Before

```mermaid
sequenceDiagram
    participant C as Client
    participant S as Server
    participant A as Apify
    C->>S: tasks/cancel
    S-->>C: ok (status=cancelled in store)
    Note over S,A: handler keeps running
    A-->>S: run finishes (compute wasted)
```

### After

```mermaid
sequenceDiagram
    participant C as Client
    participant S as Server
    participant W as Watcher
    participant A as Apify
    C->>S: tasks/cancel
    S-->>C: ok (status=cancelled in store)
    loop every 500 ms
        W->>S: getTask
    end
    W->>W: detects cancelled, aborts signal
    W->>A: run.abort()
```

`createTaskCancellationWatcher` polls the TaskStore and exposes a derived AbortSignal. Threaded through `executeToolAndUpdateTask` into both the internal-tool and direct-actor-tool branches. The abort listener already in `callActorGetDataset` then issues `apifyClient.run(runId).abort()`.

Polling rather than a callback because in multi-node deployments the cancel may arrive on a different pod than the one running the handler — the shared TaskStore is the only signal the executing pod can observe. Per MCP tasks spec, the original request's `extra.signal` is intentionally NOT chained: tasks are decoupled from the request, so client disconnect must not cancel a task.

See [`res/tasks_cancel_abort_flow.md`](./res/tasks_cancel_abort_flow.md) for full design (end-to-end signal path, hardening notes, spec compliance audit).

  mcpc end-to-end:
  - search-actors, fetch-actor-details, search-apify-docs: all return valid responses
  - Task cancellation flow (the core PR feature):
    - call-actor with --detach creates an MCP task and returns taskId immediately
    - tasks/cancel flips status to cancelled with message "Cancelled by client"
    - tasks/get confirms status: cancelled ~4.5s after cancel (consistent with the 500ms poll
   cycle in createTaskCancellationWatcher)

  The watcher is working as designed: polls every 500ms, aborts the in-flight handler, and
  the task stays cancelled on subsequent reads

https://claude.ai/code/session_01TDzAvK4kneEpDLwqAdQNk8


src/mcp/server.ts
```diff
@@ -1307,6 +1320,8 @@ export class ActorsMcpServer {
                 taskId,
                 mcpSessionId,
             });
+            // TODO(TC-3): TOCTOU with the check above — cancel arriving here attempts
+            // cancelled → working, which the store rejects. Catch the terminal-state error.
             await this.taskStore.updateTaskStatus(taskId, 'working', undefined, mcpSessionId);
```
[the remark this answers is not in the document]
jirispilka: I acknowledged the race, not fixing int this PR

src/mcp/server.ts
MQ37: change: since we introduced the watcher I think we can remove this pre-check if the task was cancelled

src/mcp/server.ts
```diff
@@ -1414,6 +1431,13 @@ export class ActorsMcpServer {
             const httpStatus = getHttpStatusCode(error);
             if (httpStatus === HTTP_PAYMENT_REQUIRED) {
                 logHttpError(error, 'Payment required while calling tool (task mode)', { toolName: tool.name });
+                // Per MCP tasks spec: once a task is cancelled it MUST remain cancelled,
+                // so guard storeTaskResult against a cancel that raced with this 402.
+                if (await isTaskCancelled(taskId, mcpSessionId, this.taskStore)) {
+                    log.debug('[executeToolAndUpdateTask] Task was cancelled, skipping 402 result storage', { taskId, mcpSessionId });
+                    finishTaskTracking(TOOL_STATUS.ABORTED, { ...buildActorFields(actorName, actorId) });
+                    return;
+                }
```
MQ37: nit: there is quite a lot of duplicate code using the `isTaskCancelled -> finishTaskTracking`, I think we should extract that to some helper func

src/mcp/utils.ts
```diff
@@ -48,3 +48,75 @@ export async function isTaskCancelled(
...
+        if (tickInFlight || controller.signal.aborted) return;
+        tickInFlight = true;
+        void (async () => {
+            try {
+                if (await isTaskCancelled(taskId, mcpSessionId, taskStore)) {
+                    // Stop the timer immediately rather than relying on dispose() —
+                    // otherwise ticks keep firing as no-ops until the caller's
+                    // finally block runs.
+                    clearInterval(interval);
+                    controller.abort(new Error(`Task ${taskId} cancelled by client`));
```
MQ37: question: why the `new Error`? this pattern is never used in the existing code before. Is that even read or surfaced anywhere?

src/mcp/utils.ts
```diff
@@ -48,3 +48,75 @@ export async function isTaskCancelled(
...
+ * `tasks/cancel` may arrive on a different node than the one running the
+ * handler. The shared TaskStore is the only signal the executing node can
+ * observe — so it must poll. 500 ms is a deliberate compromise between
+ * cancellation latency and Redis load.
+ *
+ * Returns `{ signal, dispose }`. Callers MUST call `dispose()` to stop the
+ * polling interval; otherwise it leaks for the lifetime of the process.
+ */
+export function createTaskCancellationWatcher(opts: {
+    taskId: string;
```
MQ37, line 62: nit: I think this is quite verbose for a docstring, maybe we can just reference the spec using a link? If not let's keep that - not a deal breaker

src/mcp/utils.ts
```diff
@@ -48,3 +48,75 @@ export async function isTaskCancelled(
...
+    taskStore: TaskStore;
+    pollIntervalMs?: number;
+}): { signal: AbortSignal; dispose: () => void } {
+    const { taskId, mcpSessionId, taskStore, pollIntervalMs = 500 } = opts;
+    const controller = new AbortController();
+
+    // `tickInFlight` prevents tick overlap when `getTask` is slower than the
+    // poll interval (Redis tail latency, cluster reslot). Without it, ticks
+    // pile up and amplify backend load right when the backend is struggling.
+    let tickInFlight = false;
```
MQ37: nit: what about `pollInProgress` or `tickInProgress`? The "inFlight" part is confusing

MQ37 (approved):
Great and thought through PR overall!

I would just remove the task cancelled pre-checks that the watcher can not handle to simplify the implementation and then there is few nits.

src/mcp/server.ts
jirispilka: Replaced the pre-check with a try/catch around `updateTaskStatus` — catches the `cancelled → working` rejection and treats it as a clean ABORTED. Closes the TC-3 TOCTOU the pre-check left open. Existing "cancel midway" integration test exercises both paths.

src/mcp/utils.ts
jirispilka: Renamed to `tickInProgress` in both `src/mcp/utils.ts` and `src/utils/progress.ts` (the existing copy of the pattern).

src/mcp/server.ts
jirispilka: Good catch: add `skipIfTaskCancelled`

src/mcp/utils.ts
jirispilka: Agree, claude code 🤦🏻

src/mcp/utils.ts
jirispilka: Removed

jirispilka:
> Great and thought through PR overall!
> 
> I would just remove the task cancelled pre-checks that the watcher can not handle to simplify the implementation and then there is few nits.

Thanks for the comments, they made this PR better!

PR #814, opened 2026-05-08
feat: Auto-inject storage and abort tools alongside actor tools

### LINKED ISSUE #813: feat: Storage tools accept runId; auto-inject in actor workflows
## What

Make `get-dataset-items` and `get-key-value-store-record` first-class for Actor-run workflows: accept `runId` as an alternative to the storage ID; auto-derive `flatten` from dot-notation `fields`; default `limit: 100`. Auto-inject these and `abort-actor-run` alongside `call-actor` so they're always reachable when the agent has a runId. Mark `get-actor-output` deprecated (description-only) and reorder it after `get-dataset-items`.

This is the **Wave 1** prerequisite that lets the v4 canonical-shape PRs (E and F) emit `nextStep` instructions like *"Use `get-dataset-items` with `runId=...`"* without breaking the agent flow.

## Spec

Authoritative spec: [`res/call_actor_redesign_v4.md`](https://github.com/apify/apify-mcp-server/blob/master/res/call_actor_redesign_v4.md). Relevant sections:

- **Tool surface** — `get-dataset-items`, `get-key-value-store-record`, `abort-actor-run` rows.
- **`get-dataset-items` — substantive change** — runId, auto-flatten, default limit.
- **`get-key-value-store-record` — promoted** — runId.
- **Locked decisions Q3, Q4, Q6**.

Do not implement anything from this issue beyond what those sections describe. v4 canonical-response-shape work belongs in **E** and **F** — not here.

## Current state on master (verified)

- `src/tools/common/get_dataset_items.ts` — accepts `datasetId` only. `fields` and `flatten` are separate parameters; the description mentions dot notation but no auto-derivation. No default `limit`.
- `src/tools/common/get_key_value_store_record.ts` — accepts `storeId` + `recordKey` only. No `runId`.
- `src/tools/common/abort_actor_run.ts` — takes `runId`, already implemented and in the `runs` category.
- `src/utils/tools_loader.ts:244-271` — auto-injects only `defaultGetActorRun` and `getActorOutput` after `call-actor` / actor tool / `add-actor`. Comment at line 244 documents this.
- `src/tools/common/get_actor_output.ts:70-87` — description currently positions itself as a "simplified version of `get-dataset-items` focused on Actor run outputs". Not deprecated.

### 1. `get-dataset-items` (`src/tools/common/get_dataset_items.ts`)

- Add `runId` to the Zod schema as an alternative to `datasetId`. Use `.refine(...)` so exactly one of the two is required (reject neither / both with a clear validation error).
- When `runId` is provided, resolve the run's `defaultDatasetId` via `apifyClient.run(runId).get()`. If the run lookup fails or the run has no dataset, return a soft-fail response naming the failure cause.
- Auto-derive `flatten`: when caller-supplied `flatten` is absent and any entry in `fields` contains a `.`, compute `flatten` as the unique top-level prefixes (e.g. `["metadata.url", "crawl.statusCode", "title"]` → `["metadata", "crawl"]`) and pass to `apifyClient.dataset(...).listItems(...)`. Explicit `flatten` from the caller wins (diagnostic escape hatch); document this in the parameter description.
- Default `limit` to `100` when omitted (Zod `.default(100)`).
- Update the tool description to reflect: `runId | datasetId` (exactly one), auto-flatten behavior, and the new default. Keep it short.

### 2. `get-key-value-store-record` (`src/tools/common/get_key_value_store_record.ts`)

- Add `runId` to the Zod schema as an alternative to `storeId`. Same refine pattern as above.
- When `runId` is provided, resolve `apifyClient.run(runId).get().defaultKeyValueStoreId`. Same soft-fail on lookup failure.
- Keep `recordKey` (parameter name unchanged per locked decision).
- Update the tool description.

### 3. `get-actor-output` deprecation (`src/tools/common/get_actor_output.ts`)

- Description-only change: prefix with `DEPRECATED: Use \`${HelperTools.DATASET_GET_ITEMS}\` instead.` and rewrite the body to point at `get-dataset-items` (via `runId` or `datasetId`). Keep the tool callable for one minor cycle.
- No code-path changes.

### 4. Auto-injection (`src/utils/tools_loader.ts:244-271`)

- Extend the existing block. Whenever the current condition `(hasCallActor || hasActorTools || hasAddActorTool)` holds (i.e. the same trigger that already injects `getActorOutput`), also inject:
  - `getDatasetItems`
  - `getKeyValueStoreRecord`
  - `abortActorRun`
- Final order after `call-actor`: `defaultGetActorRun` → `getDatasetItems` → `getKeyValueStoreRecord` → `abortActorRun` → `getActorOutput` (deprecated, last).
- The de-dup pass at the end of the function already protects against double-injection if the user explicitly enabled the `runs` or `storage` categories.

## Out of scope

- v4 canonical response shape on `call-actor` / `get-actor-run` — separate issues E and F.
- `nextStep` / `summary` template wiring — comes with E and F.
- Removing `get-actor-output` — that's the next minor cycle.
- `waitSecs` on either tool — only relevant once the canonical shape lands.
- Any changes to widget React.

## Surface

- `src/tools/common/get_dataset_items.ts`
- `src/tools/common/get_key_value_store_record.ts`
- `src/tools/common/get_actor_output.ts` (description only)
- `src/utils/tools_loader.ts`
- `tests/unit/...` — schema validation (refine on runId/storageId), auto-flatten derivation
- `tests/integration/suite.ts` — add cases per below

## Acceptance

- [ ] `npm run type-check`, `npm run lint`, `npm run test:unit` clean.
- [ ] Integration suite covers the cases below (single source of truth — do **not** add separate test files per CLAUDE.md).
- [ ] No public schema change on `call-actor` / `get-actor-run` (those are E/F).
- [ ] `apify-mcp-server-internal` unaffected — verify no internal caller breaks on the new optional `runId` parameter or on the `get-actor-output` description change.

### Test cases (in `tests/integration/suite.ts`)

1. `get-dataset-items` with `runId` of a finished run resolves the default dataset and returns items.
2. `get-dataset-items` with both `runId` and `datasetId` is rejected with a validation error naming the conflict.
3. `get-dataset-items` with neither `runId` nor `datasetId` is rejected.
4. `get-dataset-items` with `fields: "metadata.url"` and no explicit `flatten` returns the nested values (auto-derive works end-to-end against Apify).
5. `get-dataset-items` with `fields: "metadata.url"` and explicit `flatten: "other"` honors the explicit value (escape hatch works).
6. `get-dataset-items` without `limit` returns at most 100 items.
7. `get-key-value-store-record` with `runId` + `recordKey: "INPUT"` returns the input record (works for any actor that writes INPUT, e.g. rag-web-browser).
8. `get-key-value-store-record` with both `runId` and `storeId` is rejected.
9. `tools/list` in default mode with `call-actor` enabled includes `get-dataset-items`, `get-key-value-store-record`, `abort-actor-run` in the expected order; `get-actor-output` description starts with `DEPRECATED:`.
10. `tools/list` with no Actor-touching tools (e.g. only `docs`) does **not** auto-inject any of the three.

## mcpc end-to-end (mandatory)

After implementation, build and probe the local stdio server end-to-end. Document the exact invocations and observed JSON in the PR description.

```bash
npm run build
mcpc @stdio restart
```

Then exercise (assumes `APIFY_TOKEN` set in the server env — agents must ask the developer for one, never log or commit):

```bash
# 1) Confirm auto-injection
mcpc @stdio tools-list --json | jq '.result.tools[].name' \
  # expect: get-dataset-items, get-key-value-store-record, abort-actor-run, get-actor-output (last, DEPRECATED in description)

# 2) Run an actor to get a runId
mcpc @stdio tools-call call-actor actor:='"apify/rag-web-browser"' \
  input:='{"query":"hello world","maxResults":3}' --json | jq '.result.structuredContent'
# capture runId for the calls below

# 3) get-dataset-items with runId + dot-notation fields (no explicit flatten)
mcpc @stdio tools-call get-dataset-items \
  runId:='"<runId>"' fields:='"metadata.url,markdown"' --json \
  | jq '.result.structuredContent.items[0]'
# expect: nested fields surfaced (e.g. "metadata.url": "...") — auto-flatten worked

# 4) get-key-value-store-record with runId
mcpc @stdio tools-call get-key-value-store-record \
  runId:='"<runId>"' recordKey:='"INPUT"' --json | jq '.result.content[0].text'
# expect: the JSON we passed in step 2

# 5) Negative: both runId and datasetId rejected
mcpc @stdio tools-call get-dataset-items \
  runId:='"<runId>"' datasetId:='"x"' --json | jq '.result.isError'
# expect: true with a message naming the conflict
```

## Independence

Wave 1 — independent. Lands against current master. Does not depend on the cancel fix (#763 / PR #812) or any other in-flight work. Unblocks the `nextStep` text in E/F: once this lands, those PRs can emit `runId`-based hints without hitting validation errors at the storage tools.

## Estimate

~3–4 h (most of the time goes into the `runId` resolution paths and the integration cases).

jirispilka:
Auto-injects storage and abort helpers next to `get-actor-run` whenever any actor-running tool is present, so the LLM doesn't need to opt into the `storage` / `runs` categories to follow up on a run. 

I decide to rename `storeId` to `keyValueStoreId` as `storeId` can be anything, like is it apify/store id? This makes it explicit. 

- `get-actor-output` is deprecated but kept for the time being.

- **Auto-inject 4 helpers** when `call-actor`, an actor tool, or `add-actor` is present: `get-dataset-items`, `get-key-value-store-record`, `abort-actor-run`, and `get-actor-output` (last, deprecated). De-duped against explicit category selection.
- **Default `limit: 20` on `get-dataset-items`** to prevent accidental large reads; agents opt up by passing an explicit `limit`. LLMs tend to omit the parameter, so a low ceiling is the safe default.
- **Auto-flatten from dot-notation `fields`.** Apify's `flatten` is a foot-gun: pass `fields="metadata.url"` without `flatten="metadata"` and you silently get empty rows. The tool now derives the flatten set from `fields`. Explicit `flatten` still overrides as a diagnostic escape hatch.
- **`get-actor-output` deprecated.** Description prefixed `DEPRECATED:` and points at `get-dataset-items`. Kept callable for one minor cycle, removed next; `get-dataset-items` covers the same use case (auto-flatten, lower default limit, full pagination/projection).
- **Design doc updated separately** in #819 (merged): storage tools take their own IDs (no \`runId\` alternative); \`nextStep\` interpolates storage IDs and embeds field/key context for text-mode clients.

Closes: #813


src/tools/common/get_actor_output.ts
```diff
@@ -68,16 +68,18 @@ export const getActorOutput: ToolEntry = Object.freeze({
     type: 'internal',
     name: HelperTools.ACTOR_OUTPUT_GET,
     description: dedent`
+        DEPRECATED: Use \`${HelperTools.DATASET_GET_ITEMS}\` instead.
```
MQ37, line 71: important: let's remove this tool from all categories and the auto-loading logic, keep it for now and after we release then remove this tool completely in another release. There is no point in having this tool as we can use the `get-dataset-items`.  **Existing MCP session might have this session in context so I would keep it for this PR/release**. But please **make an issue so we do not forget to remove this tool.**

tests/integration/suite.ts
```diff
@@ -288,12 +291,16 @@ export function createIntegrationTestsSuite(
...
             const names = getToolNames(await client.listTools());
-            expect(names.length).toEqual(2);
+            // add-actor triggers storage/abort/output but NOT get-actor-run (no call-actor; not apps mode).
+            expect(names.length).toEqual(5);
             expect(names).toContain('add-actor');
-            expect(names).toContain('get-actor-output');
+            expect(names).toContain(HelperTools.DATASET_GET_ITEMS);
+            expect(names).toContain(HelperTools.KEY_VALUE_STORE_RECORD_GET);
+            expect(names).toContain(HelperTools.ACTOR_RUNS_ABORT);
+            expect(names).toContain(HelperTools.ACTOR_OUTPUT_GET);
```
MQ37: change: please let's not repeat this all over the place, either create a const for the auto injected tools or if we just auto inject all the tools from the `storage` category then we could just use the tools category key and loop through.

tests/unit/utils.tools_loader.test.ts
```diff
@@ -65,6 +65,54 @@ describe('toolNamesToInput', () => {
...
+
+    it('does not auto-inject storage or abort tools when no actor-touching tools are present', async () => {
+        const tools = await loadToolsFromInput({ tools: ['docs'] }, apifyClient);
+        const toolNames = tools.map((t) => t.name);
+
+        expect(toolNames).not.toContain(HelperTools.ACTOR_CALL);
+        expect(toolNames).not.toContain(HelperTools.DATASET_GET_ITEMS);
+        expect(toolNames).not.toContain(HelperTools.KEY_VALUE_STORE_RECORD_GET);
+        expect(toolNames).not.toContain(HelperTools.ACTOR_RUNS_ABORT);
+        expect(toolNames).not.toContain(HelperTools.ACTOR_OUTPUT_GET);
```
MQ37: change: same here, the pattern is also repeated here

MQ37 (changes requested):
Thanks, I think this is a tiny step in the right direction! I would consider auto injecting all the tools from `storage` category and not just the selected 4.

There is important change thus I am requesting changes - please remove auto-injecting of the `get-actor-output` as it is deprecated and in another release after this one we will completely remove this tool as the `get-dataset-items` should replace that.

Then there is some repetitive pattern in tests that I would abstract instead of copying all over the place.

src/tools/common/get_actor_output.ts
jirispilka: I was to defensive, you are right, let's remove it.

But, for one cycle, I would keep it in `storage`, remove only from auto-injection.

Removal would break all existing sessions during deployment.

- Removing from **all** categories also drops it from `toolsByName`, so `tools: ['get-actor-output']` and stored session tool names stop resolving
- Can't silently redirect to `get-dataset-items` — different defaults (`clean`, `limit`), local dot-notation projection, empty-property stripping, output truncation. Existing callers' output shape would change.

Effect: default mode no longer pushes it; anyone who names the tool explicitly) still gets it. Will file a follow-up issue for full removal next cycle.

tests/integration/suite.ts
jirispilka: Exported `AUTO_INJECTED_TOOL_NAMES` from `tools_loader.ts` and looped over it in tests (couldn't reuse the `storage` category as-is since `abort-actor-run` lives in `runs`, but the const reads the same).

tests/unit/utils.tools_loader.test.ts
jirispilka: Done. Same const used here too.

src/tools/common/get_actor_output.ts
jirispilka: Done in 5d8440f. Removed from auto-injection, kept in `storage` category. Follow-up: #828 tracks the full removal next cycle.

jirispilka:
> Thanks, I think this is a tiny step in the right direction! I would consider auto injecting all the tools from `storage` category and not just the selected 4.
> 
> There is important change thus I am requesting changes - please remove auto-injecting of the `get-actor-output` as it is deprecated and in another release after this one we will completely remove this tool as the `get-dataset-items` should replace that.
> 
> Then there is some repetitive pattern in tests that I would abstract instead of copying all over the place.

Thanks, I've tried to fix all thecomments
Inject: `getDatasetItems`, ` getKeyValueStoreRecord`, and `abortActorRun`
Remove from the auto-injection: `get-actor-output

Follow up issue: https://github.com/apify/apify-mcp-server/issues/828

MQ37 (approved):

jirispilka:
@MQ37 ok, ok, so you saved this PR with integration tests in CI :)
I think it still holds that integration tests are slow and painful in CI but it saved me this time :) 

image

PR #818, opened 2026-05-08
feat: Add task status notifications for long-running tools

### LINKED ISSUE #816: feat: Push notifications/tasks/status + heartbeat + _meta.related-task on tasks/result
## What

Wire MCP-spec task push notifications and the `tasks/result` `_meta` requirement:

1. Emit `notifications/tasks/status` whenever a task's state changes (every transition + every `statusMessage` update during `working`).
2. Heartbeat the same notification every ~45 s of silence while the task is still `working`, so clients with a 60 s tool-call timeout don't drop the connection. Heartbeats advance `lastUpdatedAt`.
3. Add `_meta["io.modelcontextprotocol/related-task"] = { taskId }` on the `CallToolResult` returned through `tasks/result`.

This is **Wave 1**, fully independent of the v4 canonical-shape work. The ground state is already in place — `progressTracker` runs, `onStatusMessage` updates the task store on every Apify run change — we just don't emit task-status notifications or the related-task metadata on results.

## Spec

Authoritative spec: [`res/call_actor_redesign_v4.md`](https://github.com/apify/apify-mcp-server/blob/master/res/call_actor_redesign_v4.md), sections **Push notifications during sync waits**, **Task mode**, **MCP spec compliance** (R1, the `_meta.related-task` MUST item), and **What we will test** (the bullets covering tasks/status, heartbeat, related-task on tasks/result).

Plus the [MCP tasks specification](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks) MUSTs for `_meta["io.modelcontextprotocol/related-task"]` on results, and the SDK shape `TaskStatusNotificationParams = NotificationParams & Task`.

## Current state on master (verified)

Already correct, keep it that way:

- `src/utils/progress.ts:66-73` — progress notifications already carry `_meta[RELATED_TASK_META_KEY] = { taskId }`. ✅
- `src/const.ts:224` — `RELATED_TASK_META_KEY = 'io.modelcontextprotocol/related-task'` exists.
- `src/mcp/server.ts:1326-1331` — `onStatusMessage` callback already pumps Apify run `statusMessage` into the task store on every change. The hook we need for tasks/status emission is right there.

Bug — needs implementing:

- `src/mcp/server.ts:1328` — explicit TODO: *"Also send notifications/tasks/status so clients get real-time push updates"*. Today nothing emits `notifications/tasks/status`.
- No heartbeat. A long-running task with no Apify-side `statusMessage` updates will sit silent indefinitely.
- `src/mcp/server.ts:682-703` — `GetTaskPayloadRequestSchema` handler returns the raw stored result via `taskStore.getTaskResult(...)`; no `_meta.related-task` is added at that boundary.

## Scope (surgical)

In `src/mcp/server.ts`:

### 1. Emit `notifications/tasks/status` on every state change

Inside `executeToolAndUpdateTask`, extend the existing `onStatusMessage` callback (line 1329) so that *after* the task store update succeeds, it also calls `extra.sendNotification(...)` with a `notifications/tasks/status` notification. The notification payload must be the full `Task` object per the spec (`taskId`, `status`, `createdAt`, `lastUpdatedAt`, `ttl`, optional `statusMessage` / `pollInterval`). Read it back from the task store right after the update so timestamps are fresh.

Also emit on the explicit transitions already happening in this function:

- After `updateTaskStatus(taskId, 'working', ...)` at line 1310 (initial transition).
- After `storeTaskResult(taskId, 'completed', ...)` at line 1408 (terminal completed).
- In every error branch where the task transitions to `failed`.
- On the `cancelled` transition surfaced by the AbortController watcher landed in #763 / PR #812.

Don't emit when the task is already terminal (no-op for the cases where `isTaskCancelled` short-circuits).

### 2. ~45 s heartbeat

Spin up a heartbeat interval at the start of execution (alongside or near the `progressTracker` setup at line 1335 / 1364). On each tick:

- If the last `notifications/tasks/status` emission for this task was < 45 s ago, skip.
- Otherwise: re-read the task from the store, call `updateTaskStatus(taskId, 'working', task.statusMessage, mcpSessionId)` to advance `lastUpdatedAt`, and emit a fresh `notifications/tasks/status` with the updated `Task` object. The heartbeat is itself a state change.

Clear the interval in `finally`. The heartbeat must not fire after the task reaches a terminal state.

Implementation note: tracking "last emission time" can live as a small per-execution variable shared between the explicit emissions in (1) and the heartbeat tick. Keep it local — no new module needed.

### 3. `_meta.related-task` on `tasks/result`

In `src/mcp/server.ts:682-703` (`GetTaskPayloadRequestSchema` handler), wrap the returned result with `_meta[RELATED_TASK_META_KEY] = { taskId }`. The merge must coexist with any existing `_meta` keys the tool already emits (notably `com.apify/ActorRun` for usage attribution from #775). Do not overwrite — spread.

Equivalent alternative: add the `_meta` key inside `taskStore.storeTaskResult` at line 1408 so the stored payload is already correct. Either works; pick the smaller diff.

## Out of scope

- Anything touching `call-actor` / `get-actor-run` response shape — that's E and F.
- `notifications/progress` changes — already correctly wired.
- `taskSupport: "optional"` declaration on `call-actor` — that ships with F.
- Telemetry counters for heartbeat emissions.
- Heartbeat-interval tuning. v4 locks the value at ~45 s (under the 60 s client-side timeout); leave it as a named constant in `src/const.ts` so it's tunable later.

## Surface

- `src/mcp/server.ts` — emission sites (state transitions + heartbeat + `_meta.related-task` on `tasks/result`).
- `src/const.ts` — `TASK_STATUS_HEARTBEAT_INTERVAL_MS = 45_000` constant.
- `tests/unit/...` — heartbeat scheduler logic with a fake timer; `_meta.related-task` injection on a mock `tasks/result` payload.
- `tests/integration/suite.ts` — push-notification test cases (single source of truth, per CLAUDE.md).

## Acceptance

- [ ] `npm run type-check`, `npm run lint`, `npm run test:unit` clean.
- [ ] Integration suite covers the cases below.
- [ ] No public schema change on any tool.
- [ ] `apify-mcp-server-internal` unaffected — verify the additional notifications don't break internal task-aware test harnesses.

### Test cases (in `tests/integration/suite.ts`, gated to streamable-http where tasks are exposed)

1. A task running `apify/rag-web-browser` long enough to produce a real `statusMessage` change emits at least one `notifications/tasks/status` whose payload matches the full `Task` shape (taskId, status, createdAt, lastUpdatedAt, ttl).
2. Initial `working`, terminal `completed`, and `cancelled` transitions each produce exactly one `notifications/tasks/status`.
3. **Heartbeat:** with a test-mode override that drops the interval to ~2 s (so the suite stays fast), a long-running task with no underlying Apify status changes still emits a heartbeat tick, and that tick's `lastUpdatedAt` is strictly greater than the previous emission's.
4. The heartbeat does **not** fire after the task reaches a terminal state.
5. `tasks/result` for a completed task returns a `CallToolResult` whose `_meta` contains `"io.modelcontextprotocol/related-task": { taskId }`. Existing `_meta["com.apify/ActorRun"]` (when present) survives the merge.
6. Cancelled task: emits `notifications/tasks/status` with `status: "cancelled"` after the `tasks/cancel` response (regression coverage for the boundary set up in PR #812).

## mcpc end-to-end (mandatory)

After implementation, build and probe via mcpc against the local stdio server. Streamable-HTTP is where `tasks/*` is exposed — drive that harness for the notification stream. Document exact invocations and expected output in the PR description.

```bash
npm run build
mcpc @stdio restart
```

```bash
# 1) Capture notifications during a long-running task run
mcpc tools-call call-actor \
  actor:='"apify/rag-web-browser"' \
  input:='{"query":"some long search","maxResults":50}' \
  --task --capture-notifications --json | jq '.notifications[] | select(.method == "notifications/tasks/status")'
# expect: at least one entry; .params has full Task shape

# 2) Heartbeat under test-mode override
TASK_STATUS_HEARTBEAT_INTERVAL_MS=2000 mcpc tools-call call-actor \
  actor:='"apify/rag-web-browser"' \
  input:='{"query":"slow search","maxResults":1}' \
  --task --capture-notifications --json \
  | jq '[.notifications[] | select(.method == "notifications/tasks/status")] | length'
# expect: > 1 (heartbeats fire)

# 3) tasks/result carries _meta.related-task
mcpc tasks-result <taskId> --json \
  | jq '.result._meta["io.modelcontextprotocol/related-task"]'
# expect: { "taskId": "<taskId>" }
```

(If mcpc lacks a `--capture-notifications` flag, drive the streamable-HTTP transport directly via the existing integration harness — the test scripts in `tests/integration/suite.ts` will be the source of truth and the PR can document the equivalent path.)

Probes hitting Apify need an `APIFY_TOKEN` provided by the developer.

## Independence

Wave 1 — fully independent. Lands against current master. No dependency on #813 / PR #814 (storage tools), #815 (callOptions allowlist), the cancel-fix landed in PR #812, or the upcoming E/F shape work. Builds on the existing `progressTracker` + `onStatusMessage` machinery.

## Estimate

~3–4 h.

jirispilka:
## What
Wire `notifications/tasks/status` push notifications for task-mode tool calls, and add `_meta.related-task` on `tasks/result` so clients can correlate results to their tasks.

## Non-obvious

- **`server.notification()`
- **dev_server** GET `/` previously returned 405. Now forwarded to `transport.handleRequest`.
- **`notifications/tasks/status` deliberately omits `_meta.related-task`.** Per spec, `taskId` belongs in `params`, not `_meta`. `tasks/result` *does* carry the key — that asymmetry is intentional.
- **`setImmediate(async () => …)` → `setImmediate(() => …).catch(log)`.** The prior form swallowed unhandled rejections from `executeToolAndUpdateTask`.

## Test plan
- [x] `npm run type-check`, `npm run lint`, `npm run test:unit`
- [x] Integration: `should be able to call a long running task tool call` now asserts at least one `taskStatus` event reaches the client — catches a regression where notifications get dropped on streamable HTTP.
- [x] Tested with mcpc and MCPInspector

Closes: #816

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -49,7 +49,9 @@ import {
     FAILURE_CATEGORY,
     HelperTools,
     HTTP_PAYMENT_REQUIRED,
+    RELATED_TASK_META_KEY,
     SERVER_NAME,
+    TASK_STATUS_HEARTBEAT_INTERVAL_MS,
     TOOL_STATUS,
 } from '../const.js';
 import { prepareToolCallContext } from '../payments/helpers.js';
@@ -116,6 +118,34 @@ function isUiSupportedByClient(request: InitializeRequest | undefined): boolean 
 }
 
 type ToolsChangedHandler = (toolNames: string[]) => void;
+
+/** Send notifications/tasks/status for taskId. Swallows errors — notifications are advisory. */
+export async function emitTaskStatusNotification(
+    taskId: string,
+    mcpSessionId: string | undefined,
+    taskStore: TaskStore,
+    sendNotification: RequestHandlerExtra<Request, Notification>['sendNotification'],
+): Promise<void> {
+    try {
+        const task = await taskStore.getTask(taskId, mcpSessionId);
+        if (!task) return;
+        // Per spec: notifications/tasks/status MUST NOT carry _meta.related-task (task ID is in params)
+        await sendNotification({
+            method: 'notifications/tasks/status',
+            params: {
+                taskId: task.taskId,
+                status: task.status,
+                createdAt: task.createdAt,
+                lastUpdatedAt: task.lastUpdatedAt,
+                ttl: task.ttl,
+                ...(task.statusMessage != null && { statusMessage: task.statusMessage }),
+                ...(task.pollInterval != null && { pollInterval: task.pollInterval }),
+            },
+        } as Notification);
+    } catch {
+        // Silent fail — notifications are advisory
+    }
+}
 
 /**
  * Create Apify MCP server
@@ -700,11 +730,19 @@ export class ActorsMcpServer {
                     `Task "${taskId}" is not completed yet. Current status: ${task.status}`,
                 );
             }
-            return await this.taskStore.getTaskResult(taskId, mcpSessionId);
+            const result = await this.taskStore.getTaskResult(taskId, mcpSessionId);
+            // taskId is not in the result body — _meta.related-task lets clients correlate them
+            return {
+                ...result,
+                _meta: {
+                    ...(result._meta as Record<string, unknown> ?? {}),
+                    [RELATED_TASK_META_KEY]: { taskId },
+                },
+            };
         });
 
         // Cancel task
-        this.server.setRequestHandler(CancelTaskRequestSchema, async (request) => {
+        this.server.setRequestHandler(CancelTaskRequestSchema, async (request, extra) => {
             // mcpSessionId is injected at transport layer for session isolation in task stores
             const params = (request.params || {}) as ApifyRequestParams & { taskId: string };
             const { taskId } = params;
@@ -736,6 +774,7 @@ export class ActorsMcpServer {
             await this.taskStore.updateTaskStatus(taskId, 'cancelled', 'Cancelled by client', mcpSessionId);
             const updatedTask = await this.taskStore.getTask(taskId, mcpSessionId);
             log.debug('[CancelTaskRequestSchema] Task cancelled successfully', { taskId, mcpSessionId });
+            await emitTaskStatusNotification(taskId, mcpSessionId, this.taskStore, extra.sendNotification);
             return updatedTask!;
         });
     }
@@ -1263,6 +1302,7 @@ export class ActorsMcpServer {
         // Always populate actor fields so they're tracked on both success and failure paths.
         let callDiagnostics: CallDiagnostics = { ...buildActorFields(actorName, actorId) };
         const startTime = Date.now();
+        let heartbeat: ReturnType<typeof setInterval> | undefined;
 
         log.debug('[executeToolAndUpdateTask] Starting task execution', {
             taskId,
@@ -1308,6 +1348,8 @@ export class ActorsMcpServer {
                 mcpSessionId,
             });
             await this.taskStore.updateTaskStatus(taskId, 'working', undefined, mcpSessionId);
+            let lastEmissionAt = Date.now();
+            await emitTaskStatusNotification(taskId, mcpSessionId, this.taskStore, extra.sendNotification);
 
             // Execute the tool and get the result
             let result: Record<string, unknown> = {};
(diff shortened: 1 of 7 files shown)


src/mcp/server.ts
```diff
@@ -117,6 +124,37 @@ function isUiSupportedByClient(request: InitializeRequest | undefined): boolean
...
+                status: task.status,
+                createdAt: task.createdAt,
+                lastUpdatedAt: task.lastUpdatedAt,
+                ttl: task.ttl,
+                ...(task.statusMessage != null && { statusMessage: task.statusMessage }),
+                ...(task.pollInterval != null && { pollInterval: task.pollInterval }),
+            },
+        } as TaskStatusNotification);
+    } catch {
+        // Silent fail — notifications are advisory
```
MQ37, line 154: nit: maybe add some warning so we can see it is failing?

MQ37 (approved):

src/mcp/server.ts
jirispilka: I had a warning there before, which would spam our logs.
What about to add info log?

PR #830, opened 2026-05-12
refactor: replace ALL_INTERNAL_TOOL_NAMES_CACHE lazy init with eager const (#769)

### LINKED ISSUE #769: refactor: Replace ALL_INTERNAL_TOOL_NAMES_CACHE lazy-init with eager top-level const
The module-level \`let ALL_INTERNAL_TOOL_NAMES_CACHE: Set<string> | null = null\`
pattern in \`src/utils/tools_loader.ts\` uses lazy initialization unnecessarily.

All inputs (\`getCategoryTools\`, \`CATEGORY_NAMES\`, \`WIDGET_BY_BASE_TOOL\`) are
module-level constants available at import time. The computation is cheap and
always runs on the first request anyway.

Replace with an eager top-level \`const\` built via an IIFE:

```ts
const ALL_INTERNAL_TOOL_NAMES: Set<string> = (() => {
    const names = new Set<string>();
    for (const mode of SERVER_MODES) {
        const categories = getCategoryTools(mode);
        for (const name of CATEGORY_NAMES) {
            for (const tool of categories[name]) names.add(tool.name);
        }
    }
    for (const widget of WIDGET_BY_BASE_TOOL.values()) names.add(widget.name);
    return names;
})();
```

Benefits:
- Removes mutable global state (\`let\` → \`const\`)
- Removes the null-check indirection
- Makes the invariant explicit: the set is stable for the process lifetime

mvanhorn:
## Summary

Removes the lazy-init wrapper around `ALL_INTERNAL_TOOL_NAMES_CACHE` in `src/utils/tools_loader.ts` and rebuilds the set eagerly as a `const` at module load. Per the issue body, all inputs (`SERVER_MODES`, `getCategoryTools`, `CATEGORY_NAMES`, `WIDGET_BY_BASE_TOOL`) are module-level constants available at import time, so deferring the computation buys nothing.

## Changes

- `tools_loader.ts:26-46`: replace `let ALL_INTERNAL_TOOL_NAMES_CACHE: Set<string> | null` plus the `getAllInternalToolNames()` helper with an eager `const ALL_INTERNAL_TOOL_NAMES: Set<string>` built via IIFE. Body matches the snippet in #769 verbatim.
- Four callers (`tools_loader.ts:104, 146, 211` and the JSDoc reference on line 90) updated to use the new name directly.

The behavior is identical to the previous code on first call. The only observable difference is that the set is now built at import time rather than the first request, removing the null-check indirection and the mutable module-level `let`.

## Testing

- `npm run type-check` — clean
- `npm run lint` — clean
- `npm run test:unit` — 617 passed, 4 skipped (44 files)

Fixes #769

### DIFF
--- a/src/utils/tools_loader.ts
+++ b/src/utils/tools_loader.ts
@@ -23,27 +23,21 @@ import type { ActorStore, Input, ToolCategory, ToolEntry } from '../types.js';
 import { SERVER_MODES, ServerMode } from '../types.js';
 
 // All internal tool names across all modes. Selectors matching these are not treated as Actor IDs.
-let ALL_INTERNAL_TOOL_NAMES_CACHE: Set<string> | null = null;
-function getAllInternalToolNames(): Set<string> {
-    if (!ALL_INTERNAL_TOOL_NAMES_CACHE) {
-        const allNames = new Set<string>();
-        // Collect tool names from both modes to ensure complete classification
-        for (const mode of SERVER_MODES) {
-            const categories = getCategoryTools(mode);
-            for (const name of CATEGORY_NAMES) {
-                for (const tool of categories[name]) {
-                    allNames.add(tool.name);
-                }
-            }
+// Built eagerly at module load; inputs (SERVER_MODES, getCategoryTools, CATEGORY_NAMES,
+// WIDGET_BY_BASE_TOOL) are module-level constants available at import time.
+const ALL_INTERNAL_TOOL_NAMES: Set<string> = (() => {
+    const names = new Set<string>();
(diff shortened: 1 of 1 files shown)


MQ37 (approved):
Checked the code and it seems the circular dependency is not there anymore and we can safely refactor.

Executed all tests in this and the internal repo with current PR build and everything is passing ✔️

jirispilka:
> Checked the code and it seems the circular dependency is not there anymore and we can safely refactor.
> 
> Executed all tests in this and the internal repo with current PR build and everything is passing ✔️

Then let's go! Thanks!

PR #831, opened 2026-05-12
ci(integration-tests): run integration tests on PR and master push

MQ37:
## Context
We have `npm run test:integration` but nothing runs it in CI. This wires it into PR and master-push workflows.

## Solution
New reusable workflow `_integration_tests.yaml` called from `on_pull_request.yaml` and `on_master.yaml`. Reads `APIFY_TOKEN` from the repo secret of the same name.

## Worth your attention
- **Fork PRs are skipped** — `if: github.event.pull_request.head.repo.full_name == github.repository`. Secrets aren't available to forks, so we skip rather than fail.
- **Secret handling** — declared explicitly on the reusable workflow (no `secrets: inherit`), scoped as env to the test step only, and re-masked via `::add-mask::` to cover transformed-leak edge cases (base64/url-encoded).
- **Not a required check** — added as a normal job; promote it in branch protection if/when you want it blocking.

## Follow-up
- Requires a repo (or org) secret named `APIFY_TEST_USER_API_TOKEN` to be created in `apify/apify-mcp-server` settings before the job will pass. :heavy_check_mark:

### DIFF
--- a/.github/workflows/_integration_tests.yaml
+++ b/.github/workflows/_integration_tests.yaml
@@ -1,0 +1,33 @@
+name: Integration tests
+
+on:
+    workflow_call:
+        secrets:
+            APIFY_TOKEN:
+                required: true
+
+jobs:
+    integration_tests:
+        name: Integration tests
+        runs-on: ubuntu-latest
+
+        steps:
+            -   uses: actions/checkout@v6
+            -   name: Use Node.js
+                uses: actions/setup-node@v6
+                with:
+                    node-version-file: '.nvmrc'
+                    cache: 'npm'
+                    cache-dependency-path: 'package-lock.json'
+            -   name: Install Dependencies
+                run: npm ci --force
+            # Re-mask in case the token shows up transformed (base64/url-encoded/etc.) in
+            # any downstream log output. Literal value is already masked by GitHub.
+            -   name: Mask token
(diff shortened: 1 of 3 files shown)


jirispilka (commented):
This will take a long time, to run integration tests on every pull requests.

The integration test will run on every commit for every PR.
I don't think we want to do that unless the integration test is fast or can we somehow gate it? Run only if ...?

MQ37:
@jirispilka this is why the integration tests CICD workflow is **not required**.

Why would you want to gate it? It is only optional addition, you can wait or merge straight away if you want to - it does not block or restrict PRs in any way.

Let's first try that and if that is annoying we can improve or disable - but every production grade repo should run the whole test suite on PR and merge.

MQ37:
@jirispilka let's just try that, we can always disable this if we find that annoying. I added gating logic to only execute the integration tests on relevant code changes:
```
                            - 'src/**'
                            - 'tests/integration/**'
                            - 'package.json'
                            - 'package-lock.json'
                            - 'tsconfig.json'
                            - '.nvmrc'
                            - '.github/workflows/_integration_tests.yaml'
                            - '.github/workflows/on_pull_request.yaml'
```

MQ37:
@jirispilka merging

PR #836, opened 2026-05-13
docs: add TOON format adoption spec

MQ37:
## Context

Apify MCP tool responses wire JSON text content for most tools. Every Actor run, dataset listing, KV-store key, and dataset item lands in the caller LLM's context and stays there for every subsequent turn — the repeated keys in JSON arrays compound per request.

## Solution

Research spec for an **adaptive picker** at each in-scope tool-call site: encodes both the current JSON and TOON (with a small dot-flatten transform) per response, ships whichever is smaller. `json` is always a candidate, so the picker is never worse than today.

Companion to [Code Mode (#794)](https://github.com/apify/apify-mcp-server/pull/794) — TOON shrinks results that flow through the LLM context, Code Mode skips most of them entirely. The spec proposes exposing the picker's encoder as `apify.stringifyCompact()` for Code Mode programs.

**No production code in this PR.** Spec-only, pending team review.

## Worth your attention

- **Measured savings**: −6.2% bytes combined on 18 real Apify-API fixtures; the four list endpoints save 19–44% individually. `dataset-items` averages near zero — savings concentrate on list endpoints.
- **Stacks with `fields=`** (separate proposal): 100-place Google-Maps result projected to 3 fields → 1.01 MB → 9.5 KB, ~110× smaller into LLM context.
- **`structuredContent` is unchanged**. Only `TextContent.text` shifts. `outputSchema` validation, MCP widgets, and programmatic consumers continue to see the same JSON.
- **Bounded by construction**: `json` always in candidate set; defensive `try`/`catch` around the TOON candidate; `MAX_DEPTH = 20` guard with margin of 11 over deepest real fixture; dotted-key normalisation with collision guard.
- **LLM behaviour change unmeasured**. The spec calls for an `evals/workflows/` regression gate before any production merge.

## Follow-up

- Implementation lands in a separate PR after team approval of the spec.
- Workflow eval regression gate required before merge.

### DIFF
--- a/res/toon_format_adoption.md
+++ b/res/toon_format_adoption.md
@@ -1,0 +1,386 @@
+# TOON format adoption
+
+## Problem
+
+MCP tool results land in the LLM context window. Most tools wire
+JSON-encoded payloads (Actor runs, dataset listings, KV-store keys,
+dataset items). JSON repeats every object key on every row of an
+array. Context costs tokens and dilutes model attention, and once a
+tool result lands it stays in the conversation for every subsequent
+turn — the savings compound per request.
+
+## TL;DR
+
+Switching the text content of certain MCP tool responses from JSON to
+[TOON](https://toonformat.dev/) (a more compact LLM-oriented format,
+introduced below) measurably reduces context bytes — and stacks
+massively with field projection on realistic agentic workloads.
+
+| Tool                              | Δ vs current JSON | Strategy that wins most often  |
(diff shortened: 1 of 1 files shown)


jirispilka (approved):
It looks good, I would transform this to an issue.

It will be easier to update it later, including new experiments and number.
After the implementation is done, we can publish it as an md too.

MQ37:
created issue here, closing https://github.com/apify/apify-mcp-server/issues/913

PR #841, opened 2026-05-13
docs: Tighten e2e testing rules - use mcpc and apify CLI literally

jirispilka:
When the user names a testing tool (mcpc, apify CLI), use it literally: no curl, no ad-hoc scripts, no unit-test substitutes. Bridges are usually already connected — check via `mcpc` first, then `restart` if listed or `connect` if not. Use `apify` CLI for ground-truth IDs and records.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -35,18 +35,9 @@ Zero tolerance for errors — fix before proceeding, don't defer.
 
 ## Testing the MCP server end-to-end
 
-After code changes, verify the server works — not just that it compiles.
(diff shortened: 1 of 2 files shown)


CLAUDE.md
MQ37, line 1: nit: BTW I think we should rename this to `AGENTS.md` and then make symlink to `CLAUDE.md`. I know the Claude was first but AGENTS.md is more general and embraced standard and we should not conform to a standard made by a single company that at first refused to embrace the more general standard.

MQ37 (approved):

CLAUDE.md
jirispilka: Alright, no strong opinion:)

PR #843, opened 2026-05-14
chore: Use custom action to commit changes in CI instead of `git commit`

fnesveda:
We want to enforce commit signing for all commits in our repositories.
To do that, we need to make sure even commits created by CI workflows are signed.

It would be possible to sign using GPG keys, but that would require a lot of maintenance.

Instead, we can commit using the GitHub GraphQL API, which automatically signs commits.

This PR replaces direct `git commit` / `git push` usage (and third-party commit actions like `EndBug/add-and-commit`)
with the `apify/actions/signed-commit` action, which uses the GraphQL API under the hood.

### DIFF
--- a/.github/workflows/_update_release_metadata.yaml
+++ b/.github/workflows/_update_release_metadata.yaml
@@ -56,7 +56,7 @@ jobs:
         name: Update changelog
         runs-on: ubuntu-latest
         outputs:
-            changelog_commitish: ${{ steps.commit.outputs.commit-sha || github.sha }}
+            changelog_commitish: ${{ steps.commit.outputs.commit_sha || github.sha }}
 
         steps:
             -   name: Checkout repository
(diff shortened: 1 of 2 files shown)


MQ37 (approved):

janbuchar (changes requested):
Please use the new options on the commit action 🙂

.github/workflows/_update_release_metadata.yaml
```diff
@@ -56,7 +56,7 @@ jobs:
         name: Update changelog
         runs-on: ubuntu-latest
         outputs:
-            changelog_commitish: ${{ steps.commit.outputs.commit_long_sha || github.sha }}
+            changelog_commitish: ${{ steps.commit.outputs.commit_sha || github.sha }}
```
janbuchar:
```suggestion
            changelog_commitish: ${{ steps.commit.outputs.commit_sha }}
```

janbuchar (approved):

Issue #587, opened 2026-03-22
feat: Post-#582 roadmap — direct actor tools, resource links, dynamic resources, get-actor-output(runId), elicitation, completion


jirispilka:
## Context

This is a tracking issue for follow-up work after #582 (async call-actor + waitSecs on get-actor-run). These features build on the infrastructure introduced in #582 and form a natural progression toward a cleaner, more MCP-native Actor execution experience.

Ideas sourced from #582 design sessions and team meeting notes.

### Phase 2: Direct actor tools → non-blocking uniform response
**Value: High | Effort: Low | Depends on: #582**

Make direct actor tools (RAG web browser, etc.) also non-blocking with the uniform response shape. #582 builds all the infrastructure needed (`buildCallActorStructuredContent`, `waitForRunWithAbort`, `mcpTaskExecution` flag, `actorRunOutputSchema`). Direct actor tools just need to be wired to use it.

- Update `src/tools/default/actor_executor.ts` and `src/tools/openai/actor_executor.ts`
- In task mode: wait for completion. In non-task mode: start and return immediately.
- Remove old code: `callActorGetDataset`, `buildActorResponseContent`, `callActorOutputSchema`, `buildEnrichedCallActorOutputSchema`, `ActorStore` enrichment pipeline

### Phase 3: Enhanced dataset tools — download URLs, schema, resource links
**Value: High | Effort: Medium | Depends on: #582**

Significant upgrade to `get-dataset-items` / `get-actor-output`:

- **Resource links** to download URLs in content array:
  ```typescript
  content: [
      { type: 'resource_link', uri: 'https://api.apify.com/...format=csv', name: '[dataset_ID].csv', mimeType: 'text/csv' },
      { type: 'resource_link', uri: 'https://api.apify.com/...format=json', name: '[dataset_ID].json', mimeType: 'application/json' },
      // csv, json, xml, xlsx
  ]
  ```
- **`downloadUrls`** in structured content (xlsx, csv, json, xml links)
- **`schema`** in structured content (auto-generated from items)
- **Default `limit: 10`** instead of 100 (avoid wasting LLM context)
- **Flexible `datasetUrlOrId`** input — accept dataset ID, URL, or `username~dataset-name` format

### Phase 4: `get-key-value-store-files` — new tool
**Value: Medium | Effort: Medium | Depends on: #582**

New tool for browsing KV store records as resource links. Every Actor run produces a KV store (logs, screenshots, etc.) that's currently invisible.

[...]

PR #853, opened 2026-05-16
feat: Migrate direct Actor tools to RunResponse with waitSec

### LINKED ISSUE #852: feat: migrate direct actor tools to canonical RunResponse shape
## Problem

Direct actor tools (e.g. `apify--rag-web-browser`) still use the old execution path (`callActorGetDataset` → `buildActorResponseContent`) which returns items inline with the pre-v4 `{ runId, datasetId, items, instructions }` shape. PR #825 (branch `feat/async-call-actor`) ships the canonical `RunResponse` for `call-actor` and `get-actor-run` but explicitly defers this migration. The result: two different response shapes depending on which tool the LLM uses.

## Proposed solution

Replace `defaultActorExecutor` to call `actor.start()` then `fetchActorRunData()` — the same path `call-actor` already uses — and return `buildGetActorRunSuccessResponse()`. Items are no longer inlined; the LLM follows the `nextStep` hint to `get-dataset-items`.

Internal repo: update any tests/code reading `items` / `instructions` / top-level `datasetId` from direct actor tool responses.

**Blocked by:** PR #825

---

⚠️ **Schema gotcha**: `outputSchema.storages.datasets.default.itemsSchema.properties` declares per-Actor row shape (e.g. `{ url: { type: 'string' } }`). LLMs may read this as "`url` is returned in `structuredContent`" — it isn't. Rows live in the referenced dataset; fetch with `get-dataset-items`. Mitigated by `itemsSchema.description`. Fallback if LLMs misread: move the hint to tool `description` / `_meta`.

jirispilka:
Follow-up from #825 Direct Actor tools now share the same contract as `call-actor` and `get-actor-run`. Finally one uniform experience, this is what I was looking forward to when writing #825 :)

Much smaller PR than #825 this time, just the migration and merging two executors into one. It actually removes some code, which is good. 

About the `itemsSchema` gotcha called out in #852 — LLMs may read \`itemsSchema.properties\` as "these fields are in \`structuredContent\`", which they are not. Rows live in the dataset and you fetch them with \`get-dataset-items\`. The \`itemsSchema.description\` says this explicitly, so it should be fine. 

I tested it, it works but I did not tested thoroughly, there might be some rough edges.

## Summary

- **Direct Actor tools now return the canonical \`RunResponse\` shape** — same as \`call-actor\` and \`get-actor-run\`: \`runId\`, \`status\`, \`storages\`, \`summary\`, \`nextStep\`. No more inlined items; the LLM follows \`nextStep\` to \`get-dataset-items\`.

- **Same \`waitSecs\` contract as \`call-actor\`** — same MCP-only opt-in: default 30 s, max 45 s. Task mode overrides \`waitSecs\` and waits until terminal. If the Actor's own input schema happens to declare \`waitSecs\`, ours wins so there's no collision.

- **Per-tool \`itemsSchema\` in the dataset metadata** — for direct Actor tools the target Actor is known at \`tools/list\` time, so the historical-inferred row schema lands both in the tool's \`outputSchema\` and in the runtime response under \`storages.datasets.default.itemsSchema\`. LLMs can plan a \`fields\` projection before the first call.

- **One mode-agnostic executor** — separate default-mode and apps-mode executors are gone, replaced by a single one. Direct Actor tools never render widgets (\`widget: false\`), so the apps-mode executor was not doing anything different anymore.

Closes #852. Phase 2 of #587.


jirispilka:
E2E tested via `mcpc` + Apify v2 API ground-truth. All groups pass — shippable ✅

- **Shape parity**: `call-actor`, `apify--rag-web-browser`, `get-actor-run` return identical `RunResponse` keys (`runId`, `actorId`, `status`, `storages`, `summary`, `nextStep`).
- **Direct actor tool**: no inline items, `nextStep` interpolates real `datasetId` + `limit=20`, MCP-spec `content[0]` mirrors `structuredContent`.
- **`waitSecs`**: cap at 45 / min 0 enforced via AJV (`-32602`); `waitSecs:0` returns RUNNING with `nextStep → get-actor-run`; **stripped from Actor input** (verified by reading the run's stored `INPUT` KV record).
- **Task mode**: ignores `waitSecs` and waits until terminal; `tasks-cancel` mid-run propagates — confirmed underlying Apify run reached `ABORTED` via API.
- **No widget `_meta.ui`** on direct actor tool list/response.
- **`itemsSchema` enrichment**: not reachable from stdio (no `actorStore`); unit tests cover it.

Session: https://claude.ai/code/session_0152Dv5wbydZEnhvHbPeTAbz

---

src/tools/structured_output_schemas.ts
```diff
@@ -394,6 +374,58 @@ export const getActorRunOutputSchema = {
...
+                                    itemsSchema,
+                                },
+                            },
+                        },
+                    },
+                },
+            },
+        },
+    };
+}
```
MQ37, line 427: refactor: oh my god! 🤯 we need to refactor this - what about clone and setting the fields directly?

MQ37 (approved):
Good job! I was playing with it and it works really well, it makes the MCP server experience much better.

there is only one refactor change, but let's make a separate issue where we address all the issue with this PR chain, so we can prevent any further merge conflicts.

While playing with it I discovered an issue when trying to call Actor MCP server, that was supposedly there from February essentially preventing any real client from using the Actor MCP servers through the `call-actor` tool 🤯 🤦  Fix stacked on top of this PR https://github.com/apify/apify-mcp-server/pull/859

src/tools/structured_output_schemas.ts
jirispilka: yeah, this is terrible

PR #854, opened 2026-05-16
fix: Merge dataset item schemas across heterogeneous shapes

MQ37:
## Context

`get-dataset-schema`, `call-actor`, and `get-actor-run` all share `generateSchemaFromItems`, which delegated to `to-json-schema@0.2.5`. Any two items with different key sets collapsed to `{type:'array',items:{type:'object'}}` — properties wiped out. Reported on a real NYC restaurants dataset where ~half the items carried `markdown` and half didn't.

## Solution

Replaced the library with an in-house inferrer in `src/utils/schema_generation.ts`. The merge does a set-union of property keys and recurses; primitive type conflicts emit JSON Schema `type` arrays (e.g. `["string","null"]`). Drops the `arrayMode` field from `get-dataset-schema` — it only existed as a workaround for the buggy `mode:'all'`, and all internal callers were already passing it anyway.

## Worth your attention

- **No external dependency, no supply-chain surface.** `to-json-schema` was last published in 2020 and the upstream repo is dead. Owning ~120 LOC of pure JSON-Schema inference is cheaper than auditing an unmaintained transitive surface on a server that handles customer Apify tokens.
- **Type-array unions for primitive conflicts.** `{x:1}` + `{x:"hi"}` produces `{"type":["integer","string"]}` — spec-valid JSON Schema, handled natively by LLMs reading the tool output. Verified the generated schema is never Ajv-validated downstream (checked both this repo and `apify-mcp-server-internal` — Ajv only validates tool *input* args).
- **`arrayMode` field removed from `get-dataset-schema`.** Technically a public API change. Safe because (a) all 3 internal callers always passed `arrayMode:'all'`, and (b) the `'first'` mode was never useful — `to-json-schema` applies it recursively to nested arrays too, which is almost never what callers want.
- **Drops the upstream's `format:"style"` false positive.** Free-form Markdown text was being tagged with a CSS-ish format. The new format detector covers only `uri`, `date-time`, `date`, `email`, `uuid` — the unambiguous ones.

## Follow-up

- **#855 stacks a compact rewrite on top of this PR — same behavior, 387 → 148 combined code LOC.** Merge order: this PR first, then #855. Or squash-merge #855 alone to replace this. This PR ships the verbose, easy-to-read version for review clarity.


src/utils/schema_generation.ts
```diff
@@ -1,107 +1,158 @@
...
+
+function detectFormat(value: string): string | undefined {
+    for (const [name, test] of FORMAT_DETECTORS) {
+        if (test(value)) return name;
+    }
+    return undefined;
+}
+
+function inferType(value: unknown): JsonSchemaPrimitiveType {
+    if (value === null) return 'null';
```
jirispilka:
```suggestion
    if (value === null || value === undefined) return 'null';
```

inferType falls through to return 'object' for undefined (no typeof value === 'undefined' guard). Then inferSchema calls Object.entries(undefined) which throws TypeError. JSON doesn't have
   undefined so dataset items won't hit this, but it's a latent trap that'll show up in tests if someone ever passes a sparse array.

jirispilka (approved):
One nit, otherwise it looks good.
I'll test it in the follow up PR

PR #855, opened 2026-05-16
refactor: Compact schema_generation to <150 LOC

MQ37:
## Context

Reviewer ceiling for `src/utils/schema_generation.ts` + tests is **150 LOC of code** (comments / blank lines free). The fix in #854 lands at 387.

## Solution

Compact both files to **148 combined LOC** with the same behavior. Stacked on top of #854 so the diff in this PR is purely the compression — no behavior change to review here.

| File | #854 | This PR |
|---|---|---|
| `src/utils/schema_generation.ts` | 122 | 69 |
| `tests/unit/schema_generation.test.ts` | 265 | 79 |
| **Combined** | **387** | **148** |

## Worth your attention

- **Coverage equivalent, fewer test blocks.** 28 → 21 tests. The 6 type-union cases and 5 format-detection cases are folded into two `it.each` tables. Same regression coverage: NYC sushi case, heterogeneous keys, format false-positive guard.
- **Two `!` non-null assertions added in `merge`.** Both inside branches guarded by `&&`/`||` predicates immediately above (`ap[k] && bp[k]` and `a.items || b.items`). The narrowing alternative was 4 LOC longer.
- **Internal types dropped, public surface unchanged.** `JsonSchemaArray`, `JsonSchemaObject`, `SchemaGenerationOptions`, and the `removeEmptyArrays` export were all unused outside the module. `JsonSchemaProperty` stays (imported by `actor_execution.ts`).

## Open

- If you prefer #854's verbose version, close this PR. If you prefer this one, squash-merge it as a replacement for #854 (or merge #854 first then this on top).

### DIFF
--- a/src/utils/schema_generation.ts
+++ b/src/utils/schema_generation.ts
@@ -9,150 +9,70 @@ export type JsonSchemaProperty = {
     format?: string;
 };
 
-export type JsonSchemaArray = {
-    type: 'array';
-    items: JsonSchemaProperty;
-};
-
-export type SchemaGenerationOptions = {
-    /** Maximum number of items to use for schema generation. Default is 5. */
-    limit?: number;
-    /** If true, strips empty arrays from items before inference. Default is true. */
-    clean?: boolean;
-};
-
-// Local counterpart to the dataset API's `clean=true` — empty arrays carry no schema info.
-export function removeEmptyArrays(obj: unknown): unknown {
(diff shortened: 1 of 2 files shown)


jirispilka (changes requested):
As you might have guessed already, I prefer the human readable version :D

MQ37:
@jirispilka ahh, these humans - closing then :smile:

PR #863, opened 2026-05-18
fix: Stabilise flaky rag-web-browser integration tests

MQ37:
## Context

Two unrelated rag-web-browser integration tests flake in CI:

1. `calls apify/rag-web-browser tool directly and retrieves metadata.title via get-dataset-items` — `items.length > 0` asserts before the run populates the dataset under the implicit 30s `waitSecs` default.
2. The three run-abort tests (notifications/cancelled × 2, tasks/cancel × 1) verify the abort side-effect via `apiClient.runs().list({ limit: 5 })`, which lists runs across **all** Actors. Page 1 is dominated by other tests' Actor starts; the abort-test's run is rarely in the top 5, so the wait silently times out. Empirical probe against the CI token returned 0 of 5 rag-web-browser entries — only 3 distinct Actors in the top 5.

## Solution

1. Set explicit `waitSecs: 45` (= `WAIT_SECS_MAX`) on the direct-call test — matches the other rag-web-browser tests in the file.
2. Replace `waitForActorRunAbortStatus` with `captureInflightActorRunId` + `waitForRunAborted`. The runId isn't reachable through the MCP client (no response after cancel, no runId in `notifications/progress`), so we race the Apify API for this Actor's newest non-terminal run while the call is in flight, then trigger the cancel. Verification becomes an exact-runId lookup.

## Worth your attention

- **Actor-scoped runs list** — `apiClient.actor(actorId).runs().list()` instead of the global `runs()` collection eliminates cross-actor pollution (the empirical root cause).
- **Non-terminal status filter** — excludes prior completed runs that could match the time window on retry.
- **Cancel mechanism untouched** — `notifications/cancelled` / `tasks/cancel` and server-side `onAbort: abortRunOnSignal` are unchanged; only the verification side moves from list heuristic to exact-ID lookup.

https://claude.ai/code/session_01VraSf24soKanDRbtptDJZ6

### DIFF
--- a/tests/integration/suite.ts
+++ b/tests/integration/suite.ts
@@ -24,7 +24,7 @@ import { getExpectedToolNamesByCategories } from '../../src/utils/tool_categorie
 import { AUTO_INJECTED_TOOLS } from '../../src/utils/tools_loader.js';
 import { ACTOR_MCP_SERVER_ACTOR_NAME, ACTOR_PYTHON_EXAMPLE, DEFAULT_ACTOR_NAMES, getDefaultToolNames } from '../const.js';
 import { addActor, type McpClientOptions } from '../helpers.js';
-import { assertStatusMessagePropagated, waitForActorRunAbortStatus } from './utils/task_waits.js';
+import { assertStatusMessagePropagated, captureInflightActorRunId, waitForRunAborted } from './utils/task_waits.js';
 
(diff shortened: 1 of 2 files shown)


MQ37:
YOLO @jirispilka @RobertCrupa @vojtechj-apify

PR #864, opened 2026-05-18
feat: Track tool response size in telemetry

### LINKED ISSUE #838: feat(telemetry): track tool result size in bytes
## Problem
`ToolCallTelemetryProperties` has no signal for how heavy tool results are. Bytes is a simple, language-agnostic measure we can add cheaply.

`src/types.ts:335` — current shape has no size field:

```ts
export type ToolCallTelemetryProperties = {
    // ...
    tool_status: ToolStatus;
    tool_exec_time_ms: number;
    // no result_size_bytes
};
```

Result is built via `buildMCPResponse` (`src/utils/mcp.ts`) with `texts` and optional `structuredContent` — both contribute to the payload sent to the client.

## Fix
Add `result_size_bytes: number` to `ToolCallTelemetryProperties`. Compute as UTF-8 byte length of joined `content[].text` plus `JSON.stringify(structuredContent)` when present. Capture the result in the `CallToolRequestSchema` handler and pass it into `logToolCallAndTelemetry` (`src/mcp/server.ts:1198`) so it's emitted on the Segment `MCP Tool Call` event alongside `tool_exec_time_ms`.

## Out of scope
Token counting — requires per-model tokenizer, adds cost for marginal accuracy gain over bytes.

MQ37:
## Context
We want to track tool-response payload size in Segment telemetry to spot outliers, estimate egress / storage cost, and correlate response size with tool status and latency. Closes #838.

## Solution
Add a `tool_response_size_bytes` field to `ToolCallTelemetryProperties` and compute UTF-8 byte length of the response (`content[]` text items + JSON-stringified `structuredContent`) in the call-tool `finally` block. A `captureResult()` closure stashes the return value so the `finally` can read it after the handler has already returned. The task-mode async executor passes the size in directly via `finishTaskTracking`.

## Worth your attention
- **Bytes, not characters** — `bytes / 4` is the standard token-count heuristic ([LangWatch `estimateTokens`](https://langwatch.ai/scenario/reference/javascript/scenario/functions/estimateTokens.html) uses it, justified as: *"accounts for multi-byte characters (emojis, CJK, etc.) that typically consume more tokens than ASCII text"*). Pure character heuristics are documented to break on CJK by 3-4× and on emoji by 4-5× ([hive#3546](https://github.com/aden-hq/hive/issues/3546), [openviking 6471868](https://github.com/LinQiang391/OpenViking/commit/6471868c90e66e2e783355808e13acec188a23a6)) because each CJK char is ~1-2 tokens. JS `String.length` also counts UTF-16 code units, so it under-counts astral-plane chars by 2× — bytes sidestep that entirely. Cross-language *tokens-per-unit* spread is ~2× for bytes vs ~8× for chars.
- **`captureResult()` is declared inside the handler scope** — each concurrent request gets its own closure, so the per-request `toolResult` cannot race. Must not be hoisted to a class field.
- **Field is optional** (`responseSizeBytes?: number` + conditional spreads) — telemetry payload shape stays unchanged for the few paths that throw before reaching any `return` (e.g. early `failInvalidParams`).

## Follow-up
- If dashboards show we actually need token counts (not transport bytes), swap `Buffer.byteLength(text, 'utf8')` for `encoder.encode(text).length` from [`gpt-tokenizer`](https://www.npmjs.com/package/gpt-tokenizer) at the same call sites — accuracy goes from ~2× heuristic to ~95% on cl100k_base.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -83,7 +83,7 @@ import type {
 } from '../types.js';
 import { ServerMode } from '../types.js';
 import { getHttpStatusCode, logHttpError } from '../utils/logging.js';
-import { buildMCPResponse, getToolCallErrorUserText } from '../utils/mcp.js';
+import { buildMCPResponse, computeToolResponseSizeBytes, getToolCallErrorUserText } from '../utils/mcp.js';
 import { buildPaymentRequiredResponse } from '../utils/payment_errors.js';
 import { createProgressTracker } from '../utils/progress.js';
 import { parseServerMode, resolveServerMode } from '../utils/server_mode.js';
@@ -826,6 +826,12 @@ export class ActorsMcpServer {
             let callDiagnostics: CallDiagnostics = {};
             let shouldTrackTelemetry = true;
             let resolvedToolName = name;
+            // Captured by `captureResult` so the `finally` block can measure response size for telemetry.
+            let toolResult: unknown = null;
+            const captureResult = <T>(r: T): T => {
+                toolResult = r;
+                return r;
+            };
             const failInvalidParams = async (
                 message: string,
                 details: CallDiagnostics,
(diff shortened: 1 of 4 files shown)


MQ37:
merging @jirispilka

PR #865, opened 2026-05-18
feat: Add tool call reason middleware for telemetry

MQ37:
Inject a `reason` property into every tool's input schema so LLMs can explain why they're calling each tool. Extract and forward the reason to telemetry, then strip it before validation and execution so downstream code never observes it.

## Changes

- **New module `src/utils/tool_call_reason.ts`**: Exports `injectReasonProperty()` to add the `reason` field to tool schemas, and `extractAndStripReason()` to extract and remove it from call arguments. Includes strict description guidance to prevent LLMs from leaking conversation context, PII, or secrets.
- **`src/utils/tools.ts`**: Call `injectReasonProperty()` in `getToolPublicFieldOnly()` so every tool listed in `tools/list` includes the reason field.
- **`src/mcp/server.ts`**: Extract reason from tool arguments before payment processing and AJV validation, then forward it to telemetry data.
- **`src/types.ts`**: Add optional `reason` field to `ToolCallTelemetryProperties`.
- **Tests**: Full unit test coverage for injection, extraction, idempotency, edge cases, and integration with `getToolPublicFieldOnly()`. Updated existing tests to expect `reason` in required fields.

## Implementation details

- Injection is idempotent: applying it twice returns the same schema instance.
- Extraction always strips the field from args (even if invalid), preventing it from reaching validation or tool execution.
- Reason is capped at 500 characters before forwarding to telemetry.
- The description is explicit about scope (immediate use case only) and prohibits leaking context, identifiers, or secrets.

https://claude.ai/code/session_01FeiC1otmZnAPxSAVxzRidk

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -89,6 +89,7 @@ import { createProgressTracker } from '../utils/progress.js';
 import { parseServerMode, resolveServerMode } from '../utils/server_mode.js';
 import { getServerInstructions } from '../utils/server-instructions/index.js';
 import { classifyFailureCategory, extractAjvErrorDetails, extractToolTelemetry, getToolStatusFromError } from '../utils/tool_status.js';
+import { extractAndStripReason } from '../utils/tool_call_reason.js';
 import { buildActorFields, extractActorId, extractActorName, getToolFullName, getToolPublicFieldOnly } from '../utils/tools.js';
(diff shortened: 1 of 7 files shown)


MQ37:
closing this PR as we will create a new and cleaner implementation

PR #868, opened 2026-05-19
fix: Unify search-actors keywords description across variants

### LINKED ISSUE #833: Change search-actors description, when users search for top or best Actors the query should be empty
The search should be performed without keywords, otherwise it will return results not matching user intent

Let's not over-engineer it.
- Try to tighten the search-actors tool description with a hint how Actors are sorted by default
- Play with it
- If ok, release it, if we are not able to fix it, let's close it

Image

vojtechj-apify:
## Summary

Root-caused #833. The `keywords` field description was split across two schemas in `src/tools/core/search_actors_common.ts`:

- `searchActorsBaseArgsSchema.keywords` — one sentence (*"Keywords used to search for Actors in the Apify Store."*)
- `searchActorsArgsSchema.keywords` — 17 lines of rich rules + examples

`search-actors-widget` (`src/tools/apps/search_actors_widget.ts:24`) re-uses the **base** schema via `.strict()`, so the widget tool was getting the one-sentence description. The rich guidance only reached `search-actors` (non-widget, default mode). Split introduced in #711 → #723 and locked by `tests/unit/tools.mode_contract.test.ts:146`.

Claude Desktop and ChatGPT advertise the MCP-Apps UI capability → server resolves to apps mode → bad calls hit `search-actors-widget`, the tool with zero keyword guidance.

## Change

- Move the rich `keywords` description down to `searchActorsBaseArgsSchema`.
- Remove the now-redundant `keywords` override on `searchActorsArgsSchema` (becomes an empty `extend({})`).
- Both variants now inherit the same guidance from one place.
- Add explicit empty-string rule (no specific platform/data type → empty) + new bad examples covering ranking words and bare task words.
- No schema-shape change; the locking unit test passes by construction.

## Live test results (Claude Desktop, stdio)

Iterated the description text through three versions, monitoring `~/Library/Logs/Claude/mcp-server-apify-local.log` between rounds.

| Prompt class | `keywords` value | Verdict |
|---|---|---|
| *"top/most popular/best Actors"* | `""` | ✅ |
| *"find me Instagram scrapers"* | `"Instagram"` | ✅ |
| *"what can scrape Amazon products?"* | `"Amazon products"` | ✅ |
| *"list 10 scrapers"*, *"what extractors are available?"* | `""` | ✅ |
| *"find me web crawlers"* | `"web crawler"` | ✅ (real Actors match — not noise) |

No regressions on the specific-platform path.

## Test plan
- [x] `npm run type-check` — clean
- [x] `npm run lint` — clean
- [x] `npm run test:unit` — 659 passed (1 pre-existing skip)
- [x] Manual: live Claude Desktop session with monitored stdio log
- [ ] Optional: run eval workflow (apply `validated` label) to confirm no regression on the existing 15 `search-actors-*` cases

Closes #833

### DIFF
--- a/src/tools/core/search_actors_common.ts
+++ b/src/tools/core/search_actors_common.ts
@@ -10,8 +10,8 @@ import type { PricingTier } from '../../utils/pricing_info.js';
 import { actorSearchOutputSchema } from '../structured_output_schemas.js';
 
 /**
- * Shared base schema for search-actors arguments. Used directly by the widget
- * variant; extended by `searchActorsArgsSchema` with a longer `keywords` description.
+ * Shared schema for search-actors arguments. Used by both the default and
+ * widget variants — the widget variant calls `.strict()` on it.
  */
 export const searchActorsBaseArgsSchema = z.object({
     keywords: z.string()
@@ -54,14 +54,6 @@ export const searchActorsBaseArgsSchema = z.object({
         .default(0)
         .describe('The number of elements to skip from the start (default = 0)'),
 });
-
-/**
- * Zod schema for the search-actors tool arguments. Equivalent to the base
- * schema; the widget variant uses the same base schema via `.strict()` — both
- * variants share the `keywords` description that lives on
- * `searchActorsBaseArgsSchema`.
- */
-export const searchActorsArgsSchema = searchActorsBaseArgsSchema.extend({});
 
 const SEARCH_ACTORS_DESCRIPTION = `
 Search the Apify Store to FIND and DISCOVER what scraping tools/Actors exist for specific platforms or use cases.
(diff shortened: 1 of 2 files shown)


jirispilka (approved):
Thanks!

src/tools/core/search_actors_common.ts
```diff
@@ -31,30 +56,12 @@ export const searchActorsBaseArgsSchema = z.object({
...
-            - If a user asks about "fetching Instagram posts", use "Instagram posts" as keywords
-            - The goal is to find Actors that specifically handle the platform and data type the user mentioned
-
-            Examples:
-            ✅ Good: "Instagram posts", "Twitter", "Amazon products", "weather", "news articles"
-            ❌ Bad: "Instagram posts profiles comments hashtags reels stories followers..." (too long, too many terms)
-            ❌ Bad: "data extraction scraping tools" (too generic)
-        `),
-});
+export const searchActorsArgsSchema = searchActorsBaseArgsSchema.extend({});
```
jirispilka: This line is no longer needed, let's use `searchActorsBaseArgsSchema` directly

jirispilka (approved):
It looks nice!

Although the description is a bit long.

There is one code smell that needs to be fixed: pre-approving

src/tools/core/search_actors_common.ts
vojtechj-apify: I have dropped the redundat searchActorsArgsSchema alias, so we are using searchActorsBaseArgsSchema across the project

jirispilka (approved):

MQ37 (approved):
The description is quite long, but if it works let's go 😎 I did not test it though

PR #869, opened 2026-05-19
chore: migrate from npm to pnpm 11

B4nan:
### Lockfile / package manager
- Drop `package-lock.json` (root + `src/web/`), add `pnpm-lock.yaml`.
- `packageManager: pnpm@11.1.3` + matching `devEngines.packageManager` (`onFail: error` — pnpm v11 reimplemented `config / version / pkg` as native subcommands, so `error` is safe here). `devEngines onFail: error` is enough to block accidental `npm install` / `yarn install`; no `preinstall` guard needed.
- Bump `engines.node` `>=20` → `>=22` (pnpm 11 requires Node 22+).

### `pnpm-workspace.yaml`
- Promote `src/web` to a workspace package — eliminates the second `npm install` step the build used to run inside `src/web`.
- `minimumReleaseAge: 1440` (24-hour quarantine; excludes `@apify/*` and `@apify-packages/*`).
- `allowBuilds` allowlist: `esbuild`, `@sentry/cli`, `@apify/ui-icons` allowed; `protobufjs` explicitly denied (its postinstall is a deprecated helper we don't need). `strictDepBuilds: false` so a new build-script transitive doesn't fail `--frozen-lockfile` installs.
- Move the `@modelcontextprotocol/sdk: 1.29.0` override here (pnpm 11 reads overrides from the workspace manifest, not `package.json`).

### Phantom-dep cleanup
- Add `axios ^1.16.0` as a direct dependency. The code imports `AxiosRequestConfig` / `AxiosResponse` directly; under npm's flat tree it was hoisted from `apify-client`, but pnpm's strict isolation breaks that — same pattern as the architectural cleanups in apify-core's pnpm migration.

### Scripts
- `npm run` → `pnpm run` throughout.
- `build:web` becomes `pnpm --filter @apify/mcp-web-widget run build`. The root `pnpm install` covers `src/web` too (workspace install), so the previous "Build web widgets" CI step that did its own `npm ci --force && npm run build` inside `src/web/` is now redundant — the single root `pnpm run build` runs `build:core` + `build:web` + `postbuild` copy in one pass.
- `scripts/dev_standby.js` spawns `pnpm --filter ...` instead of `cd src/web && npm run`.

### CI
- Every install step collapses to:
  ```yaml
  - uses: apify/actions/pnpm-install@v1.1.2
  ```
  (`apify/actions/*` is the canonical location for the composite actions that used to live in `apify/workflows/*`.)
- Test matrix: Node 22, 24, 26. Node 20 dropped (pnpm 11 needs Node 22+); Node 26 added for early signal on the upcoming LTS.
- `manual_publish_to_npm`: `npm publish` → `pnpm publish --tag <…> --no-git-checks`. npm publish would be rejected under `devEngines onFail: error`; pnpm publish honours the dist-tag flag and inherits the OIDC auth from `setup-node`'s `registry-url`.
- `_update_release_metadata`: `npm version` → `jq` JSON edit (we only mutate one field — no pnpm bootstrap needed).
- `manual_release_stable` (MCPB packaging): `cp -r node_modules` would copy pnpm's symlinked `.pnpm/` tree, so the packager would either follow symlinks into the entire store or break. Replaced with `pnpm deploy --prod mcpb`, which produces a flat, hoisted, prod-only `node_modules` — the pnpm idiom for self-contained deployments.
- `_publish_to_pkg_pr_new` and `manual_release_stable` no longer need their separate `src/web` install step.
- `on_pull_request` paths-filter watches `pnpm-lock.yaml` / `pnpm-workspace.yaml` instead of `package-lock.json`.

### Dockerfile
- `corepack enable` pins the exact pnpm version from `package.json#packageManager` — no global install drift.
- Builder: `pnpm install --frozen-lockfile`.
- Runtime: `pnpm deploy --prod` into `/app` (pnpm's equivalent of `npm ci --omit=dev`, but produces a self-contained tree with no registry resolution at runtime).

### Docs
- `CLAUDE.md`, `DEVELOPMENT.md`, `README.md` updated. `DEVELOPMENT.md#Installation` now documents `corepack enable && pnpm install`.

## Notes for review
- The `github-registry-token` input on `apify/actions/pnpm-install` isn't passed — this repo has no `@apify-packages/*` deps. If we add one later, pass `github-registry-token: ${{ secrets.APIFY_SERVICE_ACCOUNT_GITHUB_REGISTRY_TOKEN }}` to the action.
- Node 20 dropped from the test matrix; Node 26 added.
- Verified locally: `pnpm install` clean (no build-script warnings), `pnpm run type-check`, `pnpm run lint`, `pnpm run test:unit` (711 passing), `pnpm run build` (TS core + web widgets + postbuild copy producing all three widget bundles under `dist/web/`).


.github/workflows/manual_publish_to_npm.yaml
```diff
@@ -33,23 +33,13 @@ jobs:
...
+            -   name: Install pnpm and dependencies
+                uses: apify/actions/pnpm-install@v1.1.2
             -   name: Check version consistency and bump pre-release version (beta only)
                 if: ${{ inputs.tag == 'beta' }}
                 run: node ./.github/scripts/before-beta-release.js
-            -   name: Build web widgets
-                working-directory: src/web
-                run: |
-                    npm ci --force
-                    npm run build
```
B4nan, line 50: this is no longer needed since `web` is a pnpm workspace, so install and build runs when invoking the root install/build automatically

package.json
```diff
@@ -5,8 +5,16 @@
   "description": "Apify MCP Server",
   "mcpName": "com.apify/apify-mcp-server",
   "engines": {
-    "node": ">=20.0.0"
+    "node": ">=22.0.0"
```
B4nan, line 8: this is technically a BC, not sure if important. node 20 is EOL already, so I'd just do this, if it should be considered breaking, just ship a major bump

MQ37: yes, people shout not be using EOL node version

MQ37 (approved):

package.json
jirispilka: I agree, ppl should not use node 20 but they do and it is breaking for them when they download our package via mcpc binary in Claude Desktop.

This was real life observation in sentry and in our github issues.

Unless you are strongly against, I would keep 20.
The main reason is that, if it starts failing again in Claude Desktop, we can't easily rollback it. The roll back will take 3-5 days

B4nan: Up to you, if I would release a major today, it would be without node 20.

jirispilka: So I run the analysis again and I want to correct myself.

I know that there was a problem with a node version but I forgot it was 18 and lower, not 20.

But the downgrade to 18 is not possible, there was [dependency problem in sentry](https://github.com/apify/apify-mcp-server/blob/master/DEVELOPMENT.md#nodejs-version-policy)

image

Now, if you look at the image above in the last 90 days, version 20 was actually ok. 
It might go up, if we do a breaking change and use syntax not supported in 20 

We are not able to do anything for version 18 and lower.

There is another issue, supporting 20 in CI, we would have to downgrade pnpm to version 10 or change the github workflows.

So after some back and forth with claude, keeping node 20 would be a hassle.
I update documentation in this PR:  https://github.com/apify/apify-mcp-server/pull/873

B4nan: We could as well just lift the engines constraint back, but not testing it in the CI (keep pnpm 11, only remove node 20 from the testing matrix). We'd have to keep an eye on the dependency bumps, I don't think we really need to use Node 22 only API in our code.

But seeing the recent releases, people are dropping node < 22 a lot already.

PR #870, opened 2026-05-20
fix(docker): repair pnpm deploy and add build smoke test to CI

MQ37:
## Context

The Dockerfile broke during the npm → pnpm 11 migration (#869) — `pnpm deploy --prod /app` in the runtime stage fails with `ERR_PNPM_NOTHING_TO_DEPLOY` because the workspace now contains two packages. No CI job builds the Dockerfile, so the regression slipped through; the Docker Hub auto-build integration (which the README directs external users to) is the only place it would have surfaced.

## Solution

Two changes:

1. Fix the runtime-stage `pnpm deploy` invocation.
2. Add a `docker_build_smoke_test` job to `_check_code.yaml` so this doesn't happen again. `_check_code.yaml` is already wired into `on_pull_request.yaml` and `on_master.yaml`, so the job runs on every PR and every push to master with no extra plumbing.

## Worth your attention

- **Why both `--filter` and `--legacy`.** `--filter "@apify/actors-mcp-server"` is mandatory per the [pnpm deploy docs](https://pnpm.io/cli/deploy) — without it pnpm can't pick a package from the workspace root. `--legacy` then opts out of pnpm v10+'s `inject-workspace-packages` requirement; enabling that setting workspace-wide breaks dev tooling (watch builds, VSCode breakpoints — pnpm/pnpm#9386, #8975), and the runtime server doesn't depend on any workspace package, so the legacy path is the right scope. It also produces real file copies instead of symlinks to the global pnpm store (pnpm/pnpm#9883), which matters inside a Docker image where the store isn't present.
- **Smoke test needs no real `APIFY_TOKEN`.** The MCP `initialize` handshake never hits the Apify API, so passing `APIFY_TOKEN=apify_api_dummy_for_smoke` is enough. This lets the job run on fork PRs without secrets.
- **GHA cache for buildx layers.** `cache-from`/`cache-to: type=gha` keeps re-runs under ~30s after the first build.

### DIFF
--- a/.github/workflows/_check_code.yaml
+++ b/.github/workflows/_check_code.yaml
@@ -77,3 +77,44 @@ jobs:
                 uses: apify/actions/pnpm-install@v1.1.2
             -   name: Type checks
                 run: pnpm run type-check
+
+    docker_build_smoke_test:
+        # The Dockerfile is consumed by the external Docker Hub auto-build integration
+        # (see README) but no other CI job builds it, so breakage — like the pnpm v10+
+        # `pnpm deploy` regression — only surfaces post-merge. This job builds the image
+        # on every PR / master push and exercises the entrypoint to keep that surface honest.
+        name: Docker build & smoke test
+        runs-on: ubuntu-latest
+
+        steps:
+            -   uses: actions/checkout@v6
+
+            -   name: Set up Docker Buildx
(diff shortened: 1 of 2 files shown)


MQ37:
FYI @jirispilka

PR #876, opened 2026-05-20
feat(x402): expose multi-scheme accepts[] and structuredContent

MQ37:
closes https://github.com/apify/ai-team/issues/169

## Context

apify-core PR #27039 ships x402 `upto` scheme support alongside `exact`, so the 402 `payment-required` response now carries multiple entries in `accepts[]`. Two gaps surfaced on the MCP-server side:

1. The flat `_meta.x402` advertising on paid tools only included the first entry, hiding the other scheme from clients that don't probe the 402 round-trip.
2. The 402 tool-result response only populated `content[0].text` JSON, missing the `structuredContent` field that the x402 MCP transport spec requires (preferred by spec-compliant clients).

## Solution

- `decorateToolSchema` now writes flat preferred-scheme fields **and** the full `accepts[]` array. Preference order is `exact` → `upto` → first remaining entry, so clients that read only the flat fields stay backwards-compatible and keep the pre-#876 contract intact.
- `buildPaymentRequiredResponse` now also sets `structuredContent` on the 402 tool result, matching coinbase/x402 `specs/transports-v2/mcp.md`.
- Tightened `X402PaymentRequirements` from `Record<string, unknown>` to a concrete `{ x402Version, resource, accepts: X402PaymentAccept[] }` type.

## Worth your attention

- **Back-compat**: existing clients reading `_meta.x402.{scheme,network,asset,payTo,amount}` keep working — those flat fields are still there, just resolved against the preferred entry instead of the first one.
- **Why prefer `exact` for back-compat**: switching their default to `upto` would silently move them onto the Permit2 settlement path and require a one-time on-chain allowance approval they don't know to do. We prefer `exact` for flat fields to keep the pre-#876 contract intact.

## Follow-up

- Companion mcpc PR plumbs the client side (upto signer + scheme selection).
- `_meta["x402/payment-response"]` settlement receipts are still not returned to clients — separate spec gap, design question (Apify settles async).

### DIFF
--- a/src/payments/x402.ts
+++ b/src/payments/x402.ts
@@ -206,7 +206,8 @@ export class X402PaymentProvider implements PaymentProvider {
         }
         const metaRecord = cloned._meta as Record<string, unknown>;
         if (!metaRecord.x402) {
-            const acceptsRaw = this.requirements?.accepts;
+            const reqs = this.requirements ? structuredClone(this.requirements) : undefined;
+            const acceptsRaw = reqs?.accepts;
             const accepts = Array.isArray(acceptsRaw) && acceptsRaw.length > 0 ? acceptsRaw : undefined;
             const preferred = accepts ? this.selectPreferredAcceptEntry(accepts) : undefined;
             metaRecord.x402 = {
--- a/src/payments/x402.ts
+++ b/src/payments/x402.ts
@@ -30,9 +30,9 @@ const X402_TOOL_INSTRUCTIONS = [
  * Preferred scheme order when selecting the flat fields exposed on `_meta.x402`.
  *
  * `exact` first to keep the flat-fields contract back-compatible with clients
- * that don't iterate `accepts[]` — they'll continue to sign `exact` payments
- * just like before this PR. Clients that walk `accepts[]` (post-#876 — the
- * current mcpc, the canary) can opt into `upto` via their scheme preference.
+ * that don't iterate `accepts[]` — they continue to sign `exact` payments as
+ * before this PR. Clients that walk `accepts[]` (post-#876 — the current
+ * mcpc, the canary) can opt into `upto` via their scheme preference.
  */
 const X402_PREFERRED_SCHEMES = ['exact', 'upto'] as const;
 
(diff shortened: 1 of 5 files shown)


TateLyman:
Ran a code-only pass on this server-side companion after reviewing `apify/mcpc#238`. Scope: PR diff only; no paid Apify call, no payment headers, no wallet signatures, no settlement attempt.

The core changes look directionally right: paid tools now expose both the flat back-compat `_meta.x402` fields and the full `accepts[]`, and the 402 tool result now mirrors the payload into `structuredContent` as the MCP transport spec expects.

One merge-order risk to keep visible: because this server PR intentionally prefers `upto` for the flat `_meta.x402` fields, older or partially-updated clients that only read the flat fields will move to the Permit2/`upto` path by default. That is fine for the new richer path, but it makes the companion client PR’s scheme-preference fix important before rollout: a user/session that pins `exact` should not still sign the preferred flat `upto` leg from proactive metadata handling.

I would treat rollout as a paired change:
- server exposes `accepts[]` and keeps flat fields for compatibility;
- client selection uses `accepts[]` plus the pinned `auto|upto|exact` policy everywhere, including proactive metadata and tool-result retry paths;
- tests cover metadata with both `upto` and `exact`, `schemePreference: 'exact'`, and no Permit2 allowance check.

Also worth keeping in the launch checklist: `_meta["x402/payment-response"]` / settlement receipt behavior is still a product/spec decision. Since `upto` settlement can be deferred, clients need a clear immediate receipt vs later settlement-status story so they do not confuse authorization, work execution, and actual collected usage.

No payment material was sent or signed.

src/payments/x402.ts
jirispilka, line 1: Claude flagged pre-existing issue, which is more prominent, because the accept is an array now

Object.freeze(cloned) is shallow: nested _meta.x402.accepts (and its extra) still share references with this.requirements, so the freeze gives a false sense of immutability. Either drop the freeze or deep-clone.

jirispilka (approved):
Looks good

Please update the PR body, 
In the body you said `upto` is preferred while in the code `exact` is preferred.

PR #893, opened 2026-05-21
fix(payments): reject standby Actors and sanitize loader errors under agentic payments

### LINKED ISSUE #875: Do not include the rag web browser when ?payment=x402
```
✗ Tool apify--rag-web-browser returned an error

Content:
Error calling tool "apify--rag-web-browser": Agentic payments are not supported for standby Actors.. Verify the tool name and input parameters.
```

MQ37:
Sorry, it is a bit chunkier :( Originally wanted to keep the change focused and simple but went down the rabbit hole and did other fixes and tiny refactor. Also found out that the dev server was using API token from the env var and was not getting that from the HTTP header, so now it is fixed.

## Context
Closes #875. Standby / MCP-server Actors were visible in `tools/list` and accepted by `call-actor` / `add-actor` under external payment providers (x402, Skyfire), even though the platform hard-rejects any standby Actor run on agentic payments. Agents discovered tools that could only fail later.

## Solution
- Filter standby + MCP-server Actors in `getActorsAsTools` when `paymentProvider` is set.
- Reject standby Actors at call-time with a precise message via a new `checkPaymentProviderStandbyConflict` guard that runs **before** the generic payment-required short-circuit (`call-actor` is `paymentRequired: true`, so without this the 402 fired first).
- Plumb `throwOnError` through `getActorsAsTools` so `add-actor` / `call-actor` see real load errors instead of empty results.

## Worth your attention
- **Sanitized errors are now type-safe.** New `ActorLoadError` class in `src/errors.ts` with a `kind` discriminator (`NOT_FOUND` / `LOAD_FAILED` / `STANDBY_PAYMENT_NOT_SUPPORTED`) and static factories that own the canonical messages. `add-actor` only forwards `error.message` when `error instanceof ActorLoadError`; anything else gets logged server-side and masked. `resolveAndValidateActor` now branches on `error.kind === NOT_FOUND` instead of `errMsg.includes('was not found')` — so message-text changes cannot silently break the not-found path.
- **`dev_server.ts` now mirrors `apify-mcp-server-internal`'s auth middleware.** Token comes from `x-apify-authorization` / `authorization` / `?token=` headers, never `process.env.APIFY_TOKEN`. Token requirement is skipped when `?payment=<provider>` is set. The per-request env fallback in `server.ts:833` was masking the fact that x402 tests were silently running with a real token.
- **Shared cache access centralized in `getActorDefinitionCached`** (`src/utils/actor.ts`, next to the existing `getActorMcpUrlCached`). Replaces three duplicated get-or-fetch-and-set blocks. `getActorMcpUrlCached` now routes through it too, so a definition cached by one call path (e.g. `add-actor`) can satisfy an MCP-URL lookup elsewhere without a second API call.
- **`actorDefinitionPrunedCache` → `actorDefinitionCache`.** The cache holds `ActorDefinitionWithInfo` (pruned definition **plus** full info), not a standalone `ActorDefinitionPruned` — the old name was a leftover from an earlier shape.
- **Single source of truth** for the standby guard — `checkPaymentProviderStandbyConflict` in `call_actor_common.ts` replaces the previous inline check inside `callActorPreExecute`.

## Follow-up
- The matching internal-repo PR apify/apify-mcp-server-internal#555 adds equivalent integration tests against the production server (covers `add-actor` rejection + `tools/list` filtering for both x402 and skyfire). **Merge it only after this PR is released**, since it consumes the new `@apify/actors-mcp-server` behaviour.


TateLyman:
Source-readback note on the new standby/payment guard: the immediate `call-actor` path now rejects standby/MCP-server Actors before the generic payment-required short-circuit, but the long-running task path appears to bypass that guard.

Flow I am reading:

- `call-actor` declares `execution.taskSupport: 'optional'`, so clients can send `request.params.task` for it.
- In `src/mcp/server.ts`, the `if (request.params.task)` branch creates the task and returns before the new `checkPaymentProviderStandbyConflict(...)` block runs.
- `executeToolAndUpdateTask()` later only checks `paymentRequiredResult` before executing/storing the task result; it does not run the standby/MCP-server conflict check.

That means this intended behavior:

> reject BEFORE the payment short-circuit so the agent gets the precise reason instead of a generic 402

looks true for the normal tool call, but not for task-mode `call-actor`. A streamable HTTP client in `?payment=x402` mode calling `call-actor` with `actor: 'apify/rag-web-browser'` plus a `task` envelope would get a task created, and the stored result would be the generic x402 payment-required result rather than the new "standby Actor cannot be accessed using a third-party payment provider" response.

Suggested patch direction: move the standby rejection block above the long-running task branch, or run the same check inside `executeToolAndUpdateTask()` before the `paymentRequiredResult` handling. A focused regression would mirror the existing x402 standby test but add `request.params.task` and then assert `tasks/get` resolves to the standby-specific rejection, not the generic x402 402.

No payment headers, signatures, credentials, or paid calls used for this pass; just PR source/test readback.

src/dev_server.ts
```diff
@@ -325,9 +389,11 @@ function isInitializeRequest(body: unknown): boolean {
 // --- Entry point: start the server when run directly ---
 
 if (process.argv[1] === fileURLToPath(import.meta.url)) {
+    // Token is no longer read from the environment per-request — clients pass it
+    // via `Authorization: Bearer <token>` (or use `?payment=<provider>`). Warn if
+    // unset so local dev runs without payment mode fail fast with a clear hint.
     if (!process.env.APIFY_TOKEN) {
-        log.error('APIFY_TOKEN is required but not set in the environment variables.');
-        process.exit(1);
+        log.warning('APIFY_TOKEN is not set. Clients must pass `Authorization: Bearer <token>` or `?payment=<provider>` per request.');
```
jirispilka: I agree, the previous version was a convenient hack. I would delete the code here right away.

Without token, you will get error message anyway

jirispilka (changes requested):
Ops, the standby is a real issue, thanks for the fix!

I like the cache simplification.

For the rest of the code, I'm afraid it is quite complicated with a lot of ifs. I'm afraid that the `server.ts` is the main culprit. AI coding agents simple struggle with it and are lost.

I asked claude for a help (obviously). Sadly, it won't be much simpler.

Claude:

**1. Task-mode bypasses the standby guard** (`server.ts:985-1018` returns before line 1020). `call-actor` declares `taskSupport: 'optional'`, so `task` + `payment=x402` + standby Actor skips `checkPaymentProviderStandbyConflict` entirely.
- Fix: move lines 1020-1043 above the task branch at line 985.
- Test: add `runIf(streamable-http)` integration test — `call-actor` + `task` + `payment=x402` + `apify/rag-web-browser` → standby rejection (not 402, not platform error).

**2. Unify standby message** — two different strings for the same condition, both pinned in integration tests.
- `call_actor_common.ts:312-319` → use `ActorLoadError.standbyPaymentNotSupported(normalizedActorName).message`.
- `suite.ts:2885` → update asserted substring to canonical message.

**3. Replace `throwOnError` boolean flag with split functions** — the flag changes the return contract (throw vs. return), classic antipattern. Only two callers use `throwOnError: true`, both pass a single Actor.
- Add `loadActorAsTool(name, client): Promise<ToolEntry | ActorLoadError>` in `actor_tools_factory.ts`.
- `getActorsAsTools(names, client)` stays as the bulk/forgiving path; drop the `throwOnError` option and the throw loop at lines 363-369.
- Update `add-actor` and `resolveAndValidateActor` to use `loadActorAsTool` with `result instanceof ActorLoadError` checks. Removes the `.catch` dance at `call_actor_common.ts:432-439`.

#### Should fix

- Unit tests: `getActorsAsTools` filter under `paymentProvider`, `loadActorAsTool` returns each `ActorLoadError` kind, `checkPaymentProviderStandbyConflict` standby vs normal, task-mode standby reject.

#### Nits

- 4× `...(paymentProvider && { paymentProvider })` at `server.ts:448, 473, 487, 518` — drop the spread.
- Two-line comment on `getActorsAsTools` clarifying it's the forgiving bulk path.

#### Follow-ups

**1. Refactor PR: collapse the policy**

After #893, the rule still lives in multiple spots: two filters in `getActorsAsTools`, double cache fetch in `checkPaymentProviderStandbyConflict`, and `isActorInfoMcpServer` overlaps with standby by definition. Consolidate:

- Single helper `isStandbyActor(actorInfo)` in `tools/utils.ts`
- One filter expression in `getActorsAsTools` — `mcpActors` auto-empties under `paymentProvider`, so the MCP branch deletes
- Drop `getActorMcpUrlCached` from the call-time guard (definition is enough)

Separate PR because it's structural, not bug-fix, and the loader signature change needs `apify-mcp-server-internal` heads-up. ~50 lines deleted, ~20 added.

**2. Issue: pre-payment validation hook**

Today the generic `CallToolRequestSchema` handler hardcodes `tool.name === ACTOR_CALL` to know when to run the standby check. Next tool that needs a precise pre-402 error will accrete another branch.

Move it to tool metadata: `execution.prePaymentCheck?(toolArgs, ctx)`. Generic handler just calls it; standby logic stays inside `call-actor`. Open an issue, design later — no urgency since today's leakage is one tool.

TateLyman:
I could not push directly to this Apify-owned branch from this account, so I opened the requested-changes follow-up from my fork: #903. It addresses the task-mode standby bypass, unifies the standby rejection message, removes the throwOnError dual-mode path in favor of loadActorAsTool(...), and adds the streamable task-mode regression test. Local checks passed: type-check, lint (existing warnings only), unit suite, targeted streamable integration, and build:core.

MQ37:
@jirispilka thank you for the review and great catch with the long running tasks! I also refactored the `throwOnError` and we are instead returning the errors in the return object together with Actors.

jirispilka (approved):
Looks good.

I like that it was simplified! 

There were a couple of nits and since we need to release it today I went ahead and created a pr here: https://github.com/apify/apify-mcp-server/pull/910

Issue #560, opened 2026-03-14
Unable to connect to extension server. Please try disabling and re-enabling the extension.


Ghassan1988:
Trying to connect Apify to claude and i am receiving the below message after installing apify-mcp-server

Unable to connect to extension server. Please try disabling and re-enabling the extension.

jirispilka:
Preferably, I’d recommend using the remote Apify MCP with the following configuration:

```
{
  "mcpServers": {
    "apify": {
      "command": "npx",
      "args": [
        "-y",
        "mcp-remote",
        "https://mcp.apify.com/",
        "--header",
        "Authorization: Bearer <YOUR_API_TOKEN>"
      ]
    }
  }
}
```
It seems that Claude Desktop can sometimes behave inconsistently with the extension.

You can find the details here: mcp.apify.com

<img width="737" height="947" alt="Image" src="https://github.com/user-attachments/assets/13e4ee48-72e0-49c7-92b8-23b0e8e27310" />

jirispilka:
> > Hey, could you please provide more details? What claude client? Claude desktop, Claude code, claude.ai web? How exactly are you trying to install Apfiy MCP Server?
> 
> I connected apify via connectors in claude desktop. So the MCP client is claude desktop itself.

Thank you, just to clarify, once you uninstalled and added it again, it works for you, right?

YagneshKhamar:
for temporary solution which worked for me was this way go to 
https://mcp.apify.com/
get the link for tools what you want,
from there then go to claude -> settings -> add custom connectors -> add name -> add link and -> add token
then it will ask to connect which will redirect to browser and from browser we can then verify the connection and then it will work fine

jirispilka:
Thanks @YagneshKhamar for sharing the solution — that's exactly the right approach!

For anyone still hitting this: the most reliable fix is to connect directly to the remote server. In Claude Desktop go to **Settings → Connectors → Add custom connector**, use `https://mcp.apify.com` as the URL, and follow the OAuth flow.

Full setup and troubleshooting steps are in the [docs](https://docs.apify.com/platform/integrations/claude-desktop#troubleshooting).

Closing this — please open a new issue if the problem persists with details and logs.

chellasc:
Thanks @YagneshKhamar for sharing the solution. Worked for me as well.

PR #897, opened 2026-05-21
docs(test-organization): Add documentation for test organization across public and internal repositories

jirispilka:
## Problem

Internal's integration suite (~50 tests, ~1300 lines) duplicates much of what the public package already tests — tool loader, widget metadata, prompt registry, structured output schemas. Every fixture rename or schema change forces parallel edits in both repos.

## What this PR changes

Lands the **rule** that governs the split, in the durable docs where contributors and agents will actually see it:

- `CLAUDE.md` — concise summary in the public/internal section, with a pointer to DEVELOPMENT.md.
- `DEVELOPMENT.md` — full "Test organization across repos" section: four-layer ownership model, "keep by default" rule with three explicit drop conditions, and the decision flow for new tests.

The previous `res/test_organization_across_repos.md` spec is deleted — `res/` accumulates stale notes; the rule belongs in the docs that drive development.

## Why "keep by default"

The hosted server wraps the package with extra code — auth, rate limiter, Caddy, response handlers. Any of that can change, drop, or delay something the package produced, and the package's own tests don't see that wrapping. Duplication costs us tighter coupling and harder refactors. It still beats a silent regression on mcp.apify.com.

A test moves out of internal only when all three hold: equivalent public test by name, no hosted code can touch the behavior, and a quiet removal would surface fast.

## The actual split work

Tracked in the internal repo where the work happens: [apify/apify-mcp-server-internal#561](https://github.com/apify/apify-mcp-server-internal/issues/561). That issue covers the `createContractSuite()` / `createHostedSuite()` factory split in `apify-mcp-server-internal`, the per-test migration matrix, and the mirroring of this rule into internal's `AGENTS.md` ([internal PR #560](https://github.com/apify/apify-mcp-server-internal/pull/560)).


MQ37 (approved):
Yes, I like this proposal and we should definitely deduplicate and simplify the testing suite - it is one of the fist core steps for improving our experience.

But we should also focus and really think through what smoke tests will also be present in the internal repo as we do not want to miss out some important feature that could break in the internal repo - there may be some legacy logic or handling.

jirispilka:
> But we should also focus and really think through what smoke tests will also be present in the internal repo as we do not want to miss out some important feature that could break in the internal repo - there may be some legacy logic or handling.

@MQ37 I agree, I changed the wording to never delete tests thinking that internal repo will cover it.
Also, it is ok to have duplicated tests, better safe then sorry.

I updated CLAUDE.md and DEVELOPMENT.md

Follow up issue in the internal repo: https://github.com/apify/apify-mcp-server-internal/issues/561
Update AGENTS.md in internal: https://github.com/apify/apify-mcp-server-internal/pull/560

MQ37 (approved):

PR #903, opened 2026-05-23
fix(payments): handle standby Actor task calls in payment mode

TateLyman:
## What

Follow-up to #893 after the requested-changes review:

- moves the standby/MCP Actor payment-mode guard ahead of task handling so `call-actor` task mode does not bypass the precise standby rejection;
- keeps task-mode protocol shape valid by creating a task and storing the standby rejection as the completed task result;
- unifies the standby rejection copy through `ActorLoadError.standbyPaymentNotSupported(...)`;
- replaces the `getActorsAsTools(..., throwOnError: true)` dual-mode path with a single-Actor `loadActorAsTool(...)` path for `add-actor`/`call-actor` resolution;
- adds the streamable HTTP task-mode regression test.

## Testing

```text
pnpm run type-check
pnpm run lint
pnpm run test:unit
pnpm exec vitest run tests/integration/actor.server_streamable.test.ts -t "standby Actor via call-actor|task-mode call-actor|adding a standby Actor"
pnpm run build:core
```

`pnpm run lint` completes with existing warnings only.

## Notes

I could not push directly to the branch used by #893 because it belongs to the Apify org, so this PR carries the follow-up patch from my fork.


MQ37:
Thank you for the contribution, we already merged that PR. Please before doing any changes like this and opening PR coordinate with the maintainers. It is always better to start to work on some smaller new issue/PR instead of working on PR mid-development.

PR #909, opened 2026-05-25
fix: Add get-actor-run to auto-injected tools and update related tests

### LINKED ISSUE #908: Actor tools reference `get-actor-run` in `nextStep` but it's not exposed in MCP-mode actor-only sessions
## Problem

Actor tools' `nextStep` points the client at `get-actor-run` for polling — but in default `MCP` mode the tool is not auto-injected when a session uses only `?actors=<name>`:

```
$ mcpc connect 'https://mcp.apify.com?actors=apify/instagram-scraper' @s
$ mcpc @s tools-list
* get-dataset-items
* get-key-value-store-record
* abort-actor-run
* apify--instagram-scraper        ← returns "Use get-actor-run with runId=... to poll for completion."
```

The auto-inject gate in `src/utils/tools_loader.ts:280-282` requires either `call-actor` or `APPS` mode:

```ts
if (hasCallActor || (hasActorTools && mode === ServerMode.APPS)) {
    toolsToInject.push(defaultGetActorRun);
}
```

But `pollHint()` in `src/tools/core/actor_run_response.ts:269` produces the same `Use get-actor-run …` string for every long-running actor tool regardless of mode, so the client is told to poll a tool it doesn't have.

Workaround: add `&tools=runs` to the URL.

## Fix

Inject `get-actor-run` whenever `hasActorTools` is true, not only in `APPS` mode. The `nextStep` template is mode-agnostic; its prerequisite tool list should be too.

jirispilka:
Closes #908.

Auto-inject `get-actor-run` whenever any actor-running entrypoint is present (direct actor tools, `add-actor`, `call-actor`, or `get-actor-run` itself). Previously gated behind `call-actor || apps mode`, so `?actors=<name>` sessions in default MCP mode were told to poll a tool they didn't have.

- Move `defaultGetActorRun` into `AUTO_INJECTED_TOOLS`
- Drop the now-redundant mode-specific guard
- Add regression test (`getToolsForServerMode`, default + apps mode)
- Update integration test counts/comments that hardcoded the old behavior

### DIFF
--- a/src/utils/tools_loader.ts
+++ b/src/utils/tools_loader.ts
@@ -30,6 +30,7 @@ import { SERVER_MODES, ServerMode } from '../types.js';
  * KV record → abort.
  */
 export const AUTO_INJECTED_TOOLS: readonly ToolEntry[] = [
+    defaultGetActorRun,
     getDatasetItems,
     getKeyValueStoreRecord,
     abortActorRun,
@@ -183,7 +184,7 @@ export function getToolsForServerMode(input: Input, actorTools: ToolEntry[], mod
             toolsByName.set(tool.name, tool);
         }
     }
-    // Widgets are apps-only and not in any category; include for direct selection
+    // Widgets are apps-only and not in any category; include it for direct selection
     if (mode === ServerMode.APPS) {
         for (const widget of WIDGET_BY_BASE_TOOL.values()) {
             toolsByName.set(widget.name, widget);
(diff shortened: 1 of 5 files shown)


src/utils/tools_loader.ts
```diff
@@ -263,18 +264,10 @@ export function getToolsForServerMode(input: Input, actorTools: ToolEntry[], mod
...
-    // needs that tool available. `call-actor-widget` returns immediately and the widget UI
-    // polls run status itself — that path doesn't drive the auto-inject decision here.
-    if (hasCallActor || (hasActorTools && mode === ServerMode.APPS)) {
-        toolsToInject.push(defaultGetActorRun);
-    }
-    if (hasCallActor || hasActorTools || hasAddActorTool || hasGetActorRun) {
-        toolsToInject.push(...AUTO_INJECTED_TOOLS);
-    }
+    // Inject run-workflow helpers whenever any actor-running entrypoint is present; de-dup pass below drops repeats.
+    const toolsToInject: ToolEntry[] = (hasCallActor || hasActorTools || hasAddActorTool || hasGetActorRun)
```
MQ37: nit: honestly to me the previous if version was more readable

MQ37 (approved):
Let's ship, thanks!

src/utils/tools_loader.ts
jirispilka: Fixed

PR #919, opened 2026-05-27
chore: apply oxfmt formatting [internal]

jirimoravcik:
## Summary

Mass reformat — applies `oxfmt --write .` across the codebase using the tooling landed in #915, and uncomments the `format-check` CI step.

- 180 files reformatted.
- One manual touch-up in `evals/shared/config.ts`: oxfmt's method-chain wrapping moved a `// eslint-disable-next-line no-control-regex` off its target line, re-introducing the warning. Repositioned the directive directly above the regex `.replace` (oxfmt then wraps the chain in parens and keeps it adjacent — both formatter and linter happy).

## Verification

- `pnpm run format:check` clean
- `pnpm run lint` — 3 warnings, 0 errors (same baseline as `master`; no new warnings)
- `pnpm run type-check` passes
- `pnpm run test:unit` — 800 passed, 1 skipped

Part of apify/apify-core#28164.

## Follow-up

- A separate PR will add a `.git-blame-ignore-revs` entry with this commit's squash SHA so `git blame` skips it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XEbAE25XHNNtQLxiXDdfHz)_


jirispilka (approved):
Thanks!

Mostly looks better than before.

There are some changes that do not look very nice but .... let's go

@MQ37 fyi: dedent does not look nice and for the strings that were split to multiple lines 

If they fit 120 chars limit

before:
```
'aaaaa'
+ 'bbbbb'
```

after:
```
'aaaa' + 'bbbbb' 
```

PR #920, opened 2026-05-27
fix: Update README to clarify tool selection and enhance descriptions

jirispilka:
The prefix should be `docs:` but I want to trigger release workflow


MQ37:
I think the `call-actor` does not return the preview anymore - otherwise LGTM :+1:

PR #921, opened 2026-05-27
chore: Drop dead userRentedActorIds plumbing

### LINKED ISSUE #805: chore: drop dead `userRentedActorIds` plumbing
## Problem

`userRentedActorIds` is dead code in this repo after #737. No tool reads it, but it is still parsed off `_meta`, threaded through `InternalToolArgs`, and forwarded to three `tool.call(...)` sites:

```ts
// src/mcp/server.ts
const userRentedActorIds = meta?.userRentedActorIds;          // line 766
// ...
await tool.call({ /* ... */ userRentedActorIds });            // lines 932, 968, 1342
```

Also declared on `InternalToolArgs` (`src/types.ts:144`) and on the `_meta` shape (`src/types.ts:677`).

## Fix

Drop `userRentedActorIds` from `InternalToolArgs`, the `_meta` shape, the extraction in `mcp/server.ts`, the `executeToolAndUpdateTask` signature, and the three `tool.call(...)` sites.

## Out of scope

apify-mcp-server-internal#500 covers dropping the upstream Mongo lookup that populates `_meta.userRentedActorIds`; this issue is the receiver-side cleanup in this repo.

MQ37:
## Context

`userRentedActorIds` is dead code in this repo after #737. The metadata is parsed off `_meta`, threaded through `InternalToolArgs` / `executeToolAndUpdateTask`, and forwarded to three `tool.call(...)` sites — none of which consume it. Closes #805.

## Solution

Drop the type fields (`InternalToolArgs.userRentedActorIds`, `ApifyRequestParams._meta.userRentedActorIds`), the `meta?.userRentedActorIds` extraction in `mcp/server.ts`, the `executeToolAndUpdateTask` parameter, and the three pass-through sites.

Also strip the matching stale example from `call_actor_widget.ts`'s schema comment: `async` was replaced by `waitSecs` in #825 and `previewOutput` was deprecated even earlier, so the "stray keys such as `async` or `previewOutput`" example is doubly wrong. Comment now just says "stray keys" — the schema's contract holds regardless of which key.

## Worth your attention

- **Receiver-side only** — internal repo's Mongo lookup that populated `_meta.userRentedActorIds` is being removed in parallel in apify-mcp-server-internal#568. Order doesn't matter: once neither side reads or writes the field, both sides can land independently.
- **`InternalToolArgs` is a public interface** — removing an optional field is technically a wire-shape change for any external consumer that destructured it. Grep confirms no in-tree tool reads it; external consumers couldn't have been using it for anything useful since search filtering moved upstream.
- **Test fixtures unchanged** — `tests/unit/tools.call_actor_widget.response.test.ts` and `tests/integration/suite.ts` still use `async` / `previewOutput` as arbitrary stray-key fixtures. Those tests assert "schema strips unknown keys", which is still what they exercise; the field names are just convenient unknowns.

## Follow-up

- apify-mcp-server-internal#568 — drops the upstream Mongo lookup that fed `_meta.userRentedActorIds`


jirispilka (approved):
Thanks 💪🏻

PR #923, opened 2026-05-27
refactor: introduce TOOL_TYPE constant for tool discriminator

### LINKED ISSUE #339: Create a new ToolType to remove string literals
```
/**
 * Tool type constants
 * Use these constants instead of string literals for better type safety and maintainability.
 */
export const ToolType = {
    INTERNAL: 'internal',
    ACTOR: 'actor',
    ACTOR_MCP: 'actor-mcp',
} as const;

/**
 * Type for tool type values.
 */
export type ToolType = typeof ToolType[keyof typeof ToolType];
```

MQ37:
## What
Replace string literals (`'internal'`, `'actor'`, `'actor-mcp'`) used as the `ToolEntry` discriminator with a `TOOL_TYPE` const object + derived type union.

```ts
export const TOOL_TYPE = {
    INTERNAL: 'internal',
    ACTOR: 'actor',
    ACTOR_MCP: 'actor-mcp',
} as const;
export type TOOL_TYPE = (typeof TOOL_TYPE)[keyof typeof TOOL_TYPE];
```

## Why
- Single rename target instead of ~30 scattered string literals.
- Proper IDE go-to-definition / find-references on the discriminator.
- Typos become compile errors instead of silent dead branches.
- Matches the existing repo convention for `as const` enums (`TOOL_STATUS`, `FAILURE_CATEGORY`, `ACTOR_LOAD_ERROR_KIND`) — see `CONTRIBUTING.md:194-206`. `TOOL_TYPE.ACTOR` reads as a constant lookup, not as a static method on a class.

Closes #339. Broader cleanup (`ServerMode` rename to match the convention, `CONTRIBUTING.md` fixes) tracked in follow-up #924.

## How
- `src/types.ts`: define `TOOL_TYPE` const + derived type union; switch the 3 union members (`HelperTool`, `ActorTool`, `ActorMcpTool`) from literal types to `typeof TOOL_TYPE.X`.
- Replace every raw discriminator literal in `src/` and `tests/unit/` with the corresponding `TOOL_TYPE.*` constant — switch cases, `.filter(... === ...)`, `if (... === ...)`, and tool-definition `type:` fields.
- `src/index_internals.ts`: **not** re-exported. `apify-mcp-server-internal` does not consume this discriminator today; keeping the surface minimal. Trivial to add back later if needed.

## Compatibility
- Wire/runtime values are unchanged — the const resolves to the same strings, so JSON serialization, persisted state and external consumers are unaffected.
- The discriminated union shape is identical (`typeof TOOL_TYPE.ACTOR` is `'actor'` as a TS literal type).

## Diff stats
38 files, +108 / −110. Mechanical 1-for-1 replacement; the only net additions are the ~8-line `TOOL_TYPE` const block in `src/types.ts`.

## Verification
- ✅ `pnpm run type-check`
- ✅ `pnpm run lint` (0 errors; 3 pre-existing warnings unrelated to this change)
- ✅ `pnpm run test:unit` — 800 passed / 1 skipped
- ✅ `pnpm format:check`

## Follow-up
Add the `beta` label to publish a `pkg.pr.new` canary so the internal repo (`apify-mcp-server-internal`) can pin and validate against it before this lands.

### DIFF
--- a/src/index_internals.ts
+++ b/src/index_internals.ts
@@ -19,7 +19,6 @@ import {
 } from './tools/index.js';
 import { actorNameToToolName } from './tools/utils.js';
 import type { ActorStore, ServerCard, ToolCategory } from './types.js';
-import { ToolType } from './types.js';
 import { parseCommaSeparatedList, parseQueryParamList, readJsonFile } from './utils/generic.js';
 import { redactSkyfirePayId } from './utils/logging.js';
 import { getExpectedToolNamesByCategories } from './utils/tool_categories_helpers.js';
@@ -44,7 +43,6 @@ export {
     type ActorStore,
     type ServerCard,
     type ToolCategory,
-    ToolType,
     processParamsGetTools,
     getActorsAsTools,
     getToolPublicFieldOnly,
--- a/src/index_internals.ts
+++ b/src/index_internals.ts
@@ -19,6 +19,7 @@ import {
 } from './tools/index.js';
 import { actorNameToToolName } from './tools/utils.js';
 import type { ActorStore, ServerCard, ToolCategory } from './types.js';
+import { ToolType } from './types.js';
 import { parseCommaSeparatedList, parseQueryParamList, readJsonFile } from './utils/generic.js';
 import { redactSkyfirePayId } from './utils/logging.js';
 import { getExpectedToolNamesByCategories } from './utils/tool_categories_helpers.js';
@@ -43,6 +44,7 @@ export {
     type ActorStore,
     type ServerCard,
     type ToolCategory,
+    ToolType,
     processParamsGetTools,
     getActorsAsTools,
     getToolPublicFieldOnly,
--- a/src/mcp/proxy.ts
+++ b/src/mcp/proxy.ts
@@ -4,7 +4,7 @@ import type { Client } from '@modelcontextprotocol/sdk/client/index.js';
 
 import { fixedAjvCompile } from '../tools/utils.js';
 import type { ActorMcpTool, ToolEntry } from '../types.js';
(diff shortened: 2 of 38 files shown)


jirispilka (approved):
Love this! I wanted to do it myself for a long time, you beat it to me.

The good part. Discoverability and rename safety are real wins, and the const-object pattern is a fine fit for that. No argument here.

One thing I'd still ask for. Drop ToolType from the index_internals.ts export. Nothing in
  apify-mcp-server-internal uses it  IMO. I would minimized the coupling now, we can add it later if required.


One small favor too ask :). Pre-approving as rename is mechanical.

Repo convention for as const enums is SCREAMING_SNAKE_CASE (CONTRIBUTING.md:194-206).

Existing examples: TOOL_STATUS, FAILURE_CATEGORY, ACTOR_LOAD_ERROR_KIND. ServerMode is the only outlier. My bad :(, let's not add a second.
```ts
 export const TOOL_TYPE = {
     ACTOR: 'actor',
     ACTOR_MCP: 'actor-mcp',
      INTERNAL: 'internal',
 } as const;

  export type TOOL_TYPE = (typeof TOOL_TYPE)[keyof typeof TOOL_TYPE];
```
Why:
- ToolType.ACTOR reads as a static method on a class,
- TOOL_TYPE.ACTOR reads as constant lookup.

Cheaper now than a follow-up rename PR later.
Broader cleanup (ServerMode rename + CONTRIBUTING.md fixes) → follow-up https://github.com/apify/apify-mcp-server/issues/924

MQ37:
@jirispilka Makes sense, thanks for the review! Updated and merging :rocket:

PR #926, opened 2026-05-28
refactor: Extract storage tool helpers

### LINKED ISSUE #890: refactor: Extract storage tool helpers
## Problem
Storage tools repeat two patterns: JSON code-fence wrapping (`` `\`\`\`json\n${JSON.stringify(v)}\n\`\`\`` ``) appears 8× across the 8 storage tool files (11× total across `src/tools/common/` once `abort_actor_run.ts`, `get_actor_output.ts`, and `run_collection.ts` are counted), and the not-found soft-fail block appears 5× across storage tools.

Cleaning these up before #878 / #884 lands keeps the per-sub-issue PRs small and avoids touching the same lines twice.

## Proposed solution
Pure refactor, no behavior change. New `src/tools/common/storage_helpers.ts` with `wrapJsonText()` and `buildStorageNotFound()`. Swap call sites across the 8 storage tools (and optionally the 3 non-storage `wrapJsonText` sites while we're there).

Tests-first per CLAUDE.md: helper unit tests + happy-path snapshot tests for the touched tools to prove no behavior change. Broader gap-fill stays in #879.

## Probe findings — bugs to fold into this PR

The refactor centralises not-found handling via `buildStorageNotFound()`, so the remaining inconsistencies are easiest to fix here:

- [ ] **Needs verification:** `get-dataset-items` with non-existent id may throw an uncaught SDK error and surface as a generic "Verify the tool name and input parameters." wrapper. If `listItems()` throws on missing dataset rather than returning `null`, the `if (!v)` branch at `get_dataset_items.ts:108` is dead. Add an integration probe against a non-existent dataset id; if confirmed, wrap in try/catch and route through `buildStorageNotFound`.
- [ ] Storage tools pass `datasetId` / `keyValueStoreId` raw to the SDK; actor tools normalize via `fixActorNameInput` (trim + strip quote/backtick wrappers) at every entry point. LLM-wrapped inputs like `` `username~my-store` `` 404 here but succeed for `call-actor`. Apply the same normalization (or extract its trim/strip half) before the SDK call.
- [ ] 6 of 8 storage tool files write their `description` as a plain backtick template with flush-left continuation lines; `get_dataset_items.ts` and `dataset_collection.ts` use `dedent`. Switch the six to `dedent` while these files are already being touched, so source indentation matches surrounding TS.

## Already resolved (struck from original scope)
- `get-key-value-store` not-found soft-fail — landed in #891.
- `dataset_collection.ts` / `key_value_store_collection.ts` instantiating `new ApifyClient({ token })` — no longer the case; both use the injected `toolArgs.apifyClient`.
- `FAILURE_CATEGORY` inconsistency across not-found paths — all five storage soft-fails now set `INVALID_INPUT` consistently.

Parent: #784

jirispilka:
Closes #890.

Pure refactor of the 8 storage tool files in `src/tools/common/`, plus two scoped bug fixes that fall out once the not-found path is centralized.

I check all the tools and the way we return tool results is quite messy. 
Follow up issue: https://github.com/apify/apify-mcp-server/pull/929

## New helpers
- `wrapJsonText(value)` — JSON code-fence wrap.
- `normalizeStorageId(id)` / `normalizeRecordKey(key)` — strip LLM-leaked quote/backtick wrappers before SDK lookup.
- `buildStorageNotFound(text)` — storage-domain soft-fail (SOFT_FAIL + INVALID_INPUT).
- `stripQuoteWrappers(s)` in `src/utils/generic.ts` — shared by `fixActorNameInput` and storage normalizers.

## Bug fixes folded in
- **`listItems()` / `listKeys()` 404 handling.** They throw `ApifyApiError`, not return null (only `.get()` / `.getRecord()` soft-catch). The `if (!v)` branches in `get-dataset-items`, `get-dataset-schema`, `get-key-value-store-keys` were dead — now caught and translated to `buildStorageNotFound`. Non-404 rethrows so 500s stay `INTERNAL_ERROR`.
- 6 tool descriptions migrated to `dedent` to match siblings.

## Flag for internal repo
`get-dataset-items` `structuredContent.datasetId` now reflects the **normalized** id, not the raw input. Previously `` "`ds-1`" `` round-tripped verbatim; now it returns `'ds-1'`. No contract pinned echo-verbatim, but per `CLAUDE.md` worth checking the internal contract suite.

## Verification
`type-check`, `lint`, `test:unit` all clean (823 passing, +12 new tests). Integration tests deferred to reviewer (need `APIFY_TOKEN`).

### DIFF
--- a/src/tools/common/storage_helpers.ts
+++ b/src/tools/common/storage_helpers.ts
@@ -2,12 +2,20 @@ import { FAILURE_CATEGORY, TOOL_STATUS } from '../../const.js';
 import { stripQuoteWrappers } from '../../utils/generic.js';
 import { buildMCPResponse } from '../../utils/mcp.js';
 
+const JSON_FENCE_PREFIX = '```json\n';
+const JSON_FENCE_SUFFIX = '\n```';
+
 /**
  * Wrap a JSON-serializable value in a Markdown ` ```json ` code fence.
  * Used by every storage tool that returns SDK payloads to the LLM.
  */
 export function wrapJsonText(value: unknown): string {
-    return `\`\`\`json\n${JSON.stringify(value)}\n\`\`\``;
+    return `${JSON_FENCE_PREFIX}${JSON.stringify(value)}${JSON_FENCE_SUFFIX}`;
+}
+
+/** Inverse of `wrapJsonText`; shares the fence constants so prod + tests can't drift. */
+export function parseFencedJson(text: string): unknown {
+    return JSON.parse(text.slice(JSON_FENCE_PREFIX.length, -JSON_FENCE_SUFFIX.length));
 }
 
 /**
@@ -32,4 +40,15 @@ export function buildStorageNotFound(text: string) {
  */
 export function normalizeStorageId(id: string): string {
     return stripQuoteWrappers(id);
+}
+
+/**
+ * Normalize a key-value store record key before SDK lookup.
+ *
+ * Mirrors `normalizeStorageId` but omits the apostrophe (`'`) from the wrapper
+ * class — Apify record keys allow `'` as a valid character
+ * (`/^[a-zA-Z0-9!\-_.'()]{1,256}$/`), so stripping it could corrupt a real key.
+ */
+export function normalizeRecordKey(key: string): string {
+    return key.trim().replace(/^[`"“”‘’]+|[`"“”‘’]+$/g, '').trim();
 }
--- a/src/tools/common/storage_helpers.ts
+++ b/src/tools/common/storage_helpers.ts
@@ -1,0 +1,49 @@
+import { FAILURE_CATEGORY, TOOL_STATUS } from '../../const.js';
+import { buildMCPResponse } from '../../utils/mcp.js';
+
+/**
+ * Wrap a JSON-serializable value in a Markdown ` ```json ` code fence.
+ * Used by every storage tool that returns SDK payloads to the LLM.
+ */
+export function wrapJsonText(value: unknown): string {
+    return `\`\`\`json\n${JSON.stringify(value)}\n\`\`\``;
+}
+
+/**
+ * Soft-fail not-found response for storage tools. Centralizes
+ * `isError: true` + SOFT_FAIL/INVALID_INPUT telemetry so call sites
+ * only supply the message.
+ */
+export function buildStorageNotFound(text: string) {
+    return buildMCPResponse({
+        texts: [text],
+        isError: true,
+        telemetry: { toolStatus: TOOL_STATUS.SOFT_FAIL, failureCategory: FAILURE_CATEGORY.INVALID_INPUT },
+    });
+}
+
+const STORAGE_ID_WRAPPERS: [string, string][] = [
+    ['`', '`'],
+    ['"', '"'],
+    ['“', '”'],
+    ['‘', '’'],
+];
+
+/**
+ * Normalize a dataset / key-value store id before SDK lookup.
+ *
+ * LLMs commonly wrap ids in markdown backticks or smart quotes, which the
+ * Apify API treats as distinct strings and 404s. This mirrors the trim/strip
+ * half of `fixActorNameInput` (no slash-padding — storage ids use `~`).
+ * Valid ids pass through unchanged.
+ */
+export function normalizeStorageId(id: string): string {
+    let s = id.trim();
+    for (const [open, close] of STORAGE_ID_WRAPPERS) {
+        if (s.startsWith(open) && s.endsWith(close) && s.length >= open.length + close.length) {
+            s = s.slice(open.length, -close.length).trim();
+            break;
+        }
+    }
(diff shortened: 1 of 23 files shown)


src/tools/common/storage_helpers.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/**
+ * Normalize a dataset / key-value store id before SDK lookup.
+ *
+ * LLMs commonly wrap ids in markdown backticks or smart quotes, which the
+ * Apify API treats as distinct strings and 404s. Mirrors the trim/strip
+ * half of `fixActorNameInput` (no slash-padding — storage ids use `~`).
+ */
+export function normalizeStorageId(id: string): string {
```
MQ37: nit: I would not introduce the `normalizeStorageId` as it is a think wrapper just calling the `stripQuoteWrappers` - I know in future it might make sense to introduce this func as there will be more processing steps but they are not there yet, so let's not introduce something that we don't need right now. I would just call `stripQuoteWrappers` directly at the `normalizeStorageId` call site and remove `normalizeStorageId`

src/tools/common/storage_helpers.ts
```diff
@@ -0,0 +1,57 @@
...
+import { buildMCPResponse } from '../../utils/mcp.js';
+
+const JSON_FENCE_PREFIX = '```json\n';
+const JSON_FENCE_SUFFIX = '\n```';
+
+/**
+ * Wrap a JSON-serializable value in a Markdown ` ```json ` code fence.
+ * Used by every storage tool that returns SDK payloads to the LLM.
+ */
+export function wrapJsonText(value: unknown): string {
```
MQ37: nit: I would move this to a separate utils as this could be used everywhere where we are returning JSON and not only for the storage tools. Otherwise really nice and cool helper!

src/tools/common/storage_helpers.ts
```diff
@@ -0,0 +1,57 @@
...
+    return `${JSON_FENCE_PREFIX}${JSON.stringify(value)}${JSON_FENCE_SUFFIX}`;
+}
+
+/** Inverse of `wrapJsonText`; shares the fence constants so prod + tests can't drift. */
+export function parseFencedJson(text: string): unknown {
```
MQ37: nit: this is used just for the tests, right? if yes I would move it to the test suite if we are not using that in the production logic.

src/tools/common/storage_helpers.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Mirrors `normalizeStorageId` but omits the apostrophe (`'`) from the wrapper
+ * class — Apify record keys allow `'` as a valid character
+ * (`/^[a-zA-Z0-9!\-_.'()]{1,256}$/`), so stripping it could corrupt a real key.
+ */
+export function normalizeRecordKey(key: string): string {
+    return key
+        .trim()
+        .replace(/^[`"“”‘’]+|[`"“”‘’]+$/g, '')
+        .trim();
```
MQ37: change: normalizeRecordKey and stripQuoteWrappers are essentially the same and we might need to update them in future and we can easily forget the other one. Thus I would move the regex to a const and keep them together so it is harder to miss updating the other one.

src/tools/common/storage_helpers.ts
```diff
@@ -0,0 +1,57 @@
...
+export function parseFencedJson(text: string): unknown {
+    return JSON.parse(text.slice(JSON_FENCE_PREFIX.length, -JSON_FENCE_SUFFIX.length));
+}
+
+/**
+ * Soft-fail not-found response for storage tools. Centralizes
+ * `isError: true` + SOFT_FAIL/INVALID_INPUT telemetry so call sites
+ * only supply the message.
+ */
+export function buildStorageNotFound(text: string) {
```
MQ37, line 10: note: I would then organize and move all helpers like `buildStorageNotFound`, `buildActorNotFoundResponse`, ... to a single place.

MQ37 (approved):
Cool, approving! But please address at least the `change:` review comment of mine. Otherwise LGTM 👍

src/tools/common/storage_helpers.ts
jirispilka: Agree, this will be a separate issue though https://github.com/apify/apify-mcp-server/pull/929

src/tools/common/storage_helpers.ts
jirispilka: Thanks. Done!

src/tools/common/storage_helpers.ts
jirispilka: Thanks. Done!

src/tools/common/storage_helpers.ts
jirispilka: Thanks. Done!

src/tools/common/storage_helpers.ts
jirispilka: Thanks. Done!

PR #927, opened 2026-05-28
fix(mcp): Validate webServerMcpPath stays within Actor standby origin

MQ37:
Fixes https://github.com/apify/apify-mcp-server/security/advisories/GHSA-6gr2-qh89-hxwm

## Context

`getActorMCPServerURL()` built the Actor standby MCP URL by string-concatenating
`https://<id>.apify.actor` with `webServerMcpPath` from the Actor definition.
Because that field is controlled by whoever publishes the Actor, a crafted
value could redirect the MCP client — and the caller's Apify token — to a
third-party host via URL-authority confusion (e.g. a path like
`.attacker.tld/mcp` resolves to `https://<id>.apify.actor.attacker.tld/mcp`).

## Solution

Resolve `webServerMcpPath` against the standby URL with `new URL(path, base)`
and reject anything whose `origin` doesn't equal the standby origin. Any
userinfo that survives parsing is stripped before the URL is handed to
`connectMCPClient`.

## Worth your attention

- **Some malicious forms are normalised rather than rejected.** `@host`,
  `.host` and `\\host` all resolve to *same-origin path components* under
  `new URL(path, base)`, so the test asserts the invariant ("never leaves the
  standby origin") instead of over-specifying "must throw". `//host` and
  `https://host` do throw.
- **Plain `Error`, not `ActorLoadError`.** This is a low-level utility wrapped
  by `getActorMcpUrlCached`, which already propagates non-404 errors; the bulk
  loader `getMCPServersAsTools` catches and skips the offending Actor. Matches
  the existing `throw new Error('Actor … not found')` pattern in the same file.
- **No defence-in-depth check in `client.ts`.** Source-side validation is
  sufficient; duplicating the guard in the transport layer would couple it to
  the Apify-actor domain without a concrete future call-site needing it.

### DIFF
--- a/src/mcp/actors.ts
+++ b/src/mcp/actors.ts
@@ -35,6 +35,12 @@ export function getActorMCPServerPath(actorDefinition: ActorDefinition | ActorDe
 
 /**
  * Returns the MCP server URL for the given Actor ID.
+ *
+ * `mcpServerPath` comes from an Actor's `webServerMcpPath` field, i.e. it is
+ * controlled by whoever published the Actor. Resolve it against the standby
+ * origin and reject anything that escapes that origin so a crafted path
+ * (`@host/...`, `.host/...`, `//host/...`, `https://host/...`) cannot redirect
+ * the MCP client — and the caller's Apify token — to a third-party host.
  */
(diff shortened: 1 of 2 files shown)


MQ37:
FYI @jirispilka

PR #929, opened 2026-05-28
docs: Tool response shapes and telemetry coupling analysis

jirispilka:
Big problem, surprisingly small fix 😌

## TL;DR

**Problem:** tool handlers build MCP responses by hand, so classification is opt-in — a handler can return `isError: true` and forget the telemetry, silently landing in the wrong `SOFT_FAIL + INTERNAL_ERROR` bucket. Same telemetry literal copy-pasted at 10 sites, ~8 hand-rolled JSON fences, six response shapes. Segment failure data is unreliable.

**Benefit:** handlers say *what happened* via `respond*` constructors; correct telemetry attached by construction, one JSON fence, and branded output so a bad shape **won't compile**. P1–P5 become a compiler concern, ~200+ lines of boilerplate go, failure telemetry becomes trustworthy — as small incremental PRs, no `apify-mcp-server-internal` coordination (the builder being replaced was never public).

---

## Summary
Analysis-only PR. Inventories the response shapes across `src/tools` handlers, the drift between them (~8 sites hand-roll the same `` ```json `` fence; six distinct shapes overall), and where the `isError` / `toolTelemetry` contract leaks into call sites.

Full write-up: [`res/tool_response_analysis.md`](./res/tool_response_analysis.md).

## Concrete problems (P1–P5)

- **P1** — `extractToolTelemetry` defaults `isError: true` + no telemetry to `SOFT_FAIL + INTERNAL_ERROR`. Semantically wrong combination; almost never what callers mean. `buildPermissionApprovalResponse` and several `server.ts` paths fall into it.
- **P2** — `get_dataset_schema.ts:85` returns `toolStatus: FAILED` without a `failureCategory`. Segment dashboards get a "category unknown" bucket.
- **P3** — `telemetry: { toolStatus: SOFT_FAIL, failureCategory: INVALID_INPUT }` is re-typed at **10** sites. No shared "this is a user error" constructor.
- **P4** — `isError: true` without telemetry is structurally legal — the wire format can't enforce classification.
- **P5** — No "informational, non-error" status. `"Dataset 'X' is empty."` is indistinguishable from data success.

## Options

| | Option | One-liner | Scope |
| --- | --- | --- | --- |
| **A** | Domain constructors over `buildMCPResponse` | Add `ok` / `okJson` / `empty` / `userError` / `serverError` wrappers. Keep `buildMCPResponse` public. | ~16 files |
| **B** | Discriminated-union return type | Tools return `ToolResult`, central encoder builds the wire shape. | ~30 files |
| **C** | HTTP-status analog | Tool returns `{ status: 200 \| 400 \| 404 \| 500, body }`. | ~30 files |
| **D** | Drop `buildMCPResponse` entirely | Tools return literal shapes; ad-hoc telemetry. | ~25 files |
| **E** | Go all the way | Discriminated union + encoder + delete every domain builder + flip the default trap into a compile error. | ~50 files, atomic, coordinated with `apify-mcp-server-internal` |
| **F** | **Intent constructors as the only door** | Small `respond*` family is the only public builder; `buildMCPResponse` un-exported; branded output so an unclassified error won't compile. | small, incremental PRs |

## Recommendation — Option F

**`respondOk` / `respondJson` / `respondEmpty` / `respondUserError` / `respondServerError`** become the only public way to build a response. `buildMCPResponse` goes private. The output type is branded, so a hand-rolled shape or a bare `isError: true` **won't compile** — that's Option E's structural guarantee (the bad shape becomes unconstructable), without E's blast radius.

Why F is cheap:

- **`buildMCPResponse` is not on the `internals` surface.** `apify-mcp-server-internal` consumes only the stripped wire shape; `toolTelemetry` is stripped inside this package before the boundary. Changing *how* we build the wire shape needs **zero cross-repo coordination** as long as the serialized output is unchanged.
- **Status determines category.** `respondUserError` → `SOFT_FAIL + INVALID_INPUT`, `respondServerError` → `FAILED + INTERNAL_ERROR` (derived from the caught error via the existing `classifyFailureCategory` / `getHttpStatusCode`). The 10 duplicated literals (P3) and the missing category (P2) collapse into the constructors.
- **`respondJson` owns the JSON fence** — the ~8 hand-rolled fences die.
- **`ToolEntry.call` is `Promise<object>` today** (untyped), so narrowing it to the branded `ToolResponse` is pure gain.

Migration is incremental and reversible: ship the 1-line P1 fix test-first; add the `respond*` family; migrate tools one PR at a time (wire output byte-identical → internal sees no change); delete the six domain builders; finally add the brand + narrow `call`.

Option E (full discriminated-union rewrite) is reserved for when a concrete new outcome variant with distinct wire behavior actually lands (`elicitation-required` / `partial-data`) — then it solves a current problem, not an imagined one. Option A is strictly dominated by F.

## Test plan
- [ ] Read [`res/tool_response_analysis.md`](./res/tool_response_analysis.md) — patterns, problems, options A–F, Option F deep dive
- [ ] Confirm Option F is the direction; if yes, file the P1 fix PR first, then the `respond*` family
- [ ] Confirm the `respond*` naming (vs a namespaced `response.ok()` / `response.userError()`)

### DIFF
--- a/res/tool_response_analysis.md
+++ b/res/tool_response_analysis.md
@@ -1,0 +1,433 @@
+# Tool response shapes & telemetry coupling
+
+Inventory of the response patterns across all `src/tools` handlers, where they drift, and where the telemetry contract leaks into call sites. Plus five design options with a recommendation.
+
+## 1. TL;DR
+
+Tool handlers return MCP results in **six different shapes** with no clear taxonomy. The JSON-fence wrapper drifts (some sites use `wrapJsonText`, others inline the same string literally). `isError` + `toolTelemetry` are **structurally independent** but **semantically coupled**: every error site has to remember to attach the right telemetry, or it silently lands in `SOFT_FAIL + INTERNAL_ERROR` by default — a category that's almost never what the caller meant.
+
+The win is small and concrete in the pragmatic version. The structural fix in Option E is bigger but eliminates an entire class of drift permanently.
+
+## 2. Patterns observed
+
+| Pattern | Sites | Example call site |
+| --- | --- | --- |
+| **A** Happy-path raw shape: `{ content: [{type:'text', text}] }` | ~8 handlers | `get_dataset.ts:59`, `get_key_value_store.ts:48`, `dataset_collection.ts:69`, `key_value_store_collection.ts:71`, `get_key_value_store_keys.ts:65` |
+| **B** Happy path + `structuredContent` | 2 handlers | `get_dataset_items.ts:135`, `get_actor_output.ts:120` |
+| **C** Informational (non-error, no data), e.g. `"Dataset 'X' is empty."` | 2 sites | `get_dataset_schema.ts:78`, `buildSearchActorsEmptyResponse` |
+| **D** Soft-fail user error via wrapper (`buildStorageNotFound`, `buildActorNotFoundResponse`) | 2 helpers, ~6 sites | `get_dataset.ts:49`, `fetch_actor_details_common.ts:142` |
+| **E** Inline `buildMCPResponse({ isError: true, telemetry: {...} })` | ~12 sites | `get_dataset_schema.ts:89`, `get_actor_output.ts:85`, `call_actor_common.ts:143`, `fetch_apify_docs.ts:94` |
+| **F** Inline `buildMCPResponse({ isError: true })` **without** telemetry | ~4 sites | `buildPermissionApprovalResponse` (call_actor_common.ts:124), `server.ts:1150, 1221, 1333` |
+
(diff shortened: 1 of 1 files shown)


MQ37 (approved):
Yes, the telemetry is quite complicated, great initiative! I would also go with the option F - I like the simplicity.

jirispilka:
Closing — decision made: **Option F**. Tracked in #935 with three independent sub-issues (#936 P1 fix, #937 `respond*` contract, #938 the compile-time lock).

This analysis stays here for reference; no markdown is merged into the repo (easier to keep the plan current in the issue).

---

PR #932, opened 2026-05-29
feat: Surface dataset size hints to guide large-output fetches

### LINKED ISSUE #878: feat: Hint dataset size to prevent get-dataset-items context blowup
## Problem
`get-dataset-items` serializes items unconditionally into both `content.text` and `structuredContent.items`. On real datasets this blows the context window. Item count is not a proxy for size — measured datasets run ~27 KB/item, so even 2 items exceed 50000 chars.

The original proposal (mirror `get-actor-output`'s `slice(TOOL_MAX_OUTPUT_CHARS)`) is rejected:
- `get-actor-output` is deprecated/being removed — don't spread its code.
- Its truncation is half-broken: it slices only `content.text` (invalid JSON) and leaves `structuredContent.items` full, so the payload isn't actually bounded.
- A hard cap silently drops data and can't serve a large-context client that legitimately wants everything.

## Decision: prevent, don't truncate
No size cap, no truncation, no new input param. Instead, tell the model how big the data is **before** it fetches, so it sets `limit` / `fields` itself. This respects large-context clients and keeps the model from having to reason about byte budgets.

### Key fact (verified against the live API)
`dataset().get()` returns `stats.inflatedBytes` — the uncompressed JSON size, i.e. exactly what lands in context. It's present on every dataset (0 when empty) and **free**: `get-actor-run` and `get-dataset` already call `dataset().get()`. Use `inflatedBytes`, not `s3StorageBytes` (compressed, understates ~4×).

Caveat: `inflatedBytes` is returned at runtime but **not declared** in the `apify-client` `DatasetStats` type (which only lists `readCount/writeCount/deleteCount/storageBytes?`). Read it defensively and treat missing as "no hint":
```ts
const bytes = (dataset.stats as { inflatedBytes?: number })?.inflatedBytes;
```
This is also why we keep it advisory only — never gate behavior on an untyped field.

## Changes (three tools, prevention only)

1. **`get-actor-run`** — extend the existing summary/nextStep (`actor_run_response.ts:378-379`, already reports item/field counts) with size from the `datasetInfo` it already fetches:
   > SUCCEEDED. 200 items, ~1.2 MB (~6 KB/item); 12 fields. Use get-dataset-items with limit=20; fetching all may be large — project with fields= or page with offset.
   Surface the byte size in the `RunDataset` structured output alongside `itemCount`.

2. **`get-dataset`** — already dumps the full `stats` object, so `inflatedBytes` is already in the output but unexplained. Add to the description that `stats.inflatedBytes` is the JSON payload size to use when choosing `limit`/`fields`.

3. **`get-dataset-items`** — description + self-sourced hint, **no cap**. It doesn't call `.get()`, but knows `total` and can measure the bytes it just serialized → extrapolate:
   > Returned 20 of 200 items (~24 KB). Full dataset ≈ ~240 KB — narrow with fields= or page with offset.
   Add a description line: results can be large; prefer `fields=` and a small `limit` first.

## Explicitly out of scope
- Hard `TOOL_MAX_OUTPUT_CHARS` cap / truncation in `get-dataset-items`.
- New input params (`maxOutputChars`, `includeAllData`, `capResponseSize`). Pagination (`offset`) is the "get everything" path; the hints make the model use it.

## Accepted tradeoff
Hints are advisory. An agent that ignores them and sets a huge `limit` can still overflow. Accepted: this preserves large-context use cases, and a hard cap on an untyped byte field would be fragile.

Parent: #784

jirispilka:
Closes #878

Dataset items can be huge, so fetching them blows the LLM's context window. This surfaces a dataset's uncompressed byte size (`stats.inflatedBytes`) on the **pre-fetch** tools so the model can narrow before items land in context. Advisory only — nothing is truncated.

- **`get-actor-run` / `get-dataset`** — append a "fetching all may exceed context — narrow with `fields=`/`offset`" steer to `nextStep` when the dataset is large.
- **`dataset-collection`** — `inflatedBytes` flows through per dataset; the description points the model at it.
- **`get-dataset-items`** — no hint: items and the hint arrive together, so the data is already in context.

### Apify core to ship
`client.dataset(id).get()` returns `storageBytes` (compressed), **not** `inflatedBytes` — only the dataset-**list** endpoint returns `inflatedBytes` today. So the `get-actor-run`/`get-dataset` hints read it defensively and stay **dormant** until the API/client expose `inflatedBytes` (and `storageBytes`) consistently on both endpoints (and drop internal `s3StorageBytes`). They light up with no further MCP change. `dataset-collection` works now (list endpoint).

https://claude.ai/code/session_013sKA6tQ9Y8Ytr859xmrBNv

### DIFF
--- a/src/tools/core/actor_run_response.ts
+++ b/src/tools/core/actor_run_response.ts
@@ -3,7 +3,13 @@ import type { ActorRun, Dataset, KeyValueClientListKeysResult } from 'apify-clie
 import log from '@apify/log';
 
 import type { ApifyClient } from '../../apify_client.js';
-import { FAILURE_CATEGORY, HelperTools, TOOL_STATUS } from '../../const.js';
+import {
+    DATASET_SIZE_HINT_BYTES,
+    FAILURE_CATEGORY,
+    HelperTools,
+    NARROW_OUTPUT_HINT,
+    TOOL_STATUS,
+} from '../../const.js';
 import { getWidgetConfig, WIDGET_URIS } from '../../resources/widgets.js';
 import type { ConsoleLinkContext } from '../../types.js';
 import {
@@ -78,6 +84,12 @@ export type RunDataset = {
     title?: string;
     itemCount?: number;
     cleanItemCount?: number;
+    /**
+     * Uncompressed size of the dataset in bytes (Apify `stats.inflatedBytes`). Items are stored as BSON,
+     * so this approximates — not exactly equals — the JSON size fetched into context. The single-dataset
+     * GET response does not return it (only the dataset-list endpoint does), so it is normally absent.
+     */
+    inflatedBytes?: number;
     /**
      * Dot-notation field paths. Pure-numeric segments (array indices) are stripped and the
      * list is deduped at build time, so callers receive a flat unique projection-valid list
@@ -214,6 +226,9 @@ function buildRunDataset(
         title: datasetMeta.title,
         itemCount: resolvedItemCount ?? datasetMeta.itemCount,
         cleanItemCount: datasetMeta.cleanItemCount,
+        // Undeclared on the apify-client `DatasetStats` type (and absent from the GET response today),
+        // so read it defensively; `cleanEmptyProperties` drops it when absent.
+        inflatedBytes: (datasetMeta.stats as { inflatedBytes?: number } | undefined)?.inflatedBytes,
         fields: datasetMeta.fields ? normalizeDatasetFields(datasetMeta.fields) : undefined,
     }) as RunDataset;
 }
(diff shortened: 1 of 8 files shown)


src/tools/core/actor_run_response.ts
```diff
@@ -395,6 +410,16 @@ function fieldsProjectionHint(fields: string[] | undefined): string {
     return ` Available fields (dot notation): ${fields.join(', ')} — pass via fields="..." to project.`;
 }
 
+/**
+ * nextStep steer away from one-shot fetching a large dataset. Only appended when the known byte
+ * size exceeds DATASET_SIZE_HINT_BYTES, so small datasets get no extra noise. `inflatedBytes` is the
+ * whole-dataset uncompressed size; shared by get-actor-run and get-dataset.
+ */
+export function datasetSizeNextStepHint(inflatedBytes: number | undefined): string {
+    if (inflatedBytes === undefined || inflatedBytes <= DATASET_SIZE_HINT_BYTES) return '';
```
MQ37: change: I would return this message always, why returning the size of the dataset only in specific scenarios? I think the agent should always know how large is the dataset - we can gatekeep the instructions but let's keep the size info.

MQ37 (approved):
Nice! Pre-approving but I would always return the dataset size as it does not make sense to me to keep this information hidden in some cases - if we know the size we should return it always. We can condition the instructions and hide then when `< DATASET_SIZE_HINT_BYTES`.

src/tools/core/actor_run_response.ts
jirispilka: Make sense, changed.

Issue #943, opened 2026-06-01
Experiment: does TOON degrade MCP task performance?


jirispilka:
## Problem
TOON cuts bytes, but LLMs are trained on JSON. Switching `text` to TOON could lower task-success rates. This is the go/no-go gate for #913's TOON PR.

## Proposed solution
- Run the existing workflow eval harness (`pnpm run evals:workflow`, `evals/workflows/results.json`) on master vs a TOON branch.
- Gate: no task-success regression. Record byte savings on the same runs.
- Capture results for the blog post.

Part of #941. Blocks #913 PR 2.

RobertCrupa:
After some experimenting, it appears TOON does not degrade performance, however, it does not greatly improve our token efficiency. After my evaluation it looks like certain tools will have their tokens increased and others will have a decrease.

Overall there's a small improvement from implementing TOON however it looks like it will be more of a hassle to maintain and keep than we are benefiting from.

Issue #913, opened 2026-05-26
 TOON format adoption for MCP tool responses


MQ37:
> **Spec — pending team review. No production code.**
> Filed as an issue (not a versioned doc in `res/`) so the conversation lives where decisions are made.
> Companion to Code Mode (#794): TOON shrinks results that flow through the LLM context; Code Mode skips most of them entirely.

## Worth your attention

- **Measured savings**: −6.2% bytes combined on 18 real Apify-API fixtures; the four list endpoints save 19–44% individually. `dataset-items` averages near zero — savings concentrate on list endpoints.
- **Stacks with `fields=`** (separate proposal): 100-place Google-Maps result projected to 3 fields → 1.01 MB → 9.5 KB, ~110× smaller into LLM context.
- **`structuredContent` becomes the JSON fallback.** Programmatic consumers always see pure JSON via `structuredContent`. Only `TextContent.text` shifts. The picker may swap JSON for TOON in `text`; consumers that need JSON parse `structuredContent`.
- **Bounded by construction**: `json` always in candidate set; defensive `try`/`catch` around the TOON candidate; `MAX_DEPTH = 20` guard with margin of 11 over deepest real fixture; dotted-key normalisation with collision guard.
- **LLM behaviour change unmeasured.** The spec calls for an `evals/workflows/` regression gate before any production merge.

---

## Problem

MCP tool results land in the LLM context window. Most tools wire JSON-encoded payloads (Actor runs, dataset listings, KV-store keys, dataset items). JSON repeats every object key on every row of an array. Context costs tokens and dilutes model attention, and once a tool result lands it stays in the conversation for every subsequent turn — the savings compound per request.

## TL;DR

Switching the text content of certain MCP tool responses from JSON to [TOON](https://toonformat.dev/) (a more compact LLM-oriented format, introduced below) measurably reduces context bytes — and stacks massively with field projection on realistic agentic workloads.

| Tool                              | Δ vs current JSON | Strategy that wins most often  |
|-----------------------------------|------------------:|--------------------------------|
| `get-actor-run-list`              |            −44.0% | `toon-flatten`                 |
| `get-dataset-list`                |            −31.6% | `toon-flatten` (mixed-schema: `json`) |
| `get-key-value-store-list`        |            −31.3% | `toon-flatten` (mixed-schema: `json`) |
| `get-key-value-store-keys`        |            −19.0% | `toon-flatten`                 |
[...]

PR #950, opened 2026-06-03
feat(tools): TOON-encode list responses with structuredContent + outputSchema

MQ37:
## What
The five array endpoints (`get-actor-run-list`, `get-dataset-list`, `get-key-value-store-list`, `get-key-value-store-keys`, `get-dataset-items`) now ship:
- `structuredContent` + an `outputSchema` (pure JSON, mirroring the `apify-client` / api.apify.com response shapes), and
- dot-flattened [TOON](https://toonformat.dev/) in `content[].text`.

`encode_text.ts` exposes two encoders: `wrapJsonText` (always JSON, single-object tools) and `encodeToon` (always TOON, list tools). `encodeToon` falls back to a JSON fence only when `dotFlatten`/`encode` throws (depth overflow or a normalisation key collision) — a lossless, logged error path, not a byte-driven picker.

This branch combines PR 1 (structuredContent/outputSchema) with the now-merged #952 (always-TOON encoding) into one change. Implements #913.

## Why
JSON repeats every object key on every array row; TOON's tabular form drops that, shrinking the list payloads that sit in the LLM context. `structuredContent` is the JSON source of truth for programmatic consumers and `outputSchema` validators, independent of how `text` is encoded — so the schema work is the additive base the TOON encoding builds on. One stable shape per tool beats a format that flips on byte count: easier to reason about and debug.

## Worth your attention
- **`content[].text` is TOON (or JSON on the rare error fallback).** All five tools declare `outputSchema`/`structuredContent`, authoritative for machine consumers. A client that blindly `JSON.parse`s `content[0].text` and ignores `structuredContent` would need to handle TOON. This knowingly deviates from the spec SHOULD ("a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block").
- **`encodeToon` serialises through `JSON.stringify` first**, so `dotFlatten` works on the JSON data model only — a `Date`/`toJSON` carrier becomes its ISO string, not an empty object (pinned by a regression test).
- **JSON fallback is an error path, not a picker.** A top-level key named `constructor`/`toString`/`__proto__`/… trips the `key in out` collision guard; TOON is dropped and JSON ships (lossless, logged). The guard also blocks `__proto__` pollution.
- Schemas validated against the **live api.apify.com**, which caught nullability gaps the SDK `.d.ts` hides — `name`, `title`, `actId`, `actRunId`, `exclusiveStartKey`, `nextExclusiveStartKey` are returned as `null` and are now typed `['string','null']`.
- `@toon-format/toon@2.3.0` — pinned, 0 deps.

## Testing
- Rebased on latest master; `type-check` and `lint` clean; unit suite green (855 passed, 2 skipped). Integration tests require `APIFY_TOKEN` and were not run.
- Live verification against a local build via [`mcpc`](https://github.com/apify/mcpc); TOON output decoded back with the real `@toon-format/toon` decoder to confirm losslessness.

## Before merge
Gated on the TOON task-performance experiment (#943) — `pnpm run evals:workflow`, no regression vs master. The byte win is proven; the LLM-comprehension win is not yet.

### DIFF
--- a/tests/unit/tools.run_collection.test.ts
+++ b/tests/unit/tools.run_collection.test.ts
@@ -1,0 +1,77 @@
+import { describe, expect, it, vi } from 'vitest';
+
+import { HelperTools } from '../../src/const.js';
+import { getUserRunsList } from '../../src/tools/common/run_collection.js';
+import type { HelperTool, InternalToolArgs } from '../../src/types.js';
+import { parseFencedJson, stubToolCallContext, type TextToolResult } from './helpers/tool_context.js';
+
+const listMock = vi.fn();
+
+vi.mock('../../src/apify_client.js', () => ({
+    ApifyClient: vi.fn().mockImplementation(function () {
+        return { runs: () => ({ list: listMock }) };
+    }),
+}));
+
+const MOCK_RUNS = {
+    total: 1,
+    offset: 0,
+    limit: 10,
+    desc: false,
+    count: 1,
+    items: [
+        {
+            id: 'run-1',
+            actId: 'act-1',
+            status: 'SUCCEEDED',
+            startedAt: '2026-05-12T09:18:27.527Z',
+            finishedAt: '2026-05-12T09:19:01.000Z',
+            defaultDatasetId: 'ds-1',
+            defaultKeyValueStoreId: 'kv-1',
+        },
+    ],
+};
+
+// run_collection constructs its own ApifyClient from the token, so the injected client is unused.
+const noClient = null as unknown as InternalToolArgs['apifyClient'];
+
+describe('get-actor-run-list', () => {
+    it('has the expected tool name', () => {
+        expect(getUserRunsList.name).toBe(HelperTools.ACTOR_RUN_LIST_GET);
+    });
+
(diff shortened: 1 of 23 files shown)


tests/unit/tools.run_collection.test.ts
```diff
@@ -0,0 +1,79 @@
...
+import { HelperTools } from '../../src/const.js';
+import { getUserRunsList } from '../../src/tools/common/run_collection.js';
+import type { HelperTool, InternalToolArgs } from '../../src/types.js';
+import { parseFencedJson, stubToolCallContext, type TextToolResult } from './helpers/tool_context.js';
+
+const listMock = vi.fn();
+
+vi.mock('../../src/apify_client.js', () => ({
+    // Vitest 4 constructs mocked classes via `Reflect.construct`, which needs a real
+    // constructable function (an arrow has no [[Construct]]).
```
RobertCrupa: Nit: I think that's an unnecessary comment

RobertCrupa (approved):
Nice one 🚀

jirispilka (approved):
LGTM 🚀

MQ37:
The evals are passing, merging.

@jirispilka @RobertCrupa

PR #896, opened 2026-05-21
fix(ci): repair mcpb package build under pnpm 11

MQ37:
## Context
The repository migrated from npm to pnpm, causing the `pnpm deploy --prod mcpb` step in the stable release GitHub workflow to fail with `[ERR_PNPM_NOTHING_TO_DEPLOY] No project was selected for deployment`.

## Solution
Specify the target workspace filter and deploy configurations under pnpm 11:
- Add `--filter=@apify/actors-mcp-server` to select the root project for deployment.
- Add `--legacy` to bypass the missing `inject-workspace-packages` setting in workspaces.
- Add `--node-linker=hoisted` to build a flat/hoisted `node_modules` without `.pnpm/` store or symlinks (as required for the MCPB packer).

## Worth your attention
- `--node-linker=hoisted` is critical to match the flat structure expected by the MCPB packer.

PR #951, opened 2026-06-03
ci(release): smoke-test packed .mcpb and drop redundant clean step

### LINKED ISSUE #862: v0.10.4: ERR_MODULE_NOT_FOUND @opentelemetry/core in release MCPB asset (peerDep not hoisted)
## Summary

The prebuilt `apify-mcp-server.mcpb` asset attached to release [v0.10.4](https://github.com/apify/apify-mcp-server/releases/tag/v0.10.4) fails to start with `ERR_MODULE_NOT_FOUND: @opentelemetry/core`. The bundle's top-level `node_modules/@opentelemetry/` is missing the `core` subdirectory that `@sentry/opentelemetry` (version 10.50.0 in this build) requires as a peer dependency. A clean `npm install --omit=dev` against the v0.10.4 `package.json` does produce a top-level `node_modules/@opentelemetry/core/`, so the dependency declaration is fine — the loss appears to happen in the packaging step that produces the `.mcpb`.

## Environment

- Asset: `apify-mcp-server.mcpb` (18,608,641 bytes) from release `v0.10.4`
- Node: v22.21.1
- OS: macOS (also reproducible on Linux — no native code involved)

## Reproduction (released asset)

```bash
# 1. Download the release asset
gh release download v0.10.4 \
  --repo apify/apify-mcp-server \
  --pattern apify-mcp-server.mcpb

# 2. Extract (.mcpb is a zip)
unzip -q apify-mcp-server.mcpb -d extracted

# 3. Confirm @opentelemetry/core is missing at the top level
ls extracted/node_modules/@opentelemetry/
# Output lists: api, instrumentation-*, resources, sdk-trace-base,
# semantic-conventions, sql-common — no "core/" subdirectory.

# 4. Confirm 13 nested copies exist but are unreachable from @sentry/opentelemetry
find extracted/node_modules -type d -path '*@opentelemetry/core' | wc -l   # → 13

# 5. Run the server's stdio entrypoint
cd extracted
node dist/stdio.js --tools actors,docs
```

### Error output

```
node:internal/modules/package_json_reader:314
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@opentelemetry/core' imported from
  /.../extracted/node_modules/@sentry/opentelemetry/build/esm/index.js
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:314:9)
    at packageResolve (node:internal/modules/esm/resolve:767:81)
    ...
  code: 'ERR_MODULE_NOT_FOUND'
}

Node.js v22.21.1
```

The server exits at `node dist/stdio.js` startup before any MCP message is exchanged.

## Cross-check (clean install from source at the same tag)

```bash
gh api 'repos/apify/apify-mcp-server/zipball/v0.10.4' > source.zip
unzip -q source.zip && cd apify-apify-mcp-server-*

npm install --no-audit --no-fund --ignore-scripts --omit=dev
# added 436 packages in 3s

ls node_modules/@opentelemetry/core/package.json
# → exists
jq .version node_modules/@opentelemetry/core/package.json
# → "1.30.1"
```

So a fresh `npm install --omit=dev` against the published `package.json` / `package-lock.json` **does** place `@opentelemetry/core` (version 1.30.1) at top-level, which would satisfy the peer-dep requested by `@sentry/opentelemetry` (`^1.30.1 || ^2.1.0`). The released `.mcpb` does not have this directory — so a step in the packaging pipeline is dropping it between install and pack.

## Root cause

`@sentry/opentelemetry` (version 10.50.0) declares `@opentelemetry/core` as a peer dependency:

```json
{
  "name": "@sentry/opentelemetry",
  "version": "10.50.0",
  "peerDependencies": {
    "@opentelemetry/api": "^1.9.0",
    "@opentelemetry/core": "^1.30.1 || ^2.1.0",
    "@opentelemetry/sdk-trace-base": "^1.30.1 || ^2.1.0",
    "@opentelemetry/semantic-conventions": "^1.39.0"
  }
}
```

In the released `.mcpb`, `@opentelemetry/api`, `@opentelemetry/sdk-trace-base`, and `@opentelemetry/semantic-conventions` are present at the top level. `@opentelemetry/core` is not. It appears only in nested locations, e.g.:

```
node_modules/@fastify/otel/node_modules/@opentelemetry/core
node_modules/@sentry/node/node_modules/@opentelemetry/core
node_modules/@opentelemetry/instrumentation-pg/node_modules/@opentelemetry/core
node_modules/@opentelemetry/instrumentation-http/node_modules/@opentelemetry/core
node_modules/@opentelemetry/sql-common/node_modules/@opentelemetry/core
node_modules/@opentelemetry/sdk-trace-base/node_modules/@opentelemetry/core
... (13 nested copies in total, all v2.7.0)
```

Node's ESM resolver walks up from `@sentry/opentelemetry`'s location and never sees a sibling/parent `@opentelemetry/core/package.json`, so resolution fails before the module graph is linked.

`@opentelemetry/core` lands in the install tree only because it satisfies `@sentry/opentelemetry`'s peer-dep contract — it isn't declared in `apify-mcp-server`'s own `dependencies`. Any post-install step that prunes top-level `node_modules` against the direct-dependency set (a "production prune") would remove this directory while leaving the deeper nested copies untouched — which matches exactly what the released `.mcpb` shows.

Looking at [`.github/workflows/manual_release_stable.yaml`](https://github.com/apify/apify-mcp-server/blob/v0.10.4/.github/workflows/manual_release_stable.yaml), the `build_mcpb` job copies `node_modules` into a working directory, then runs `pack` followed by `clean` from the `@anthropic-ai/mcpb` CLI (pinned to version 2.1.2). The `mcpb clean` step is the most likely place where the top-level directory is dropped — the released `.mcpb` looks like the output of a production-style prune against the declared `dependencies` (the missing module is the only one present purely as a peer-dep resolution).

## Why earlier releases were not affected

For context: the v0.9.17 bundle also ships without a top-level `node_modules/@opentelemetry/core/`, but the server starts cleanly. The reason is layout: v0.9.17's install resolves `@sentry/opentelemetry` nested under `@sentry/node/node_modules/`, and the `@sentry/node` package version used in that build declares `@opentelemetry/core` as a direct dependency, so a corresponding nested copy at `node_modules/@sentry/node/node_modules/@opentelemetry/core/` survives the prune and is reachable by Node's resolver from the nested `@sentry/opentelemetry`.

In v0.10.4 the install hoists `@sentry/opentelemetry` to top-level `node_modules/@sentry/opentelemetry/`. From that location the nested copy is no longer reachable, so the missing top-level `@opentelemetry/core` becomes fatal. Both versions rely on a peer-dep edge that isn't preserved by a `dependencies`-only prune; v0.9.17 happens to dodge it through the nested layout.

So this is best treated as a latent issue that v0.10.x's install layout exposed, not a regression in `@sentry/node` or `@sentry/opentelemetry` themselves.

## Possible fixes

1. **Add `@opentelemetry/core` as a direct `dependencies` entry** in `package.json`. This is the most defensive fix: the package becomes a first-class dependency and survives any post-install pruning step that filters top-level `node_modules` against the declared dependency set.
2. **Skip the `Clean MCPB package` step**, or replace it with a manual size-trim that preserves peer-dep-resolved top-level modules. Removing just the `mcpb clean` step from the release workflow and re-running against `v0.10.4` is a quick way to confirm whether that step is the source.
3. **Bundle with esbuild/rollup** so imports are inlined and Node's runtime module resolver is not consulted for these packages.

## Workaround in use

Pinning to v0.9.17 in our local consumption until a fixed release ships.

## Closing

Happy to test a candidate fix against the reproduction above and confirm the released asset starts cleanly. Thanks!

jirispilka:
## What

Adds a startup smoke test for the packed `.mcpb` to the stable release pipeline, and removes the now-redundant `mcpb clean` step.

- **`_smoke_test_mcpb.yaml`** (new, reusable `workflow_call`, mirrors `_smoke_test_npm_package.yaml`): downloads the uploaded `.mcpb` artifact, installs `mcpc`, extracts the bundle, and launches it over MCP — `connect → ping → tools-list → search-actors → fetch-actor-details → close`.
- **`manual_release_stable.yaml`**: wires in a `smoke_test_mcpb` job (`needs: build_mcpb`) and gates `create_github_release` on it, so a bundle that can't start never reaches the release. Drops the `Clean MCPB package` step.

## Why

#862: the v0.10.4 `.mcpb` failed at startup with `ERR_MODULE_NOT_FOUND: @opentelemetry/core`. That release's pipeline ran `mcpb clean` after pack, which production-pruned the top-level `node_modules` and dropped `@opentelemetry/core` (present only as `@sentry/opentelemetry`'s peer-dep resolution).

The underlying bug was **already fixed in v0.10.8** by the npm→pnpm packaging rewrite (#896) — `pnpm deploy --prod --node-linker=hoisted` keeps the peer-resolved module at top level. This PR is hardening, not the fix:

- The release pipeline built, validated, and packed the `.mcpb` but **never started it**, so an unstartable bundle could ship undetected. The smoke test closes that gap.
- `mcpb clean` is a no-op on the `pnpm deploy --prod` output (verified: byte-identical, zero files removed) and was the original culprit, so removing it eliminates that prune at zero size cost.

## Verification

- Reproduced the full pipeline locally (`pnpm deploy --prod` → `pack`): the packed `.mcpb` ships `@opentelemetry/core@2.7.1` at top level and `node dist/stdio.js` starts cleanly.
- Ran the new workflow's exact steps in a fresh-runner simulation (`npm install @apify/mcpc`, cwd-relative `connect`, `ping`, `tools-list`, a real tool round-trip via docs, `close`) — all pass.
- `mcpb clean` measured as a no-op on the deploy output (14,820,495 → 14,820,494 bytes, no file-list change).
- Note: `search-actors` / `fetch-actor-details` calls could not be exercised in the sandbox (host-specific egress throttle, not a general block) — they are the same calls the existing npm smoke test makes and that job passes in CI.

Closes #862.

https://claude.ai/code/session_01MSfvap2vxLGcbSGjVT1rYe

---
_Generated by [Claude Code](https://claude.ai/code/session_01MSfvap2vxLGcbSGjVT1rYe)_


.github/workflows/_smoke_test_mcpb.yaml
```diff
@@ -0,0 +1,49 @@
...
+        steps:
+            -   name: Download MCPB artifact
+                uses: actions/download-artifact@v4
+                with:
+                    name: ${{ inputs.artifact_name }}
+
+            -   name: Use Node.js
+                uses: actions/setup-node@v6
+                with:
+                    node-version: 22
```
MQ37:
```suggestion
                    node-version: 24
```

Please let's use the latest LTS

MQ37 (approved):
LGTM! Please update the node version to the latest LTS.

PR #952, opened 2026-06-03
feat(tools): always TOON-encode list responses

MQ37:
Stacked on #950 (base = `feat/list-tools-structured-content`). Implements #913.

## What

The five array endpoints (`get-actor-run-list`, `get-dataset-list`, `get-key-value-store-list`, `get-key-value-store-keys`, `get-dataset-items`) always ship dot-flattened [TOON](https://toonformat.dev/) in `content[].text`; `structuredContent` stays pure JSON. `encode_text.ts` exposes two encoders: `wrapJsonText` (always JSON, single-object tools) and `encodeToon` (always TOON, list tools). `encodeToon` falls back to a JSON fence only when `dotFlatten`/`encode` throws (depth overflow or a normalisation key collision) — a lossless error path, logged, not a byte-driven alternative.

## Why

JSON repeats every object key on every array row; TOON's tabular form drops that, shrinking the list payloads that sit in the LLM context. Per review (#952 discussion), one stable shape per tool beats a format that flips on byte count: easier to reason about and debug. `structuredContent` (from #950) remains the JSON source of truth for programmatic consumers and `outputSchema` validators.

## Worth your attention

- **`content[].text` is TOON (or JSON on the rare error fallback).** Spec-sound because all five tools declare `outputSchema`/`structuredContent`, authoritative for machine consumers. A client that blindly `JSON.parse`s `content[0].text` and ignores `structuredContent` would need to handle TOON — flagged so reviewers are aware. This knowingly deviates from the SHOULD in the spec ("a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block").
- **`encodeToon` serialises through `JSON.stringify` first**, so `dotFlatten` works on the JSON data model only — a `Date`/`toJSON` carrier becomes its ISO string, not an empty object (pinned by a regression test).
- **JSON fallback is an error path, not a picker.** A `get-dataset-items` payload with a top-level key named `constructor`/`toString`/`__proto__`/… trips the `key in out` collision guard; TOON is dropped and JSON ships (lossless, logged). `Object.create(null)` would remove the false positive; the guard also blocks `__proto__` pollution.
- `@toon-format/toon@2.3.0` — pinned, 0 deps.

## Testing

- Unit suite green (844 passed, 2 skipped); `tsc --noEmit` clean; `oxlint` 0/0. Tests cover the TOON round-trip, the `Date`/null/nested case, dot-flatten collision + depth fallback to JSON, and `content == structuredContent`.
- Live verification against a local build via [`mcpc`](https://github.com/apify/mcpc) (real Apify runs); TOON output decoded back with the real `@toon-format/toon` decoder to confirm losslessness.

## Before merge

Gated on the TOON task-performance experiment (#943) — `pnpm run evals:workflow`, no regression vs master. The byte win is proven; the LLM-comprehension win is not yet.

### DIFF
--- a/src/utils/encode_text.ts
+++ b/src/utils/encode_text.ts
@@ -75,29 +75,24 @@ export function wrapJsonText(value: unknown): string {
 }
 
 /**
- * Encodes a JSON-serialisable value as fenced text for the LLM, shipping whichever of the JSON or
- * dot-flattened-TOON encodings is smaller (UTF-8 bytes). JSON is always a candidate, so the result
- * is never larger than the plain JSON fence; the TOON candidate is dropped if `dotFlatten`/`encode`
- * throws.
+ * Encodes a JSON-serialisable value as a dot-flattened ```toon code fence for the LLM, falling back
+ * to a ```json fence only if `dotFlatten`/`encode` throws (depth overflow or a normalisation key
+ * collision). TOON is the stable shape; JSON is the error path, not a byte-driven alternative.
  *
  * The text may be TOON — programmatic JSON consumers must read `structuredContent` instead. Only
  * wire this on tools that ship `structuredContent` as the JSON fallback.
  */
-export function encodeSmallest(value: unknown): string {
-    // Single source of truth: both candidates derive from this one serialisation, so the TOON
-    // candidate can never encode different data than JSON. A `Date` (or any `toJSON` carrier)
-    // becomes its ISO string here, not an empty object — `dotFlatten` works on the JSON data
-    // model only. A non-serialisable value (circular, BigInt) throws here; API payloads are
-    // always JSON, so let it crash.
+export function encodeToon(value: unknown): string {
+    // `value` is serialised through JSON so `dotFlatten` works on the JSON data model only: a `Date`
+    // (or any `toJSON` carrier) becomes its ISO string here, not an empty object. A non-serialisable
+    // value (circular, BigInt) throws here; API payloads are always JSON, so let it crash.
     const jsonStr = JSON.stringify(value);
-    const json = fence('json', jsonStr);
-    let toon: string | undefined;
     try {
-        toon = fence('toon', encode(dotFlatten(JSON.parse(jsonStr))));
+        return fence('toon', encode(dotFlatten(JSON.parse(jsonStr))));
     } catch (err) {
-        // dotFlatten threw (depth overflow or key collision) or encode failed — JSON ships.
-        // Log it: the fallback is correct, but a recurring failure means TOON is silently disabled.
-        log.warning('encodeSmallest: TOON candidate dropped, shipping JSON', { err });
+        // dotFlatten threw (depth overflow or key collision) or encode failed — ship JSON instead.
+        // Log it: the fallback is lossless, but a recurring failure means TOON is silently disabled.
+        log.warning('encodeToon: TOON encoding failed, shipping JSON', { err });
+        return fence('json', jsonStr);
     }
-    return toon !== undefined && Buffer.byteLength(toon, 'utf8') < Buffer.byteLength(json, 'utf8') ? toon : json;
 }
--- a/src/utils/encode_text.ts
+++ b/src/utils/encode_text.ts
@@ -1,0 +1,72 @@
+import { encode } from '@toon-format/toon';
+
+import { JSON_FENCE_PREFIX, JSON_FENCE_SUFFIX } from './mcp.js';
+
+/**
+ * Recursion guard for {@link dotFlatten}. The deepest real Apify-API fixture measured is
+ * depth 9 (a user-declared dataset schema), so 20 leaves a comfortable margin while still
+ * stopping runaway recursion on pathological input.
+ */
+const MAX_DEPTH = 20;
+
+const TOON_FENCE_PREFIX = '```toon\n';
+const TOON_FENCE_SUFFIX = '\n```';
+
+function flattenValue(value: unknown, depth: number): unknown {
+    if (depth > MAX_DEPTH) throw new RangeError('dotFlatten: max depth exceeded');
+    if (Array.isArray(value)) return value.map((item) => flattenValue(item, depth + 1));
+    if (value !== null && typeof value === 'object') {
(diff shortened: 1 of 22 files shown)


RobertCrupa (approved):

src/utils/encode_text.ts
```diff
@@ -0,0 +1,103 @@
...
+/** Wrap an already-encoded body in the Markdown code fence for its format. */
+function fence(format: keyof typeof FENCES, body: string): string {
+    return `${FENCES[format].prefix}${body}${FENCES[format].suffix}`;
+}
+
+function flattenValue(value: unknown, depth: number): unknown {
+    if (depth > MAX_DEPTH) throw new RangeError('dotFlatten: max depth exceeded');
+    if (Array.isArray(value)) return value.map((item) => flattenValue(item, depth + 1));
+    if (value !== null && typeof value === 'object') {
+        const out: Record<string, unknown> = {};
```
jirispilka:
```suggestion
        const out: Record<string, unknown> = Object.create(null);
```

Edge case,  but it's a one-line change that deletes a 7-line cautionary comment below it

Use a null-prototype accumulator — it deletes the 7-line comment and fixes a real misfire.

// src/utils/encode_text.ts:27
const out: Record<string, unknown> = Object.create(null);
What triggers the current bug. dotFlatten throws a false-positive collision whenever an object — a dataset-item row or the top-level response — has a direct key that's an Object.prototype name (constructor, toString, valueOf, hasOwnProperty, __proto__, …) with a scalar/array value, because key in out is true on a plain {} before anything is assigned. encodeSmallest catches the throw → TOON dropped, JSON ships, log.warning fires. Nested occurrences (metadata.__proto__) get a dotted prefix and slip past, so the bite is exactly at the top level — where dataset-item columns live.

Real Apify payloads that hit it:

get-dataset-items from a custom page function returning a field literally named constructor — a normal domain term (construction/contractor data, programming/reference content). {role, constructor} → current code throws, fix builds TOON.
Verbatim capture of wild JSON (JSON-LD, __NEXT_DATA__, GraphQL mirrors) where __proto__/constructor are literal keys. Note JSON.parse('{"id":1,"__proto__":"x"}') creates a real own __proto__ key (Object.keys → ["id","__proto__"]), which the current accumulator throws on.
The cruel part: these are often uniform tabular rows — the payloads TOON compresses best — so the guard misfires on the highest-value case.

It also removes a fragile coupling. The comment admits the only thing preventing __proto__ prototype-pollution is that "the guard fires before assignment" — i.e. pollution-safety depends on the collision check happening to run first. With Object.create(null) it's robust by construction:

'constructor' in out → false → no false positives.
out['__proto__'] = x → ordinary own key, no setter in the chain → lossless, no pollution, independent of ordering.
key in out then means exactly Object.hasOwn, so only genuine normalisation collisions (a.b vs a_b) remain.

src/utils/encode_text.ts
```diff
@@ -0,0 +1,103 @@
...
+    // always JSON, so let it crash.
+    const jsonStr = JSON.stringify(value);
+    const json = fence('json', jsonStr);
+    let toon: string | undefined;
+    try {
+        toon = fence('toon', encode(dotFlatten(JSON.parse(jsonStr))));
+    } catch (err) {
+        // dotFlatten threw (depth overflow or key collision) or encode failed — JSON ships.
+        // Log it: the fallback is correct, but a recurring failure means TOON is silently disabled.
+        log.warning('encodeSmallest: TOON candidate dropped, shipping JSON', { err });
```
jirispilka: Do we want to see warning for weird / wrong data? I would go with info / debug

src/utils/encode_text.ts
```diff
@@ -0,0 +1,103 @@
...
+        const out: Record<string, unknown> = {};
+        flattenInto(value as Record<string, unknown>, '', depth, out);
+        return out;
+    }
+    return value; // scalar or null — unchanged
+}
+
+function flattenInto(obj: Record<string, unknown>, prefix: string, depth: number, out: Record<string, unknown>): void {
+    if (depth > MAX_DEPTH) throw new RangeError('dotFlatten: max depth exceeded');
+    for (const [rawKey, v] of Object.entries(obj)) {
```
jirispilka, line 39: nit: In `flattenValue` you are using `value` and here you use `v` for the same thing

jirispilka (commented):
Overall it looks good.
We will deviate from a spec a little bit:

> For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block

But I think it is worth the tokens :) 

One think that I would like to reconsider is the adaptive "smallest of JSON/TOON" picker, please pick one format (or per tool format if required).

My concern is consistency. The picker makes the response format depend on the data, not the tool: the same tool returns JSON on small payloads and TOON on large ones — and with TOON the data shape also changes (dot-flattened metadata.httpStatus vs nested metadata.httpStatus). That costs us:

- It is non-deterministic (I know LLM are non-determinisitc) but it can make it hard to debug
- One stable shape per tool is easier to reason about than a format that flips on byte count.

Claude is suggesting ( I don't really agree with the second point):
- The four uniform-schema list tools (*-list, kv-keys) → always TOON (reliable win, reliable round-trip).
- get-dataset-items (arbitrary nested data) → always JSON (TOON ties/loses anyway, per the table).

That drops the picker, the byte-compare, and likely the limit=250 cap, and gives every client and the model one predictable shape per tool.

src/utils/encode_text.ts
MQ37: Done in 0b6b0aa. Switched the accumulator to `Object.create(null)` and deleted the cautionary comment.

`key in out` now tests own keys only, so a top-level `constructor`/`__proto__`/`toString`/… column no longer trips a false-positive collision, and `__proto__` (own key from parsed JSON) can't reach the prototype setter — robust by construction, not by guard ordering. Only a genuine normalisation clash (`a.b` vs `a_b`) still throws.

Added two regression tests: a prototype-name column now ships TOON (round-trips lossless) instead of falling back to JSON, and an own `__proto__` key survives as data with the prototype left `null`.

PR #958, opened 2026-06-04
feat: return personalized Console links for Console UI token sessions

vojtechj-apify:
## Summary

Actor, **run, dataset, and key-value-store** links in tool outputs are minted as **personalized Apify Console links** when the session is authenticated with a Console UI token (`apify_ui_...`). All other callers are byte-identical to today: Actors keep public `apify.com` links; runs/storages stay link-free.

Relates to apify/apify-core#27286 (AI Chat — use console.apify.com links in chat, not apify.com). Counterpart apify-core branch: `feat/ai-chat/27286-token-replacement` (chat authenticates against MCP with the Console session's UI token; its system prompt instructs presenting Actor/run/storage links from the exact tool URLs).

## Link policy

UI tokens are issued only to Console sessions, so they are a verifiable signal that the caller is in Console. Personal vs org is resolved from the cached `users/me` lookup (org-scoped tokens add an `/organization/<orgId>` prefix).

| Resource | No token / API token (non-Console) | Console UI token (personal) | Console UI token (org-scoped) |
|---|---|---|---|
| Actor | `apify.com/<user>/<name>` (unchanged) | `console.apify.com/actors/<id>` | `…/organization/<orgId>/actors/<id>` |
| Run | no link (unchanged) | `…/actors/runs/<id>` | `…/organization/<orgId>/actors/runs/<id>` |
| Dataset | no link (unchanged) | `…/storage/datasets/<id>` | `…/organization/<orgId>/storage/datasets/<id>` |
| Key-value store | no link (unchanged) | `…/storage/key-value-stores/<id>` | `…/organization/<orgId>/storage/key-value-stores/<id>` |

**Tools affected:** `search-actors`, `fetch-actor-details` (Actor links); `call-actor`, `get-actor-run`, direct actor tools (run + default dataset/KV); `get-dataset`, `get-dataset-items`, `get-key-value-store`, `get-key-value-store-keys`, `get-key-value-store-record` (storage links).

**Widget tools are out of scope here**: `search-actors-widget` / `fetch-actor-details-widget` (and the run widget's Actor link) still emit public `apify.com` links for all sessions. We need to double-check whether Console uses them before wiring links through.

## How

- **Gate**: `isConsoleUiToken()` on the `apify_ui_` prefix. `getConsoleLinkContext()` short-circuits to `undefined` for non-UI tokens (no API call); for UI tokens it resolves the acting account from the already-cached `users/me` lookup (which now also captures `organizationOwnerUserId` → `CachedUserInfo.isOrganization`). Zero extra API calls.
- **URL builders**: `buildConsoleActorUrl` plus `buildConsoleRunUrl` / `buildConsoleDatasetUrl` / `buildConsoleKeyValueStoreUrl`, all sharing one base+org-prefix helper.
- **Mint at source** (no post-hoc rewriting):
  - Actor links — `extractActorData` (`actor_card.ts`) and `fetch_actor_details_common.ts` (markdown card, structured `url`, input-schema heading; `/pricing` and `/input` sub-links collapse to the detail page — no Console equivalents).
  - Run + storage links — central `applyConsoleUrls` in `actor_run_response.ts` (run + default dataset/KV); storage getters via shared `buildConsoleLinkContent`.
- **Placement**: `consoleUrl` fields in `structuredContent` (run + dataset-items output schemas extended) plus a `Console: run … | dataset … | key-value store …` line in the run narrative and a `Console: <url>` line on storage tools.
- **Verbatim-links nudge** (Console sessions only): instructs the model to present the returned URLs verbatim — models otherwise "correct" Console links to the public `apify.com` URLs from training data.
- **Config**: `CONSOLE_BASE_URL` defaults to `https://console.apify.com`; env-overridable for local dev/staging. No prompt/instruction text contains any domain.
- Deliberately untouched: developer profile links (no Console equivalent), docs links, README passthrough content.

## Testing

- **Unit**: `tests/unit/console_link.test.ts` (builders + gate) plus per-tool gating cases across `search-actors`, `fetch-actor-details`, `get-actor-run`, and all storage tools. Full suite **855 passed**, lint + type-check clean.
- **E2E on a local platform** (mcpc against the public dev server, plus the live Console chat UI):
  - Personal & org UI tokens → Console links for Actor/run/dataset/KV (org-prefixed under an org session).
  - Personal & org API tokens → **no links** (confirms the gate is the UI-token prefix, not account type).
  - Unauthenticated → rejected.
  - Default/non-UI output is byte-identical — `consoleUrl` is absent from `structuredContent`, no extra content item, no nudge.

## Notes for reviewers

- Known pre-existing edge (unchanged by this PR): a non-Console client fetching a non-public Actor still gets a public `apify.com` link that 404s. Fixing that would need a per-resource `isPublic` check on top of the per-session gate — possible follow-up.


vojtechj-apify:
NOTE - the PR seems long but approximately 600 lines of changes are just addded tests.

src/tools/common/storage_helpers.ts
```diff
@@ -1,7 +1,17 @@
...
 import { QUOTE_WRAPPER_CHARS } from '../../utils/generic.js';
 import { buildMCPResponse } from '../../utils/mcp.js';
 
+/**
+ * Optional extra text content item carrying the storage's personalized Console link
+ * (Console UI token sessions only). Spread into the tool's `content` array.
+ */
+export function buildConsoleLinkContent(consoleUrl: string | undefined): { type: 'text'; text: string }[] {
+    if (!consoleUrl) return [];
+    return [{ type: 'text', text: `Console: ${consoleUrl}\n${VERBATIM_LINKS_NUDGE}` }];
```
MQ37: nit: I do not like returning just the "Console" string but do not have honestly better string to return. What about returning "Apify Console" so it is clear it is Apify console?

src/tools/core/actor_run_response.ts
```diff
@@ -65,6 +72,8 @@ async function raceAbort<T>(promise: Promise<T>, abortSignal: AbortSignal | unde
 
 export type RunDataset = {
     id: string;
+    /** Personalized Apify Console link; set only for Console UI token sessions. */
+    consoleUrl?: string;
```
MQ37: nit: I do not like the console name as some users/agents may not know what is Apify Console but I do not have any better name. I would just be more descriptive and change all the `consoleUrl` to `apifyConsoleUrl` so it is clear we refer to the Apify console. Whad do you think? @jirispilka

src/tools/core/actor_run_response.ts
```diff
@@ -229,6 +242,30 @@ function errMessage(error: unknown): string {
...
+ */
+export function consoleLinksText(response: RunResponse): string {
+    if (!response.consoleUrl) return '';
+    const parts = [`run ${response.consoleUrl}`];
+    const datasetUrl = response.storages.datasets?.default.consoleUrl;
+    if (datasetUrl) parts.push(`dataset ${datasetUrl}`);
+    const keyValueStoreUrl = response.storages.keyValueStores?.default.consoleUrl;
+    if (keyValueStoreUrl) parts.push(`key-value store ${keyValueStoreUrl}`);
+    return `\nConsole: ${parts.join(' | ')}\n${VERBATIM_LINKS_NUDGE}`;
+}
```
MQ37: change: this seems quite complicated to me, it took me a while to understand the flow and how the data is passed. What  about we merge `consoleLinksText` and `applyConsoleUrls` into one function so instead of two passes through the object we do only one, it modifies and also at the end returns the string? I think it will be easily understandable as it will only be one func to read and also more efficient from data POV.

src/utils/console_link.ts
```diff
@@ -0,0 +1,78 @@
...
+ * For an organization-scoped token the `users/me` lookup resolves to the
+ * organization itself, which yields org-prefixed links.
+ */
+export function resolveConsoleLinkContext(
+    apifyToken: string | undefined,
+    userInfo: CachedUserInfo,
+): ConsoleLinkContext | undefined {
+    if (!isConsoleUiToken(apifyToken)) return undefined;
+    return {
+        consoleBaseUrl: process.env.CONSOLE_BASE_URL || CONSOLE_BASE_URL,
```
MQ37: change: please remove this env var `CONSOLE_BASE_URL` as it is not set anywhere and let's use the `HOSTNAME` - see `getActorMCPServerURL` function for inspiration in the `./src/mcp/actors.ts`.

src/types.ts
```diff
@@ -649,6 +649,18 @@ export type StructuredActorCard = {
...
 };
 
+/**
+ * Context for minting Apify Console links instead of public website links.
+ * Resolved from the session token by `resolveConsoleLinkContext` — set only for
+ * Console UI token sessions. When absent, tools mint public `https://apify.com/...` links.
+ */
+export type ConsoleLinkContext = {
+    /** Console origin, e.g. `https://console.apify.com`. Trailing slashes are ignored. */
+    consoleBaseUrl: string;
```
MQ37: change: I would remove the `consoleBaseUrl` from the context type as we can use the global const directly as this value is the same for all the request.

MQ37 (changes requested):
Cool, great feature!

Couple of changes I would like to see addressed before merging - mainly simplify the URL passing flow and let's remove the unused env var and use the `HOSTNAME` to differentiate between prod and staging. Also couple of nits, feel free to either address or ignore.

Please also add some integration tests - not necessarily a new test case but at least some check in existing tests to see if we are returning the console link for logged in users.

MQ37:
@vojtechj-apify I think we should also change the logic in the Actor run widget.

Here is the current code, we should also return the personalized URLs.
`ActorRun.tsx:578`:
```
   const handleOpenRun   = () => app.openLink({ url: `https://console.apify.com/actors/runs/${runData.runId}` });                                                                        
   const handleOpenActor = () => app.openLink({ url: `https://apify.com/${runData.actorFullName}` });
```

vojtechj-apify:
> @vojtechj-apify I think we should also change the logic in the Actor run widget.
> 
> Here is the current code, we should also return the personalized URLs. `ActorRun.tsx:578`:
> 
> ```
>    const handleOpenRun   = () => app.openLink({ url: `https://console.apify.com/actors/runs/${runData.runId}` });                                                                        
>    const handleOpenActor = () => app.openLink({ url: `https://apify.com/${runData.actorFullName}` });
> ```

I have discussed that during implementation with @jirispilka and we decided that for now, we are going to scope the changes only to non-widget toolcalls.

vojtechj-apify:
Thanks for the review! Addressed everything except the rename. Summary:

**1. Drop `CONSOLE_BASE_URL` env var, derive from `HOSTNAME`** ✅ — `getConsoleBaseUrl()` mirrors `getActorMCPServerURL` (`STAGING_MCP_HOSTNAME` → `console-securitybyobscurity.apify.com`, else prod). This also fixes a latent bug: the env var was never set, so staging would have minted **prod** Console links.

**2. Remove `consoleBaseUrl` from `ConsoleLinkContext`** ✅ — the origin is global per cluster, so the builders derive it; the context now carries only `organizationId`.

**3. Merge `applyConsoleUrls` + `consoleLinksText`** ✅ — single `applyConsoleLinks(response, ctx)` does one pass (sets the `consoleUrl` fields and returns the narrative string). `linkContext` is forwarded through `FetchActorRunResult` so the run response builds links once at the response layer.

**5. `Console:` → `Apify Console:`** ✅ — run narrative + storage tools.

**4. Rename `consoleUrl` → `apifyConsoleUrl`** ⏸ — holding for @jirispilka's call since you tagged him. Cheap/mechanical, happy to apply if you both want it.

**6. Integration test** ✅ (partial) — folded a check into the `call-actor` run-response validation (`expectNormalModeTestStructuredContent`): with CI's API token, the run + storage responses must carry **no** `consoleUrl` and no `Apify Console:` nudge.

On the **positive** (logged-in/Console) path in CI — I dug into this and it isn't feasible without new surface:
- Console links gate on a **UI token** (`apify_ui_`). UI tokens are login-session tokens (a distinct IAM access-key type), minted only via the login flow (`/login-with-password`, …) — there is **no API→UI token exchange endpoint**.
- CI authenticates with an **API token** (`APIFY_TEST_USER_API_TOKEN`); there are **no login credentials** among the CI secrets, so we can't mint a UI token at test time.
- A **fake** `apify_ui_` token passes the prefix gate, but the tool then makes real authenticated API calls (`run.get`, `listItems`) with that same string → 401, so you never reach a link-bearing response. The token string is used for both gating and auth, so it can't be one without the other.

So the positive path stays covered by **unit tests** (mocked `users/me` + client, incl. personal/org and the staging origin) plus the **manual end-to-end** I ran on a local platform (personal + org UI tokens → correct prefixed links; API tokens → none; verified through the actual Console chat UI).

src/types.ts
vojtechj-apify: Done

src/tools/core/actor_run_response.ts
vojtechj-apify: I understand the motivation as we could then have better context, @jirispilka what is your verdict?

vojtechj-apify: We agreed that we will go with `apifyConsoleUrl`

tests/unit/console_link.test.ts
```diff
@@ -0,0 +1,125 @@
...
+
+describe('isConsoleUiToken', () => {
+    it('detects Console UI tokens by prefix', () => {
+        expect(isConsoleUiToken('apify_ui_abc123')).toBe(true);
+        expect(isConsoleUiToken('apify_api_abc123')).toBe(false);
+        expect(isConsoleUiToken('legacy-token-format')).toBe(false);
+        expect(isConsoleUiToken(undefined)).toBe(false);
+        expect(isConsoleUiToken('')).toBe(false);
+    });
+});
```
jirispilka:
```suggestion
```

Nit: I would remove these tests. It is quite trivial and covered in the other parts of the tests. It tests trivial prefix

src/utils/console_link.ts
```diff
@@ -0,0 +1,79 @@
...
+ * website links should be used.
+ *
+ * Policy (apify/apify-core#27286): UI tokens are issued only to Console sessions
+ * (e.g. the Console AI chat), so they are a verifiable signal that the user is in
+ * Console → Console links. All other sessions → public `apify.com` links.
+ *
+ * For an organization-scoped token the `users/me` lookup resolves to the
+ * organization itself, which yields org-prefixed links.
+ */
+export function resolveConsoleLinkContext(
```
jirispilka: change: We export 3 functions for one decision: isConsoleUiToken, resolveConsoleLinkContext, getConsoleLinkContext. One is enough. getUserInfoCached is cached, so the extra call in getConsoleLinkContext is free. Keep only getConsoleLinkContext, make the other two private or delete them. Then search_actors.ts and fetch_actor_details_common.ts can go back to their old shape plus one line. Tests for the two removed functions go away too.

src/tools/core/actor_run_response.ts
```diff
@@ -672,6 +711,7 @@ export async function fetchActorRunData(params: {
     abortSignal?: AbortSignal;
     mcpSessionId?: string;
     onAbort?: (runId: string, client: ApifyClient) => Promise<void>;
+    linkContext?: ConsoleLinkContext;
```
jirispilka: change: fetchActorRunData takes linkContext but never uses it — it only copies it into the result. The callers already have linkContext, so they can pass it straight to buildGetActorRunSuccessResponse({ ...fetchResult.result, widget, linkContext }). Then we can drop the param, the field on FetchActorRunResult, and this comment.

src/tools/common/get_dataset.ts
```diff
@@ -58,6 +60,11 @@ export const getDataset: ToolEntry = Object.freeze({
         // normalization `buildRunDataset` applies so this tool's `fields` matches
         // the structured `storages.datasets.default.fields` shape.
         const normalized = dataset.fields ? { ...dataset, fields: normalizeDatasetFields(dataset.fields) } : dataset;
-        return { content: [{ type: 'text', text: wrapJsonText(normalized) }] };
+        return {
+            content: [
+                { type: 'text', text: wrapJsonText(normalized) },
+                ...buildConsoleLinkContent(linkContext ? buildConsoleDatasetUrl(linkContext, dataset.id) : undefined),
+            ],
+        };
```
jirispilka: change: This linkContext ? buildConsoleXUrl(...) : undefined line repeats in all 5 storage tools. Let the URL builders accept undefined and return undefined. Then every call site is one short expression, no ternary

tests/unit/tools.get_dataset.test.ts
```diff
@@ -147,4 +154,24 @@ describe('get-dataset fields normalization', () => {
         expect(json).toEqual(raw);
         expect(json).not.toHaveProperty('fields');
     });
+
+    it('appends the dataset Console link (from the API-returned id) for Console UI token sessions', async () => {
+        vi.mocked(getUserInfoCached).mockResolvedValue({
+            userId: 'USER_ID',
+            userPlanTier: 'FREE',
+            isOrganization: false,
+        });
```
jirispilka: change: This same vi.mock block + the same mockResolvedValue({ userId, userPlanTier, isOrganization }) is copied in 6 test files. Put a mockUserInfo() factory in [tests/unit/helpers/tool_context.ts](https://github.com/apify/apify-mcp-server/blob/claude/nice-mccarthy-hgr5vd/tests/unit/helpers/tool_context.ts). Bonus: 4 old test files were touched only to add isOrganization: false — with a factory, the next field change touches one file, not ten.

tests/unit/tools.search_actors.default.response.test.ts
```diff
@@ -169,4 +169,61 @@ describe('search-actors without widget (defaultSearchActors)', () => {
...
+
+            const { structuredContent, content } = await callWithToken('apify_ui_test');
+            const consoleUrl = `https://console.apify.com/actors/${MOCK_STORE_ACTOR.id}`;
+
+            expect(structuredContent.actors[0].url).toBe(consoleUrl);
+            expect(content[0].text).toContain(`## [${MOCK_STORE_ACTOR.title}](${consoleUrl})`);
+            expect(content[0].text).not.toContain(`${APIFY_STORE_URL}/apify/web-scraper`);
+        });
+
+        it('mints org-prefixed Console links for an org-scoped UI token', async () => {
```
jirispilka: change: The org prefix is built in one place (buildConsoleUrl) and fully tested in console_link.test.ts. Here and in buildStartRunResponse we test it again through the whole tool stack. One token case per tool is enough — leave the org/personal matrix to the builder tests.

jirispilka (approved):
I like it. Great job! Pre-approving.

But I'm not really happy about the repetitions in the tests.

It was a bit hard to convince claude to simplify it but eventually we agreed on a couple of spots that would be nice to simplify.

A couple of suggestions is below

jirispilka:
@vojtechj-apify I can see there will be many conflicts :( 

Btw, for the changes I suggested. I have it in claude session so If you agree I can create a new branch and, after review, we can merge it to this PR, say a word :)

vojtechj-apify:
> @vojtechj-apify I can see there will be many conflicts :(
> 
> Btw, for the changes I suggested. I have it in claude session so If you agree I can create a new branch and, after review, we can merge it to this PR, say a word :)

@jirispilka sure lets do it that way and thanks for review

jirispilka:
> > @vojtechj-apify I can see there will be many conflicts :(
> > Btw, for the changes I suggested. I have it in claude session so If you agree I can create a new branch and, after review, we can merge it to this PR, say a word :)
> 
> @jirispilka sure lets do it that way and thanks for review

PR here: https://github.com/apify/apify-mcp-server/pull/988

I'm not sure whether there are some pending comments or issues from @MQ37 
If not (and If you are happy with #988 ) we can resolve the merge conflicts and merge it

vojtechj-apify:
> > > @vojtechj-apify I can see there will be many conflicts :(
> > > Btw, for the changes I suggested. I have it in claude session so If you agree I can create a new branch and, after review, we can merge it to this PR, say a word :)
> > 
> > 
> > @jirispilka sure lets do it that way and thanks for review
> 
> PR here: #988
> 
> I'm not sure whether there are some pending comments or issues from @MQ37 If not (and If you are happy with #988 ) we can resolve the merge conflicts and merge it

I have resolved issues and comments from @MQ37, #988 looks good so lets finish this

MQ37 (approved):

PR #961, opened 2026-06-07
fix: Wrong default telemetry for isError without classification

### LINKED ISSUE #936: fix: Wrong default telemetry for isError without classification (P1)
Part of #935. Fully independent — ships without the `respond*` refactor.

## Problem

`extractToolTelemetry` (`src/utils/tool_status.ts:117-126`) defaults a response that is `isError: true` with no telemetry to:

```
toolStatus: SOFT_FAIL          // "user error"
failure_category: INTERNAL_ERROR // "server's fault"
```

That combination is incoherent and almost never what the caller meant. `buildPermissionApprovalResponse` and several `server.ts` paths fall into it today, polluting Segment.

## Fix

Flip the default to a coherent **`FAILED + INTERNAL_ERROR`** ("we don't know what happened → assume our fault, keep it visible as a server failure to investigate").

## Acceptance

- Test-first: add a failing unit test asserting the corrected default for `isError: true` + no telemetry, confirm it fails, then apply the one-line fix.
- `pnpm run type-check`, `pnpm run lint`, `pnpm run test:unit` clean.

mvanhorn:
## Summary

`extractToolTelemetry` now defaults an `isError: true` response without explicit telemetry to `FAILED` instead of `SOFT_FAIL`. The previous default paired `SOFT_FAIL` ("user error") with `failure_category: INTERNAL_ERROR` ("server fault"), which is an incoherent combination.

## Why this matters

As described in #936, `buildPermissionApprovalResponse` and several `server.ts` paths fall into the no-telemetry branch today, so Segment receives events that claim a user error caused by an internal server problem. Flipping the default to `FAILED + INTERNAL_ERROR` keeps unclassified errors visible as server failures to investigate, and any caller that means `SOFT_FAIL` can still say so explicitly via `toolTelemetry.toolStatus` (that path is untouched and now covered by a test).

## Changes

- The default branch in `src/utils/tool_status.ts:118-126` returns `TOOL_STATUS.FAILED`; explicit telemetry still wins (resolution order unchanged, doc comment updated to match).

## Testing

- Updated the existing default-branch unit test to assert `FAILED + INTERNAL_ERROR`, written test-first against the old behavior.
- Added a regression test asserting explicit `toolTelemetry` (`SOFT_FAIL + INVALID_INPUT`) is preserved when `isError` is true.
- `pnpm format:check`, `pnpm lint`, `pnpm type-check`, and `pnpm test:unit` (828 passed, 2 skipped) all clean locally.

Fixes #936

### DIFF
--- a/src/utils/tool_status.ts
+++ b/src/utils/tool_status.ts
@@ -98,7 +98,7 @@ export function extractAjvErrorDetails(errors: ErrorObject[] | null | undefined)
  *
  * toolStatus resolution:
  * 1. `toolTelemetry.toolStatus` present → use it directly.
- * 2. `isError` set without toolStatus → SOFT_FAIL.
+ * 2. `isError` set without toolStatus → FAILED.
  * 3. Neither → SUCCEEDED.
  *
  * actorName/actorId are server-side context (from tool resolution).
@@ -118,7 +118,7 @@ export function extractToolTelemetry(
     if (!telemetry) {
         if (res.isError) {
             return {
-                toolStatus: TOOL_STATUS.SOFT_FAIL,
+                toolStatus: TOOL_STATUS.FAILED,
                 callDiagnostics: { failure_category: FAILURE_CATEGORY.INTERNAL_ERROR, ...actorFields },
             };
         }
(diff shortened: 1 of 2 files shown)


MQ37 (approved):
LGTM, thanks!

jirispilka (approved):
Cool, thanks!

mvanhorn:
Thanks for the merge, @jirispilka. Telemetry defaulting correctly for unclassified isError keeps the metrics honest.

PR #964, opened 2026-06-08
fix: Return binary key-value store records as MCP content blocks

### LINKED ISSUE #883: fix: get-key-value-store-record binary handling
## Problem
`src/tools/common/get_key_value_store_record.ts:59` always `JSON.stringify(record)`. For binary records (`Buffer`) this dumps `{"type":"Buffer","data":[...]}` — unusable. MCP spec 2025-11-25 supports `image`/`audio`/`resource` content types for exactly this case.

## Proposed solution
Branch the response on the record's content type — return `image`/`audio`/`resource` for binary, text for text, JSON stringify only for JSON.

Parent: #784

RobertCrupa:
## Problem

`get-key-value-store-record` always `JSON.stringify`d the record. For binary values (which the SDK returns as a `Buffer`) this serialized to the useless `{"type":"Buffer","data":[...]}`.

## Fix

Branch the response on the parsed value:

- `Buffer` + `image/*` → image content block (base64 data)
- `Buffer` + `audio/*` → audio content block (base64 data)
- `Buffer` (other) → embedded resource block (base64 blob + record URL)
- `string` → plain text content block
- JSON/object → JSON-stringified record (unchanged)

MCP spec 2025-11-25 supports `image`/`audio`/`resource` content types for exactly this case.

## Tests

Added unit tests in `tests/unit/tools.get_key_value_store_record.test.ts` covering each content type.

Closes #883


MQ37 (approved):
Code LGTM good job!

The only thing is that we should limit the response size. I think we should set hard limit of let's say 100 KB, if it is a chat client like ChatGPT web ui then it is not for data processign and we do not want to clutter the context too much and if it is say Claude code we can return a link and the Claude code can download it and process it smartly or compress it. In some cases we might need signed links to access the resource - so we may need to do it in the internal repo or maybe use some API endpoint for that if it is available.

jirispilka (changes requested):
Human:
I like the direction, thanks for fixing it!

There are two things to consider:
- No size limit. A large file gets fully encoded into the response and eat the context. Fix: keep small files inline, return a link for large ones. I'm not sure about the threshold.  1MB?
- The download link is built by hand. Use the SDK to build it so it's signed and works for private records

The rest is small, many of them pre-existing. 

I ask Claude to review it 😊 

### Claude

**1. Type the return value** (`types.ts:187`)
`ToolEntry.call` returns `Promise<object>`, so the new blocks get no shape checking — a typo'd discriminant or missing required field (`mimeType` is required on `ImageContent`/`AudioContent`) only fails at runtime as an SDK `-32600` against `CallToolResultSchema` (`server.ts:1183`).
→ Annotate `call` as `Promise<CallToolResult>` (already imported in the test). Assignable to the interface's `Promise<object>`, so no interface change.

**2. Use the SDK to build the resource URL** (`:70`)
The hand-built URL re-implements `store.getRecordPublicUrl()` and drops three things it does:
- **signature** — appends `?signature=<HMAC>` for stores with `urlSigningSecretKey`; without it private-store URLs 401.
- **ID normalization** — `_toSafeId` turns `username/store-name` → `username~store-name`; raw interpolation breaks the slash form.
- **key encoding** — keys allow `' ( ) ! . space`; the SDK encodes via `new URL()`, raw interpolation yields malformed URIs.
→ `const uri = await store.getRecordPublicUrl(recordKey)` (drops the `getApifyAPIBaseUrl` import). Costs one extra `store.get()`. Matters most for the link path in #3, where the URL is the only way to the data.

**3. Cap inline binary size** (`:71`)
`getRecord()` loads the whole Buffer, `.toString('base64')` adds ~33%, shipped inline → a 20–50 MB record gives a multi-MB response (~3× transient memory) that blows transport/context limits or OOMs. No cap exists (`TOOL_MAX_OUTPUT_CHARS=50000` is defined but unused). Not a regression — but this PR makes large-binary retrieval actually useful, so the latent problem becomes reachable.
→ Size-based hybrid: **image/audio always inline** (model consumes them directly; clients often won't fetch a link), **small other binaries inline**, **large binaries → `resource_link`** with the signed URL. Don't default everything to links — `resource_link` client support is uneven and the bytes only help if the client re-fetches. Add `MAX_INLINE_BINARY_BYTES` (start ~1 MB, tune to transport limit).
Note: neither block needs `resources/list` registration — tool-result resources are decoupled from the catalog. Our `readResource` only handles `file://`/`ui://`, so a link works purely because the URL is signed/public.

## Worth a decision @MQ37 

**4. Inconsistent value contract** (`:74`)
Text returns the bare value, but JSON returns the whole `{key, value, contentType}` envelope (fallback wraps `record`, not `value`). So `get OUTPUT` gives raw value for text/binary but a metadata-wrapped object for JSON — agent must special-case per type.
→ Either `wrapJsonText(value)` for a uniform "return the value" contract (behavior change, breaks 2 JSON tests → update to `.value`), or keep the envelope deliberately with a comment. Pick one.

**5. Byte telemetry blind to binary** (`mcp.ts:83`) — *internal repo consumes this, coordinate*
`computeToolResponseBytes` only sums `content[].text`. Binary payloads live in `data` (image/audio) and `resource.blob`, so multi-MB responses report ~0 bytes to Segment + the "tool call completed" log — the biggest payloads are invisible to monitoring.
→ Also `Buffer.byteLength` of `data`, `resource.blob`, `resource.text`; update the doc comment.

## Minor

**6. Lowercase the MIME type** (`:64`)
`mimeType?.startsWith('image/')` is case-sensitive, but Content-Type isn't — `Image/PNG` skips the image branch and degrades to a generic blob. Low-probability (Apify emits lowercase) but constructible.
→ `.split(';')[0].trim().toLowerCase()`.

**7. Misleading comment** (`:59`)
Says "JSON → object, text/xml → string, else → Buffer" (the SDK's parsing rule), but the code dispatches on JS type (`Buffer.isBuffer` → `typeof === 'string'` → else). A maintainer may "fix" it by adding a mimeType check, duplicating SDK logic.
→ Reword to "SDK already parsed the body; branch on the value's JS type."

RobertCrupa:
As @MQ37 asked for 100Kb and @jirispilka for 1Mb, I went with 256Kb as a nice round number.

I would prefer to make this a more data-driven decision. Once we have a distribution of the file sizes we are serving, we can choose a more appropriate amount to balance convenience and performance

jirispilka:
> As @MQ37 asked for 100Kb and @jirispilka for 1Mb, I went with 256Kb as a nice round number.
> 
> I would prefer to make this a more data-driven decision. Once we have a distribution of the file sizes we are serving, we can choose a more appropriate amount to balance convenience and performance

I like that, 256 Kb is a nice number.
@RobertCrupa please create a new github issue to track it and re-evaluate the value based on data

RobertCrupa:
@jirispilka Should the computeToolResponseBytes function also take base64 files into account? As you mentioned earlier, they now emit 0, but they will also skew results against the current text responses.

This may be a separate issue that we can address by adding a more robust telemetry solution.

MQ37:
@RobertCrupa @jirispilka what about we track the files separately from the text/structured output so they do not skew the results?

RobertCrupa:
I have created a stacked [PR](https://github.com/apify/apify-mcp-server/pull/971) to add the telemetry for file size

src/tools/common/get_key_value_store_record.ts
```diff
@@ -55,6 +56,47 @@ export const getKeyValueStoreRecord: ToolEntry = Object.freeze({
...
+                        type: 'resource',
+                        resource: { uri, blob: data, ...(mimeType && { mimeType }) },
+                    } satisfies EmbeddedResource,
+                ],
+            };
+        }
+        if (typeof value === 'string') {
+            return { content: [{ type: 'text', text: value }] };
+        }
         return { content: [{ type: 'text', text: wrapJsonText(record) }] };
```
jirispilka:
```suggestion
        return { content: [{ type: 'text', text: wrapJsonText(value) }] };
```

We should return `value` instead of a `record`. For other content types we are also returning a value.

Say, the value is json {"name": "Bob"}, we should return it.

Now we are returning: {"key":"***","value":{"name":"Bob"},"contentType":"application/json"}

jirispilka: You will need to fix the tests too

src/tools/common/get_key_value_store_record.ts
```diff
@@ -55,6 +56,47 @@ export const getKeyValueStoreRecord: ToolEntry = Object.freeze({
...
+                const mimeSuffix = mimeType ? ` (${mimeType})` : '';
+                return {
+                    content: [
+                        {
+                            type: 'text',
+                            text: `Record '${recordKey}' is ${sizeKb} KB${mimeSuffix}, over the ${limitKb} KB inline limit. Retrieve it directly:\n${uri}`,
+                        } satisfies TextContent,
+                    ],
+                };
+            }
```
jirispilka, line 99:
```suggestion
                return {
                    content: [
                        {
                            type: 'resource_link',
                            uri,
                            name: recordKey,
                            size: value.length,
                            ...(mimeType && { mimeType }),
                        } satisfies ResourceLink,
                    ],
                };
            }
```


what if we return standardized resource_link?
But I'm not sure if the majority of clients support it.
It makes sense to me to use native MCP types, can you try it in claude, whether it will work?

jirispilka (approved):
Nice, pre-approving

I left two small comments.

I was wondering if claude or other major clients support `resource_link`?

src/tools/common/get_key_value_store_record.ts
RobertCrupa: Done [2b0292c](https://github.com/apify/apify-mcp-server/pull/964/commits/2b0292c3a15b871b1392c29fda896e864305d91a)

src/tools/common/get_key_value_store_record.ts
RobertCrupa: image
It works!

src/tools/common/get_key_value_store_record.ts
```diff
@@ -56,9 +57,50 @@ export const getKeyValueStoreRecord: ToolEntry = Object.freeze({
                 : `Key-value store '${keyValueStoreId}' not found.`;
             return buildStorageNotFound(text);
         }
-        const bytes = computeValueBytes(record.value);
+        // The SDK already parsed the body by Content-Type (JSON -> object, text/xml -> string, else -> Buffer);
+        // branch on the resulting JS type, not on the MIME type.
+        const { value, contentType } = record;
+        // Binary values can't go in structuredContent (a Buffer serializes to useless {"type":"Buffer",...});
+        // return MCP content blocks instead.
+        if (Buffer.isBuffer(value)) {
```
RobertCrupa, line 74: **Blocking: binary records declare `outputSchema` but return no `structuredContent` — official MCP SDK clients reject every binary fetch.**

This tool advertises `outputSchema: keyValueStoreRecordOutputSchema` (line 35), and `tools/list` exposes it (`getToolPublicFieldOnly`, `src/utils/tools.ts:110`). The official SDK client caches a validator for any tool with an `outputSchema` and, in `callTool`, throws when the result has neither `structuredContent` nor `isError`:

```js
// @modelcontextprotocol/sdk client/index.js:500
if (!result.structuredContent && !result.isError) {
    throw new McpError(ErrorCode.InvalidRequest,
        `Tool ${params.name} has an output schema but did not return structured content`);
}
```

Every branch in this `if (Buffer.isBuffer(value))` block (image, audio, embedded `resource`, `resource_link`) returns `{ content: [...] }` only. So fetching any binary record (PNG, MP3, PDF, >256 KB blob) throws client-side before the content blocks are ever seen — the fix only works for clients that skip output validation.

Fix options: relax the declared `outputSchema` (e.g. `anyOf` that also covers the content-block shape, or drop the `value`/`key` `required` constraints), or return a minimal schema-conforming `structuredContent` alongside the content blocks.

src/tools/common/get_key_value_store_record.ts
```diff
@@ -56,9 +57,50 @@ export const getKeyValueStoreRecord: ToolEntry = Object.freeze({
...
+                };
+            }
+            const data = value.toString('base64');
+            if (mimeType?.startsWith('image/')) {
+                return { content: [{ type: 'image', data, mimeType } satisfies ImageContent] };
+            }
+            if (mimeType?.startsWith('audio/')) {
+                return { content: [{ type: 'audio', data, mimeType } satisfies AudioContent] };
+            }
+            const uri = await store.getRecordPublicUrl(recordKey);
```
RobertCrupa: **Low (efficiency): extra network round-trip to build a `uri` that isn't needed for an inlined resource.**

`getRecordPublicUrl` does an internal `await this.get()` to fetch store metadata for URL signing (apify-client `key_value_store.js:175`). This embedded-resource branch already inlines the full base64 `blob`, so the `uri` is decorative — yet it costs a store-metadata fetch on every non-image/non-audio binary under the limit. Consider a cheaper synthetic `uri`, or only resolve the public URL when the payload isn't inlined.

src/tools/common/get_key_value_store_record.ts
```diff
@@ -56,9 +57,50 @@ export const getKeyValueStoreRecord: ToolEntry = Object.freeze({
...
+        // branch on the resulting JS type, not on the MIME type.
+        const { value, contentType } = record;
+        // Binary values can't go in structuredContent (a Buffer serializes to useless {"type":"Buffer",...});
+        // return MCP content blocks instead.
+        if (Buffer.isBuffer(value)) {
+            // Content-Type is case-insensitive; lowercase so the image/audio checks below don't miss `Image/PNG`.
+            const mimeType = contentType?.split(';')[0].trim().toLowerCase();
+            if (value.length > KV_RECORD_MAX_INLINE_BYTES) {
+                // base64-inlining a large binary would blow up the context window; return a link instead.
+                const uri = await store.getRecordPublicUrl(recordKey);
```
RobertCrupa, line 86: **Low (verify): `resource_link` can point at an unsigned, inaccessible URL.**

`getRecordPublicUrl` only appends `?signature=` when `store.urlSigningSecretKey` is present (apify-client `key_value_store.js:177`). When the token can't read the signing key (e.g. a shared store), the returned URL needs API auth — but a `resource_link` carries no token and has no inline fallback, so a client following it for a large record gets 401/403. The owner-token path on one's own store returns a signed URL, so this is narrow; worth confirming it's always signed for the cases this tool serves.

RobertCrupa (commented):
Code review (xhigh, recall mode) of the binary key-value-store handling.

**1 blocking issue**: the binary branches return content blocks but no `structuredContent`, while the tool declares an `outputSchema` — official MCP SDK clients reject that response before the content is ever read. Plus 2 low-severity notes (a wasted network call, and an unsigned-URL edge case).

The telemetry/byte-accounting changes (`computeToolResponseBytes`, `buildResponseBytesTelemetry`, `fileBytes`) and the new tests check out — no issues found there.

src/tools/common/get_key_value_store_record.ts
RobertCrupa: Fixed in [33eb892](https://github.com/apify/apify-mcp-server/pull/964/commits/33eb892341a843f77d38f7b472cd986786b50dbb) by returning a minimal schema-conforming structuredContent.

I decided to add unit tests that use a pedantic MCP client, which proved the comment was correct: it did reject the binary records.

src/tools/common/get_key_value_store_record.ts
RobertCrupa: This is not the case

src/tools/common/get_key_value_store_record.ts
RobertCrupa: Only a minor performance increase but why not, added in [8532be7](https://github.com/apify/apify-mcp-server/pull/964/commits/8532be7113319ac878dbb3e834e883bb0063a1ba)

PR #967, opened 2026-06-08
feat: Add MCP Storage Access telemetry event

### LINKED ISSUE #887: chore: Add mixpanel analytics for storage tools + error rates

RobertCrupa:
Closes #887.

Adds a dedicated `MCP Storage Access` Segment event for storage tools (dataset / key-value store), so storage usage and error rates can be analysed separately from the generic `MCP Tool Call` event.

## What changed

- **New event.** Storage tools now emit `MCP Storage Access` in addition to `MCP Tool Call`. `getStorageType()` maps each storage tool name to `DATASET` or `KEY_VALUE_STORE`; non-storage tools return `null` and emit nothing extra.
- **Event properties.** `buildStorageAccessProperties()` projects the finalized tool-call event onto the storage event: keeps the common envelope plus status / error fields, adds `storage_type`, and drops actor / validation fields that carry no meaning for storage tools. Absent optional fields are omitted rather than set to `undefined`.
- **Telemetry plumbing.** Extracted a shared `trackEvent()` helper; `trackToolCall()` and the new `trackStorageAccess()` both delegate to it.
- **Error classification fix.** `get-dataset-schema` now classifies a no-schema generation failure as `INTERNAL_ERROR` with a `failureDetail`, instead of leaving the failure uncategorised — so it shows up correctly in the error-rate breakdown.

## Tests

- `getStorageType()` — dataset / KVS mapping + null for non-storage tools.
- `trackStorageAccess()` — emits the dedicated event.
- `buildStorageAccessProperties()` — field projection, dropped fields, omitted optionals.
- Updated `get-dataset-schema` failure test to assert the new category + detail.

type-check, lint, and `test:unit` (834 pass) all green.

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -160,6 +160,12 @@ export const FAILURE_CATEGORY = {
     INTERNAL_ERROR: 'INTERNAL_ERROR',
 } as const;
 
+// Storage type for the dedicated `MCP Storage Access` telemetry event.
+export const STORAGE_TYPE = {
+    DATASET: 'DATASET',
+    KEY_VALUE_STORE: 'KEY_VALUE_STORE',
+} as const;
+
 // Apify API error types
 export const APIFY_ERROR_TYPE_FULL_PERMISSION_NOT_APPROVED = 'full-permission-actor-not-approved';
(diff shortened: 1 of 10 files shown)


RobertCrupa:
Closing as I misunderstood the issue. No code is required; only Mixpanel dashboard tweaks.

PR #971, opened 2026-06-09
feat: Track file response bytes separately in telemetry

RobertCrupa:
Stacked on #964.

## What

Adds a third tool-response size metric, `tool_response_file_bytes`, to `ToolCallTelemetryProperties`, alongside the existing `tool_response_content_bytes` (text) and `tool_response_structured_content_bytes`.

## Why

WIP work counted base64 media and embedded-resource bytes as text content, which would **skew** `tool_response_content_bytes` (meant to measure conversational text the tool wrote for the model). File/record payloads can dwarf the text, so this PR keeps them in a separate bucket.

## Changes

- `computeToolResponseBytes` now returns `fileBytes` — the UTF-8 bytes of returned files/records in `content[]`: image/audio base64 `data` and embedded `resource` `blob`/`text`. `contentBytes` stays text-only (`content[].text`).
- New `tool_response_file_bytes` telemetry field + matching `responseFileBytes` log field.
- `buildResponseBytesTelemetry` is the single source of truth for the `tool_response_*_bytes` field set, so the call site spreads one helper instead of enumerating each field (the repetition this metric would otherwise have tripled).
- Unit tests for the new bucket and the helper.

## Internal repo

Touches `ToolCallTelemetryProperties` (a field the hosted server consumes via Segment). The new field is optional and additive — no breaking change — but `apify-mcp-server-internal`'s telemetry contract suite may want a matching assertion.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -88,7 +88,12 @@ import type {
 } from '../types.js';
 import { ServerMode, TOOL_TYPE } from '../types.js';
 import { getHttpStatusCode, logHttpError } from '../utils/logging.js';
-import { buildMCPResponse, computeToolResponseBytes, getToolCallErrorUserText } from '../utils/mcp.js';
+import {
+    buildMCPResponse,
+    buildResponseBytesTelemetry,
+    computeToolResponseBytes,
+    getToolCallErrorUserText,
+} from '../utils/mcp.js';
 import { buildPaymentRequiredResponse } from '../utils/payment_errors.js';
 import { createProgressTracker } from '../utils/progress.js';
 import { getServerInstructions } from '../utils/server-instructions/index.js';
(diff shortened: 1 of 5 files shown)


MQ37 (approved):

src/mcp/server.ts
```diff
@@ -1391,6 +1396,7 @@ export class ActorsMcpServer {
             ...(params.responseBytes !== undefined && {
                 responseContentBytes: params.responseBytes.contentBytes,
                 responseStructuredContentBytes: params.responseBytes.structuredContentBytes,
+                responseFileBytes: params.responseBytes.fileBytes,
```
jirispilka, line 1399: Should you use `buildResponseBytesTelemetry` here as well? Instead of listing all the types individually: responseContentBytes, responseStructuredContentBytes, responseFileBytes

jirispilka (approved):
Nice, I left a small nit, otherwise looks ok

src/mcp/server.ts
RobertCrupa: The problem is that one is camelCase and the other is snake_case. I can create a new helper for camel case to keep telemetry consistent, but it would only be used once.

I think I will leave it, but I will create the helper if we duplicate this code again.

PR #972, opened 2026-06-09
docs: split AGENTS.md into a hierarchical per-directory tree

MQ37:
## What

Layer the root `AGENTS.md` into a router + per-directory child docs:

- **Flip the symlink** — `AGENTS.md` is now the real file; `CLAUDE.md` aliases it.
- **Root** keeps the full instructions (communication style, scope discipline, git, verification, mcpc testing, the public/internal boundary, tool-authoring conventions) and gains a child link index + a "keep AGENTS.md updated" rule + `check:agents` in the verify set.
- **Child `AGENTS.md`** for `src/`, `src/mcp/`, `src/tools/`, `src/payments/`, `src/resources/`, `src/web/` — each a map of the directory plus its non-obvious gotchas (tool-name cap/hash, URL-keyed proxy IDs, Skyfire-vs-x402 payment data location, the separate `src/web` build).
- **`DESIGN_SYSTEM_AGENT_INSTRUCTIONS.md`** relocated into `src/web/`, next to the widget code it governs.
- **`scripts/check-agents-links.mjs`** + a CI step validate the link tree.

## Why

A single root file can't be both small and detailed. Splitting by directory keeps each fact next to its code while the root stays the instruction hub; the link-checker prevents silent doc rot when files move. Naming/coding standards remain owned by `CONTRIBUTING.md` (linked, not copied).

## Testing

`pnpm run lint`, `pnpm run format:check`, `pnpm run type-check`, `pnpm run check:agents` (7 docs) all green locally. Docs + one CI step; no runtime code touched.


scripts/check-agents-links.mjs
```diff
@@ -0,0 +1,78 @@
...
+// Validates the AGENTS.md tree so the "keep AGENTS.md updated" rule is mechanical:
+//   - every relative link resolves on disk (no rot when a file is renamed/moved)
+//   - every AGENTS.md is reachable from the root AGENTS.md (no orphan docs)
+//   - every non-root AGENTS.md links up to an ancestor AGENTS.md
+//   - an optional `<!-- agents-scope: dir -->` marker matches the file's location
+// Anchors are not validated — heading slugs are fuzzy and not worth the false positives.
+import { readdirSync, readFileSync, existsSync, statSync } from 'node:fs';
+import { dirname, join, relative, resolve } from 'node:path';
+
+const root = process.cwd();
```
jirispilka, line 10: Let's use: `const root = resolve(dirname(fileURLToPath(import.meta.url)), '..');` instead of cwd().

`cwd` will fail if you are in other dir, expect root.

scripts/check-agents-links.mjs
```diff
@@ -0,0 +1,78 @@
...
+    const text = readFileSync(doc, 'utf8');
+    const dir = dirname(doc);
+    const here = relative(root, dir) || '.';
+
+    const scope = text.match(/<!--\s*agents-scope:\s*(\S+?)\/?\s*-->/)?.[1];
+    if (scope && scope !== here) failures.push(`${relative(root, doc)}: scope '${scope}' != location '${here}'`);
+
+    const targets = [...text.matchAll(/\]\(([^)\s]+)\)/g)]
+        .map((match) => match[1])
+        .filter((link) => !/^(https?:|mailto:|#)/.test(link));
```
jirispilka: do we need mailto? :)

jirispilka (approved):
Thanks!

I added two nits.

Then I decided to go further and I asked Claude to simplify the script to check AGENTS.md

### Here is Claude review, feel free to ignore :) 


The link checker hand-rolls a recursive `walk()` over the whole working tree behind an open denylist (`node_modules|dist|coverage|.git`). That design causes three real issues:

1. **Crash on broken symlinks (correctness).** `walk()` has no error handling and  `statSync` follows symlinks. The walk descends untracked dirs like `.pnpm-store/`  (pnpm stores are full of symlinks); one dangling symlink throws an uncaught `ENOENT`  and aborts the whole check with a stack trace instead of a clean failure.

2. **Force-enrolls stray docs → false CI failures (altitude).** Anything not in the
   denylist is walked. The moment an `AGENTS.md` lands under an unlisted dir
   (`.claude/`, `.cursor/`, `.idea/`, `evals/`, a plugin/IDE scaffold, a vendored copy),   it gets pulled into the tree and fails the orphan + up-link checks — reddening CI for  a file unrelated to the repo's docs. The "fix" is to keep appending to the denylist forever.

3. **Wasted I/O.** It `statSync`s every entry in untracked trees (`.pnpm-store/`, etc.) on every run.

#### Suggested fix: drive the walk from `git ls-files`

```js
import { execFileSync } from 'node:child_process';

const root = execFileSync('git', ['rev-parse', '--show-toplevel'], { encoding: 'utf8' }).trim();
const docs = execFileSync('git', ['ls-files', '-z', '*AGENTS.md'], { cwd: root, encoding: 'utf8' })
    .split('\0').filter(Boolean).map((p) => join(root, p));
```

This lists exactly the tracked `AGENTS.md` files and nothing else. It deletes `walk()`, the `SKIP` denylist, the `statSync`/`readdirSync` imports, and the `fileURLToPath` root anchor — collapsing issues #1, #2, #3, and the hardcoded `..` root hop in one move.

**Tradeoff:** adds a dependency on git being present and running inside a checkout. For a repo-internal CI / pre-commit check that's already guaranteed, so it's a safe assumption here.

#### Smaller fixes worth folding in

- **Undirected reachability (correctness).** The orphan-check BFS follows *all* links (down, sideways, up), so it doesn't enforce the stated "reachable by descending from root" invariant — a sibling sideways-link can mask a missing parent→child link. If the intent is a strict tree, only traverse downward links in the BFS.
- **`isDoc` substring match.** `endsWith('AGENTS.md')` also matches `MYAGENTS.md`. Use a basename `===` check.
- **Windows path separators (only if non-Linux contributors run it).** The scope-marker comparison and `SKIP` use `/` literals; `relative()`/`join()` emit `\` on Windows. Moot if the walk moves to `git ls-files`, which always returns `/`-separated paths.

scripts/check-agents-links.mjs
MQ37: you never know :smile:

MQ37:
I refactored the check script.

Issue #935, opened 2026-06-01
refactor: Tool response contract via respond* constructors (Option F)


jirispilka:
Tracking issue for the tool-response contract refactor. Full analysis and option comparison: #929.

## Problem

Tool handlers build MCP responses by hand, so classification is **opt-in** — a handler can return `isError: true` and forget the telemetry, silently landing in the wrong `SOFT_FAIL + INTERNAL_ERROR` bucket. Nothing catches it (not the compiler, not review). Concretely:

- **P1** — `extractToolTelemetry` defaults `isError: true` + no telemetry to `SOFT_FAIL + INTERNAL_ERROR`. Wrong combination; almost never what callers mean.
- **P2** — `get_dataset_schema.ts:85` returns `toolStatus: FAILED` with no `failureCategory` → "category unknown" bucket in Segment.
- **P3** — `{ toolStatus: SOFT_FAIL, failureCategory: INVALID_INPUT }` is re-typed at **10** sites. No shared "this is a user error" constructor.
- **P4** — `isError: true` without telemetry is structurally legal — the wire format can't enforce classification.
- **P5** — No "informational, non-error" status; `"Dataset 'X' is empty."` is indistinguishable from data success.

Net effect: ~8 hand-rolled `` ```json `` fences, six response shapes, and unreliable failure telemetry in Segment.

## Chosen solution — Option F

Make a small constructor family the **only** public way to build a response. `buildMCPResponse` becomes private to `utils/mcp.ts`. The output type is branded, so a hand-rolled shape or a bare `isError: true` **won't compile** — the bad shape becomes unconstructable.

```ts
respondOk(text, opts?)          // success with text
respondJson(value, opts?)       // success carrying JSON — owns the ```json fence
respondEmpty(text)              // success, intentionally no data
respondUserError(text, opts?)   // SOFT_FAIL; category defaults to INVALID_INPUT
respondServerError(text, opts?) // FAILED; category/httpStatus derived from the caught error
```

Why F is cheap:

- **`buildMCPResponse` is not on the `internals` surface.** `apify-mcp-server-internal` consumes only the stripped wire shape; `toolTelemetry` is stripped inside this package before the boundary. Changing *how* we build the wire shape needs **zero cross-repo coordination** as long as the serialized output is unchanged.
[...]

PR #983, opened 2026-06-11
feat(telemetry): correlate tool calls to runs and storages

MQ37:
## What
Add identifiers to the `"MCP Tool Call"` Segment event so a call ties to the resource it touched:
- `run_id` + `run_status` on `call-actor`, `get-actor-run`, and direct Actor tools.
- `run_id` only on `get-actor-run-log` and `abort-actor-run` (their responses carry no `RunResponse`,
  so the run's status isn't available at the telemetry sink).
- `dataset_id` / `key_value_store_id` on the storage tools (`get-dataset(-items|-schema)`,
  `get-key-value-store(-keys|-record)`), on success and not-found.
- `anonymousId` for unauthenticated calls = `mcp_session_id` (was a fresh random UUID per event).

## Why
We had no way to correlate a tool call to its Apify run/storage for observability. `run_status` is
the run's own outcome, **distinct from `tool_status`** — a call succeeds even when it returns a
still-`RUNNING`/`FAILED` run. The per-event random `anonymousId` made every unauthenticated call
look like a new user, so loops/retries/funnels were unreconstructable; the session id fixes that
(random UUID kept as a last-resort fallback).

All four ids are **derived once at the telemetry sink** (`deriveResourceIds`) from data the server
already holds, so no tool or response builder is touched and nothing internal rides the response
(nothing to strip, nothing to leak):
- `run_id`/`run_status` from the public, spec'd `RunResponse` `structuredContent`.
- `run_id` (for `abort`/`log`/not-found) and `dataset_id`/`key_value_store_id` from the validated
  input args (storage ids quote-stripped to match the tools).

The `structuredContent` read is **typed to `RunResponse`**, so renaming or retyping `runId`/`status`
is a compile error, and a regression test ties `deriveResourceIds` to a real `buildStartRunResponse`
output so a runtime shape drift fails CI rather than silently blanking the field.

Behavior / limits to expect: unique-anonymous-user counts drop (a session's events now share one
identity). A `call-actor`/direct-Actor run that *starts and then fails the wait* won't carry `run_id`
(its error response has no `RunResponse` and `call-actor` has no `runId` arg) — a deliberate trade for
not threading ids through every tool.

## Testing
`pnpm type-check`, `pnpm lint`, `pnpm format:check`, `pnpm test:unit` (909 pass). Added unit tests for
`deriveResourceIds` (run ids from `structuredContent`, quote-stripped storage ids + raw `runId` from
args, structuredContent-wins precedence, not-found via args, the real-builder regression guard) and
`anonymousId` = session id (+ random fallback). Verified the compile-time guard by retyping
`RunResponse.status` and confirming the build breaks at `deriveResourceIds`.


vojtechj-apify:
Good, i like the fact this gives us better observability and replaces random UUID with session ID. Logs should provide cleaner insight right now. 

As far as review goes I haven't found any issue. Telemetry wise, maybe consider adding following in the future (sorry if it is already present, i just took a quick look)
  1. **Search telemetry (`search-actors`, `search-apify-docs`)** — currently zero semantic data. Add `search_query` and `result_count`. Zero-result rate is the best signal for "users want
  something we don't surface." Highest value.
  2. **`fetch-actor-details` success path** — only logs `actor_id`/`actor_name` on failure today; add it on success so we can see which Actors users inspect.
  3. **`fetch-apify-docs` success path** — same gap; log the doc URL/path on success, not just on error.
  4. **add-actor** — log `actor_id` so we can see which Actors get added as dynamic tools.

vojtechj-apify (approved):
Good job, nothing on my end.

jirispilka (approved):
Honestly, this looks more complicated than I would expect.
I think the main culprit is a bad pre-existing design. Something that we will need to fix, as discussed here: https://github.com/apify/apify-mcp-server/issues/935

I had some back and forth with Fable :).
There is a simpler solution that would deliver 95% of the feature.

I would prefer a simpler solution (perhaps your council would prefer that too ;) ), pre-approving, your decision.

### The simpler design that was missed

The key observation: almost every id this PR threads through 12 files is already in the server's hands at the existing telemetry choke point, logToolCallAndTelemetry (server.ts:1373) — which both the sync and task paths already call:

- For get-dataset(-items|-schema), get-key-value-store(-keys|-record), get-actor-run, abort-actor-run, get-actor-run-log, the resource id is a validated input argument (args.datasetId / args.keyValueStoreId / args.runId). It covers not-found paths for free.
- For call-actor, direct Actors, and widgets, runId / status / actorId are already on the public structuredContent (the spec'd RunResponse shape) — the server holds that exact object in toolResult.

So a single ~20-line deriveResourceIds(args, result) folded into logToolCallAndTelemetry, plus the types.ts fields and the 1-line anonymousId change, delivers ~95% of this feature with zero tool-file changes, zero builder changes, zero new strip sites, and zero leak risk (it reads public fields rather than smuggling internal ones). Roughly 6 files instead of 21.

Honest trade-offs — what derivation would not capture:

- run_id on call-actor error responses after the run started (error responses have no structuredContent; the PR threads runId into buildCallActorErrorResponse for this).
- run_status for abort-actor-run (its response is raw JSON text).
Args-derived ids are pre-stripQuoteWrappers (trivially fixable inside the helper).

**Slim it before merge (if MQ37 has appetite): keep types.ts, telemetry.ts, the two server.ts strip calls, and optionally the buildCallActorErrorResponse runId threading; replace the ~12 tool/builder changes with derivation at logToolCallAndTelemetry. You lose only the abort-actor-run run_status edge case.**

MQ37:
@jirispilka I reworked this PR to use the approach of deriving the IDs from the structured content and input args - it is much smaller but more fragile on case the shape changes.

src/utils/tool_status.ts
```diff
@@ -2,10 +2,51 @@ import { ErrorCode, McpError } from '@modelcontextprotocol/sdk/types.js';
...
+ *
+ * - `run_id` / `run_status` come from the spec'd RunResponse `structuredContent` (call-actor,
+ *   get-actor-run, direct Actors, widgets). `run_status` is the run's own outcome (RUNNING/SUCCEEDED/
+ *   FAILED), distinct from `tool_status` — the call SUCCEEDS even when the run is still running/failed.
+ * - `run_id` also comes from `args.runId` for tools whose response carries no RunResponse
+ *   (abort-actor-run, get-actor-run-log) and for not-found.
+ * - `dataset_id` / `key_value_store_id` come from `args`, quote-stripped to match the storage tools.
+ */
+export function deriveResourceIds(args: Record<string, unknown> | undefined, result: unknown): ResourceIds {
+    const a = (args ?? {}) as { datasetId?: unknown; keyValueStoreId?: unknown; runId?: unknown };
```
jirispilka: Nit: Single letter arg `a` reads badly. I would put `argObj` here

jirispilka (approved):
Nice PR! I like it!

One suggestions from Claude though below. 
It is more about style but ti will simplify the code a bit :) 

1. make logToolCallAndTelemetry take the raw result (and args) and compute both
  derived-from-result telemetry internally:

  // inside logToolCallAndTelemetry, replacing the responseBytes param with `result` + `args`
```ts
const responseBytes = result === undefined ? undefined : computeToolResponseBytes(result);
  const finalizedTelemetryData = {
      ...telemetryData,
      tool_status, tool_exec_time_ms: durationMs,
      ...buildResponseBytesTelemetry(responseBytes),
      ...callDiagnostics,
      ...deriveResourceIds(args, result),   // applied uniformly, last
  };
````

  Payoff:
  - Deletes 6 computeToolResponseBytes(...) call-site invocations and 2 { ...callDiagnostics, ...deriveResourceIds(...) } spreads →
  one internal site each.
  - Applies resource-id derivation to every path uniformly — the task-mode gap disappears for free.
  - args is already in scope at both callers (sync args; task closes over toolArgs), and every call site already passes the result
  into computeToolResponseBytes(result) — so they pass result instead. Net smaller call-site surface.

PR #984, opened 2026-06-11
fix(mcp): contain webServerMcpPath origin errors to one actor during tool load

MQ37:
## What
When loading MCP-server Actors as tools, resolve each Actor's standby MCP URL **inside** the per-Actor `try` in `getMCPServersAsTools`, so a failure is contained to that one Actor instead of rejecting the whole batch.

## Why
`getActorMCPServerURL` throws when an Actor's publisher-controlled `webServerMcpPath` resolves outside the Actor's standby origin (the cross-origin guard added in #927). It was called *before* the `try`, so the throw escaped the per-Actor isolation: a single Actor with a crafted `webServerMcpPath` rejected the `Promise.all`, failing tool loading for **every** Actor in the session — a denial-of-service with a trivial precondition (get one bad Actor into a session's tool set via add-actor/search). Moving the resolve inside the existing `try` degrades it to "skip this one Actor", which is how connection failures already behave.

## Testing
- New unit test `tools.mcp_servers_as_tools.test.ts`: two Actors, one whose path makes `getActorMCPServerURL` throw — asserts the valid Actor's tools still load. Verified it **fails** on the pre-fix code (whole batch rejects) and passes after.
- `pnpm type-check`, `pnpm lint`, `pnpm format:check`, `pnpm test:unit` (864 pass / 2 skipped) all green.

### DIFF
--- a/src/tools/core/actor_tools_factory.ts
+++ b/src/tools/core/actor_tools_factory.ts
@@ -211,19 +211,21 @@ export async function getMCPServersAsTools(
             return [];
         }
 
-        const mcpServerUrl = await getActorMCPServerURL(
-            actorInfo.definition.id, // Real ID of the Actor
-            actorInfo.webServerMcpPath,
-        );
-        log.debug('Retrieved MCP server URL for Actor', {
-            actorFullName: actorInfo.definition.actorFullName,
-            actorId,
-            mcpServerUrl,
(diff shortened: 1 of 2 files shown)


RobertCrupa (approved):

jirispilka (approved):
Thanks!

PR #985, opened 2026-06-11
fix(cache): scope Actor definition cache to its owner

MQ37:
> Mostly tests + comments — the actual logic change is small: a ~10-line ownership gate in `getActorDefinitionCached`, a simplified `getActorMcpUrlCached`, and deleting the `mcpServerCache` global.

## What
Gate `actorDefinitionCache` hits by ownership: public Actors are served to any caller, private Actors only to their owner; everyone else falls through to a fresh fetch under their own token. Drop the separate `mcpServerCache` and derive the MCP URL from the now-gated definition cache.

## Why
The cache is process-wide and was keyed by Actor name only. On a multi-tenant HTTP deployment, token B requesting an Actor that token A already cached got a cache **hit that bypassed B's authorization check** — leaking a private Actor's definition (input schema, README, run options) across tenants. `getActorMCPServerURL` does no I/O, so `mcpServerCache` guarded nothing expensive and re-introduced the same leak for the MCP URL; removing it closes that vector and deletes a global. Owner check reuses the already-cached `getUserInfoCached`, so no extra API call on the hot path. Found in an internal security review of the public package.

The gate hinges on two invariants (documented in `callerMaySeeCachedActor`): `info.userId` is the platform-set **owner**, and the caller's identity is resolved with the same `user('me')` identity the platform authorizes with. Verified against the **live Apify API**: a private Actor returns `isPublic: false` and `userId` equal to its owner's `user('me').id`; a public Actor returns `isPublic: true`. So a non-owner token resolves to a different `user('me').id`, the predicate denies, and the cache can never grant more than a bare re-fetch would.

## Testing
- New `tests/unit/utils.actor.cache.test.ts` (7 cases) pinning the authorization truth table: public→served to all (asserts no `user('me')` lookup), private→owner served, private→non-owner returns the **re-fetched** object (never the cached one), private→anonymous denied, and the `getActorMcpUrlCached` gate (owner→URL, non-owner private→`false`/re-fetch, missing→`false` no throw).
- `pnpm type-check`, `pnpm lint`, `pnpm format:check`, `pnpm test:unit` (870 pass / 2 skipped) all green.

> **Residual risk (accepted):** an Actor cached while public, then flipped private by its owner, is still served from the public fast-path until the 30-min TTL expires. The data was world-readable at cache time and the window is bounded; closing it would require per-request visibility re-checks that defeat the cache.
>
> **Note:** a private Actor legitimately shared with another user re-fetches each call (cache miss) rather than serving from cache — safe, slightly less efficient for that uncommon case.


RobertCrupa (approved):

src/utils/actor.ts
```diff
@@ -1,9 +1,26 @@
...
+/**
+ * `actorDefinitionCache` is process-wide, so a private Actor's definition must never be served from it to
+ * anyone but its owner — else another token on the same process reads it with no auth check. Two invariants
+ * keep this gate from inverting into a leak:
+ *   1. `info.userId` is the platform-set OWNER, not the fetching token — so a non-owner's re-fetch can't
+ *      overwrite the cached ownership.
+ *   2. The caller is identified by `user('me')` under their own token (the same identity the platform
+ *      authorizes with) and `null` is the sole non-identity sentinel — so a hit grants no more than a bare
+ *      re-fetch would. Don't drop the `!== null` guard or swap in a cheaper identity source.
+ */
```
jirispilka, line 22:
```suggestion
 *  Trade-off: an org-owned private Actor is cached under the org's userId, so an org member 
 *  calling with a personal token never matches and re-fetches every time. 
 *  Fail-safe (no leak), just uncached for members - accepted over an org-membership lookup
 *   that would put a per-call API round trip back on * this path.
 */
```

I would document org-member access defeats the cache for org-owned private Actors.

jirispilka (approved):
Nice PR.

As Claude puts it: This is a textbook "delete the layer" move, not a "polish the layer" move

PR #986, opened 2026-06-12
feat: Drop raw schema from get-dataset; stop nudging schema tool

### LINKED ISSUE #882: design: Calibrate get-dataset-schema output detail
## Problem
`get-dataset-schema` (`generateSchemaFromItems` over 5 items) returns a thin schema; raw Apify `dataset.schema` is too bloated. No agreed token budget for what a schema response should look like.

## Proposed solution
Open investigation:
- derive-from-items vs. trim `dataset.schema` vs. hybrid
- evals to measure agent success vs. schema verbosity
- pick a budget, land the strategy

## Probe finding — same problem in get-dataset

`get-dataset` also dumps the raw Apify `dataset.schema` field — the exact bloated structure this issue is calibrating away from. Whatever trim/inference strategy lands for `get-dataset-schema` should also apply when surfacing schema info via `get-dataset` (or `get-dataset` should stop returning the schema entirely and defer to `get-dataset-schema`).

Parent: #784

jirispilka:
Closes #882

## What

- **`get-dataset` stops returning the raw Apify `schema`.** The flat `fields` list it already returns is the complete, projection-ready field inventory.
- **Remove the nudges that steer the LLM into `get-dataset-schema`** (a context-heavy call): dropped from `get-dataset`'s description, and from `get-dataset-items`' last-page `nextStep` (now points at `get-dataset` for the field list). This keeps the loaded-tool gating added in #1007 — it just redirects the gated reference from `get-dataset-schema` to `get-dataset`.
- **`get-dataset-schema` is unchanged** — still available on demand, returns the full inferred schema (no depth cap).

## Why

Two findings drove this:

- **Probe of the 10 top store Actors** ([#882 comment](https://github.com/apify/apify-mcp-server/issues/882#issuecomment-4685753613)): the raw `dataset.schema` is 93–95% of the `get-dataset` response bytes on top Actors and declares 23–39% phantom fields (disabled add-ons, etc.). Meanwhile the flat `fields` list already carries every projectable path, including deep ones (`media.videoDeliveryLegacyFields.dash_manifest_url`), and is exactly the format `get-dataset-items` `fields=` consumes.
- **Mixpanel: `get-dataset-schema` is rarely called.** So the win isn't capping its output — it's (a) dropping the redundant raw schema from the tool people *do* use, and (b) not steering the model into a heavy call it seldom needs. `fields` is the one canonical description of dataset shape.

The tradeoff: the model no longer sees types/formats up front. It learns them by reading an item (which it does anyway), or by calling `get-dataset-schema` explicitly.

## Effect

`get-dataset` on a `compass/crawler-google-places` dataset: **41 KB → 2.9 KB**, key set identical minus `schema`.

## Tests

- Unit: `get-dataset-items` terminal `nextStep` points at `get-dataset` (the #1007 gating tests updated accordingly); `get-dataset` field normalization unchanged. Full suite: 968 passed.
- E2e via mcpc on a fresh dataset (the authoritative check for the schema drop): `get-dataset` has no `schema` key and `nextStep` mentions only items; `get-dataset-items` routes to `get-dataset`; `get-dataset-schema` still returns the full schema on demand.

## Not in scope

For wide datasets the flat `fields` list is itself large (~307 paths for a Facebook-posts dataset). Capping or making `fields` navigable is a separate question — worth its own issue.

## Note for apify-mcp-server-internal

`get-dataset` `structuredContent` no longer carries the `schema` key — contract suite may need a matching update. Not a tool removal (so not breaking in that sense), but the response shape changed.

https://claude.ai/code/session_01Sf9wACoa9h9y2m2WZ2Sde5

### DIFF
--- a/src/tools/storage/get_dataset.ts
+++ b/src/tools/storage/get_dataset.ts
@@ -25,7 +25,7 @@ export const getDataset: ToolEntry = Object.freeze({
     title: 'Get dataset',
     description: dedent`
         Get metadata for a dataset (collection of structured data created by an Actor run).
-        The results will include dataset details such as itemCount, schema, fields, and stats.
+        The results will include dataset details such as itemCount, fields, and stats.
         Use fields to understand structure for filtering with ${HelperTools.DATASET_GET_ITEMS}.
         stats.inflatedBytes (when present) is the approximate uncompressed byte size — use it with itemCount to pick a safe limit and fields before fetching.
         Note: itemCount updates may be delayed by up to ~5 seconds.
@@ -56,6 +56,10 @@ export const getDataset: ToolEntry = Object.freeze({
             return buildStorageNotFound(`Dataset '${datasetId}' not found.`);
         }
         const linkContext = await getConsoleLinkContext(apifyToken, client);
+        // The API also returns a raw `schema` (untyped in apify-client). It is 93–95% of the
+        // response bytes on top store Actors and declares fields that may be absent from the
+        // data, so drop it — get-dataset-schema infers a compact schema from real items (#882).
+        const { schema, ...metadata } = dataset as typeof dataset & { schema?: unknown };
         // Apify returns `fields` slash-separated AND with array indices expanded
         // (e.g. `latestComments/0/owner/username`). For a real Instagram-scraper
         // dataset this inflates ~78 schema fields into 528 paths (~85% bloat) and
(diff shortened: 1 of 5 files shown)


MQ37 (approved):
Great!

RobertCrupa (approved):

PR #987, opened 2026-06-12
feat: Surface multiple datasets, kv-stores

### LINKED ISSUE #912: feat: Surface aliased storages in run-response
## Problem
`ActorRun.storageIds` (apify-client) exposes aliased dataset / KV-store IDs per run, but the MCP server's run-response builder only emits `default` (`src/tools/core/actor_run_response.ts:587, 701`). Agents can't discover aliased storage IDs through MCP.

## Proposed solution
Iterate `actorRun.storageIds.datasets` and `keyValueStores` and emit one entry per alias. `default` keeps its metadata enrichment; aliased entries ship with `{ id }` only. The `RunStorages` type already allows the shape.

Parent: #784

jirispilka:
Closes #912.

Surface aliased (non-default) storages from `run.storageIds` in actor run responses, alongside the default entries. Every entry — default and aliases — is enriched with fetched metadata (dataset item counts/fields, KV keys).

The code handles datasets and key-value stores generically. Note: apify-core currently populates aliases for **datasets only** — key-value stores and request queues stay single-`default` at the platform level — so KV/RQ aliases will surface automatically once core adds them.

Output: 
```
  "storages": {
    "datasets": {
      "default": {
        "id": "sSkOWcRuvGoKU5pEn",
        "itemCount": 1,
        "fields": [ "isSumPrime", "firstNumber", "secondNumber",      .....
        ]
      },
      "books": {
        "id": "9qh1CQefaLzRdNqM3",
        "itemCount": 3,
        "fields": ["title", "author",    ....
        ]
      }
    },
    "keyValueStores": {
      "default": {
        "id": "6XPkBgHJgm5BijGjV",
        "keys": [  "COVER",   ....     ],
        "keyCount": 5
      }
    }
  },
```

## Non-obvious design decisions

**`inflatedBytes=0` is suppressed intentionally.**
The platform reports `inflatedBytes: 0` as a "not yet populated" sentinel, not a literal empty dataset. The field is omitted from responses when the value is `0`.

**Dataset metadata is fetched on every poll, not just terminal runs.**
This surfaces partial progress (`itemCount`) while the run is still executing. KV `listKeys` is terminal-only because non-terminal KV summaries are not useful — records aren't final until the run completes.

**Per-alias lag-fallback probes run in parallel.**
Each alias independently executes `resolveItemCountWithLagFallback`. The probe for alias B starts as soon as alias B's own metadata arrives, not after alias A finishes. This is safe: each probe only reads its own dataset and has no shared mutable state.

**`cleanItemCount` removed.**
The field was part of the apify-client `Dataset` type but was never surfaced in any actor run response and its semantics were unclear. All references (doc comment + three test fixtures) dropped.


src/tools/structured_output_schemas.ts
MQ37, line 1: nice simplification 😎

src/tools/core/actor_run_response.ts
```diff
@@ -775,60 +826,55 @@ export async function fetchActorRunData(params: {
...
+                );
+                return [alias, buildRunDataset(id, meta, resolvedItemCount)] as const;
+            }),
+        ),
+        Promise.all(
+            Object.entries(kvIds).map(async ([alias, id]) => {
+                const keys = isTerminal ? await fetchKvKeys(client, id, mcpSessionId) : null;
+                return [alias, buildRunKeyValueStore(id, keys)] as const;
+            }),
+        ),
```
MQ37: nit: this is quite complex :(

my clanker is suggesting this:
 ```ts                                                                                                                                                                                   
   /**                                                                                                                                                                                   
    * alias → entry map for one storage type. Fetches run in parallel across aliases; callers run the                                                                                    
    * two storage types in parallel too, so total latency is one fetch wave regardless of alias count.                                                                                   
    */                                                                                                                                                                                   
   async function buildStorageMap<T>(                                                                                                                                                    
       ids: Record<string, string>,                                                                                                                                                      
       buildEntry: (id: string) => Promise<T>,                                                                                                                                           
   ): Promise<Record<string, T>> {                                                                                                                                                       
       const pairs = await Promise.all(                                                                                                                                                  
           Object.entries(ids).map(async ([alias, id]) => [alias, await buildEntry(id)] as const),                                                                                       
       );                                                                                                                                                                                
       return Object.fromEntries(pairs);                                                                                                                                                 
   }                                                                                                                                                                                     
 ```                                                                                                                                                                                     
                                                                                                                                                                                         
 Call site becomes two named, symmetric calls — the as const / fromEntries dance is gone, and alias (unused in the bodies) disappears from the callbacks:                                
                                                                                                                                                                                         
 ```ts                                                                                                                                                                                   
   const [datasetMap, kvMap] = await Promise.all([                                                                                                                                       
       buildStorageMap(datasetIds, async (id) => {                                                                                                                                       
           const meta = await fetchDatasetMeta(client, id, mcpSessionId);                                                                                                                
           const itemCount = await resolveItemCountWithLagFallback(                                                                                                                      
               client, id, meta, isSucceeded, waitSecs, mcpSessionId, abortSignal,                                                                                                       
           );                                                                                                                                                                            
           return buildRunDataset(id, meta, itemCount);                                                                                                                                  
       }),                                                                                                                                                                               
       buildStorageMap(kvIds, async (id) =>                                                                                                                                              
           buildRunKeyValueStore(id, isTerminal ? await fetchKvKeys(client, id, mcpSessionId) : null),                                                                                   
       ),                                                                                                                                                                                
   ]);                                                                                                                                                                                   
                                                                                                                                                                                         
   const datasets = buildStorageEntries(datasetMap);                                                                                                                                     
   const keyValueStores = buildStorageEntries(kvMap);                                                                                                                                    
 ```

MQ37 (approved):
Cool, LGTM! 😎

src/tools/core/actor_run_response.ts
jirispilka: Thanks! So I did mostly as suggested. But I also simplified the call function by removing `meta`, `mcpSessionId` so that it becomes less bloated

PR #989, opened 2026-06-12
fix: Classify MCP protocol errors as client faults, not server errors

jirispilka:
Human

There were a couple of `level:ERROR` entries in Mezmo that aren't server faults.
This reclassifies them to `softFail` and gives the `concurrent-run-limit` a friendly user message.

I'm sorry for a larger blast-radius :) but I had to move a couple of things around to have the code at a right place.

### The Mezmo errors and why they aren't bugs

| Entry | Cause |
|---|---|
| `Parse error: Invalid JSON-RPC message` | client sent a malformed body |
| `Bad Request: Server not initialized` | request before `initialize`, or a stale/evicted session on a multi-node pod |
| `Only one initialization request is allowed` | client sent a second `initialize` |
| `MCP error -32001: Request timed out` | already soft, but the literal "error" in the value re-promoted it |
| `cannot-start-actor-runs … concurrent Actor runs` | user hit their plan's run limit — a billing condition |

### Non-obvious notes

- **Mezmo promotes any log line containing the lowercase word `error` to ERROR.** So `mcpErrorCode` / `INTERNAL_ERROR` (capitalized) are safe, but `MCP error -32001` is not — `sanitizeMezmoMessage` strips lowercase `error` only (the old `/ error:/gi` missed it: no trailing colon).
- **The run-limit reaches us two ways, both handled:** a direct run throws an `ApifyApiError` with `type: cannot-start-actor-runs` (the token is on `.type`, not in the message); a remote MCP-server Actor wraps it as a 500 whose body carries the token. `isActorRunLimitError` checks both.
- **It can surface as a 402**, so it's explicitly excluded from the x402 "payment required" branches (`isX402PaymentRequiredError`) — otherwise it would be mis-rendered as a payment challenge.
- **One predicate is the single source of truth** across log level, tool status (`SOFT_FAIL`), failure category (`INVALID_INPUT`), and the user message — so the 402-direct and 500-wrapped shapes are classified identically.

### Refactor (not strictly required)

Error-classification predicates were scattered (`isActorRunLimitError` in `logging.ts`, `isMemoryQuotaError`/`isPermissionApprovalError` in `call_actor_common.ts`). Collected them into a new leaf module `src/utils/apify_errors.ts` (imports only `const` + `apify-client`, so every consumer shares them with no import cycles). `logging.ts` is logging-only again.

### Not fixed here (needs triage)

- Remote-Actor `socket hang up` 500s (`lulzasaur/marketplace-search-mcp` and others) — external dependency failures.
- `red.cars/email-intelligence-mcp` ZodError — the Actor returns tools without `inputSchema`, violating the MCP spec.

https://claude.ai/code/session_01AUFhPCs7eeSTSoGvHqMccH


src/utils/logging.ts
```diff
@@ -30,6 +32,49 @@ export function getHttpStatusCode(error: unknown): number | undefined {
...
+        'No connection established',
+        'Only one SSE stream',
+        'Controller is already closed',
+        'Not Acceptable: Client must accept',
+        'Parse error',
+        'Server not initialized',
+        'Only one initialization request',
+    ].join('|'),
+    'i',
+);
```
MQ37: change: I know we had string matching like this before but I think we should make it less generic as it can possibly match some other libraries errors.

I went through the MCP SDK version 1.29.0 we are using and these are the exact error messages with dynamic prefixes:
```ts
   // Pinned to @modelcontextprotocol/sdk@1.29.0 — server/webStandardStreamableHttp.js.                                                                                                  
   // The SDK calls onerror with a bare `new Error(message)` (no JSON-RPC code, no HTTP                                                                                                  
   // status — those go only into createJsonErrorResponse()), so message text is the only                                                                                                
   // signal here. Re-verify these literals on every SDK bump (the guard test below fails loudly).                                                                                       
   const MCP_CLIENT_FAULT_MESSAGES: ReadonlySet<string> = new Set([                                                                                                                      
       'Bad Request: Server not initialized',                                                                                                                                            
       'Invalid Request: Only one initialization request is allowed',                                                                                                                    
       'Not Acceptable: Client must accept text/event-stream',                                                                                                                           
       'Not Acceptable: Client must accept both application/json and text/event-stream',                                                                                                 
       'Parse error: Invalid JSON',                                                                                                                                                      
       'Parse error: Invalid JSON-RPC message',                                                                                                                                          
       'Conflict: Only one SSE stream is allowed per session',                                                                                                                           
       'Not connected',                                                                                                                                                                  
   ]);                                                                                                                                                                                   
                                                                                                                                                                                         
   // Transport/runtime disconnects with variable tails — anchored, not substring-anywhere.                                                                                              
   const MCP_CLIENT_FAULT_PREFIXES: readonly string[] = [                                                                                                                                
       'No connection established for request ID:',          // webStandardStreamableHttp.js:699                                                                                         
       'Failed to send response: Error: Not connected',      // send-path wrap                                                                                                           
       'Invalid state: Controller is already closed',        // Node web-streams ERR_INVALID_STATE                                                                                       
   ];                                                                                                                                                                                    
                                                                                                                                                                                         
   export function isMcpClientFaultMessage(message: string): boolean {                                                                                                                   
       return MCP_CLIENT_FAULT_MESSAGES.has(message)                                                                                                                                     
           || MCP_CLIENT_FAULT_PREFIXES.some((p) => message.startsWith(p));                                                                                                              
   }
```

MQ37 (approved):
Nice, in general looks good - pre-approving! 

But I would make the error more focused and match the exact strings from the MCP SDK when possibly to prevent accidentally catching other libraries errors.

src/utils/logging.ts
jirispilka: Done in 321d480, applied your suggestion. `isMcpClientFaultMessage` now matches an exact-literal `Set` of the SDK 1.29.0 messages plus `startsWith()` for the three variable-tail disconnects. No substring-anywhere matching. Added guard tests that assert the literals match and near-misses (e.g. `Parse error: Invalid YAML`, `Database connection: Not connected to replica`) do not.

PR #991, opened 2026-06-12
fix(ajv): prevent ReDoS and compile-time DoS from untrusted schemas

MQ37:
## What
Two small changes to the shared AJV usage that close a confirmed denial-of-service from untrusted schemas:
1. Disable the `pattern` / `patternProperties` / `format` keywords on the shared validator (`src/utils/ajv.ts`).
2. Cap untrusted schema size in `fixedAjvCompile` before AJV codegen runs (`src/tools/utils.ts`).

## Why
A malicious Store Actor (loadable via `add-actor` / search) or a proxied Actor-MCP server controls the input schema we feed to AJV. Two synchronous, uninterruptible costs result, and Node is single-threaded — on hosted `mcp.apify.com` either one stalls every concurrent session:
- **ReDoS**: a `pattern` regex with catastrophic backtracking. Measured **12.7 s** event-loop freeze from a 91-byte schema + a 31-char input.
- **Compile blowup**: a nested-combinator schema (proxy path; arbitrary JSON Schema). Measured **4.7 s** synchronous `compile()`.

`removeKeyword` is position-aware (AJV knows keyword vs. data) so a field literally named `pattern` is untouched. No `src/` schema uses these keywords, and this layer only sanitizes LLM args — the Actor re-validates its real input on the run — so dropping regex enforcement loses no protection. The byte cap is 2× the largest real Actor input schema observed (~31 KB); both untrusted callers route through `fixedAjvCompile` while the trusted `compileSchema()` path stays uncapped. A rejected schema throws and is already caught fail-closed at both call sites (`actor_tools_factory` skips the tool; `mcp/proxy` drops that server's tool list).

## Testing
`pnpm type-check`, `pnpm lint`, `pnpm format:check`, `pnpm test:unit` — 868 pass. Added 5 regression tests: oversized schema throws before codegen; `pattern`/`patternProperties`/`format` no longer enforced. Reproduced the original 12.7 s freeze against the repo's ajv and confirmed it drops to ~0.06 ms after the change.

### DIFF
--- a/src/tools/utils.ts
+++ b/src/tools/utils.ts
@@ -69,8 +69,17 @@ export function getToolSchemaID(actorName: string): string {
     return `https://apify.com/mcp/${actorNameToToolName(actorName)}/schema.json`;
 }
 
+// Largest real Apify Actor input schema measured ~31 KB; cap at 2× to bound AJV's synchronous
+// codegen so a pathological untrusted schema can't freeze the event loop. Only the untrusted
+// callers (actor_tools_factory, mcp/proxy) reach AJV through here; trusted compileSchema() is
+// intentionally uncapped.
+const MAX_UNTRUSTED_SCHEMA_BYTES = 65_536;
+
 // source https://github.com/ajv-validator/ajv/issues/1413#issuecomment-867064234
 export function fixedAjvCompile(ajvInstance: Ajv, schema: object): ValidateFunction<unknown> {
+    if (Buffer.byteLength(JSON.stringify(schema)) > MAX_UNTRUSTED_SCHEMA_BYTES) {
+        throw new Error(`Input schema exceeds ${MAX_UNTRUSTED_SCHEMA_BYTES}-byte safety limit`);
+    }
     const validate = ajvInstance.compile(schema);
     ajvInstance.removeSchema(schema);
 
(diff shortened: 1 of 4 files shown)


src/tools/utils.ts
```diff
@@ -69,8 +69,17 @@ export function getToolSchemaID(actorName: string): string {
     return `https://apify.com/mcp/${actorNameToToolName(actorName)}/schema.json`;
 }
 
+// Largest real Apify Actor input schema measured ~31 KB; cap at 2× to bound AJV's synchronous
+// codegen so a pathological untrusted schema can't freeze the event loop. Only the untrusted
+// callers (actor_tools_factory, mcp/proxy) reach AJV through here; trusted compileSchema() is
+// intentionally uncapped.
+const MAX_UNTRUSTED_SCHEMA_BYTES = 65_536;
```
jirispilka, line 76: I get your point but still I would like to verify this with apify-core.

The limit should be there, no in the mcp server

jirispilka (approved):

src/tools/utils.ts
MQ37: see https://apify.slack.com/archives/C0L33UM7Z/p1781280165107159

PR #999, opened 2026-06-15
fix(ajv): raise schema cap to 256 KB and soft-log oversized-schema errors

MQ37:
## What
Raise `MAX_UNTRUSTED_SCHEMA_BYTES` from 64 KB to 256 KB, and when a schema still exceeds the cap, throw a typed `SchemaTooLargeError` that the central `logHttpError` classifier logs as a soft fail instead of an error — mirroring how MCP SDK client faults and the run-limit condition are handled. The tool is still dropped (fail-closed), it just no longer raises an error-level alert.

## Why
Real Store Actors hit the old 64 KB cap and were dropped with error-level `Failed to compile schema` logs on `mcp.apify.com` (e.g. `websift/seek-job-scraper` 75 KB, `radeance/wellfound-job-listings-scraper` 142 KB post-transform). The cap only bounds AJV's synchronous codegen (event-loop freeze) — it is not a server fault, so it shouldn't page like one. 256 KB clears all observed real schemas while still bounding codegen; the soft classification keeps a genuinely pathological (>256 KB) schema from alerting.

## Testing
`pnpm type-check`, `pnpm lint`, `pnpm format:check`, `pnpm test:unit` (932 pass). `fixedAjvCompile` test asserts it throws `SchemaTooLargeError`; added a `logHttpError` test asserting that error soft-fails (no `error`/`exception`). Verified against the two failing Actors above using the repo's real transform: both now fit under 256 KB.

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -2,6 +2,19 @@ export class TimeoutError extends Error {
     override readonly name = 'TimeoutError';
 }
 
+/**
+ * Thrown by `fixedAjvCompile` when an untrusted Actor / proxied-MCP input schema exceeds the byte
+ * cap that bounds AJV's synchronous codegen. It's a property of the schema, not a server fault, so
+ * `logHttpError` logs it as a soft fail and the caller drops just that one tool.
+ */
+export class SchemaTooLargeError extends Error {
+    override readonly name = 'SchemaTooLargeError';
+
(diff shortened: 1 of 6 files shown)


MQ37:
FYI @jirispilka @RobertCrupa @vojtechj-apify

jirispilka (approved):

PR #1001, opened 2026-06-16
refactor: guard dev server DELETE handler and fix port parsing

jirispilka:
Minor fixes + cleanup to the dev/standby test server (`src/dev_server.ts`); does not affect the hosted server.

- DELETE handler wrapped in try/catch (like POST/GET) — a rejection no longer crashes the process
- `PORT` falls back to 3001 for empty/non-numeric values
- error logged under the `errMessage` key; dead duplicate session-id injection removed; single-value enums inlined

### DIFF
--- a/src/dev_server.ts
+++ b/src/dev_server.ts
@@ -23,14 +23,6 @@ import { parseServerMode } from './utils/server_mode.js';
 // analytics. Still overridable by an explicit TELEMETRY_ENV (e.g. PROD) in the env.
 process.env.TELEMETRY_ENV ??= 'DEV';
 
-enum TransportType {
-    HTTP = 'HTTP',
-}
-
-enum Routes {
-    MCP = '/',
-}
-
 /**
  * Extracts the Apify API token from the incoming request.
  *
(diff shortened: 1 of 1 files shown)


src/dev_server.ts
```diff
@@ -126,11 +121,6 @@ export function createExpressApp(): express.Express {
             if (sessionId && transports[sessionId]) {
                 // Reuse existing transport
                 transport = transports[sessionId];
-                // Inject session ID into request params for existing sessions
-                if (req.body?.params) {
-                    req.body.params._meta ??= {};
-                    req.body.params._meta.mcpSessionId = sessionId;
-                }
```
jirispilka, line 133: This was duplicated, the session is inject below

src/dev_server.ts
```diff
@@ -23,14 +23,6 @@ import { parseServerMode } from './utils/server_mode.js';
...
 process.env.TELEMETRY_ENV ??= 'DEV';
 
-enum TransportType {
-    HTTP = 'HTTP',
-}
-
-enum Routes {
-    MCP = '/',
-}
-
```
jirispilka, line 33: This was really not worth keeping

MQ37 (approved):

Issue #998, opened 2026-06-15
[Feature]: Expose datasets and key-value stores as MCP resources


jancurn:
### Problem or motivation

MCP protocol supports resources, but we don't use them at all and delegate everything to MCP tools, which are not optimal for this. For example, MCP resources enable "subscribe" operation where MCP clients can listen to changes to a resource, so the MCP client can keep a localy up-to-date copy of remote files.

Also, having this would help us test resources implementation in mcpc.

### Proposed solution

Implement the MCP `resources/read` and later also `resources/subscribe`  for Apify datasets, key-value store records, and in fact every Apify API GET endpoint, re-using the existing API URLs as resource URIs like:

```
https://api.apify.com/v2/datasets/:datasetId/items?format=xlsx&clean=true
https://api.apify.com/v2/key-value-stores/:storeId/records/:recordKey
...
```

The reason why we should use existing URL rather than some new schema (e.g.  `apify://`) is that Actors and other MCP tools will return these URLs in their responses, so LLMs will not need to translate them. We just need to add information about this behavior to MCP server instructions, like "All Apify API GET requests can be fetched via MCP resources".

This way, we can provide a generic MCP resource support for **any** GET HTTP requests for Apify API. The MCP server will simply inject authentication and return the result from the API.

In version 2, we can also implement `resources/subscribe` generically too, by sending HEAD requests periodically to poll if `Last-Modified` changed and if so, send the subscribed MCP clients a notification. The problem is that Apify API doesn't implement `Last-Modified` all that much - we'll solve it elsewhere. To handle this gracefully, the MCP server can check if `Last-Modified` is available - if yes, the do the polling, if not, then just return an error that this resource cannot be subscribed to. 

Notes:

* API endpoints can return arbitrary content, so `resources/read` should preserve the record MIME type/content encoding.
* These resources can be returned from tools as `resource_link`s even if they are not all listed via `resources/list`. It's perfectly okay if `resources/list` returns nothing, but `resources/read` works like this.
* Once we have this, we can use the resource URIs also as part of tool responses, for clients that support it.

This is also cool for marketing.

jancurn:
For the record, some API GET requests on Apify API now set correctly Last-Modified, which can be used for resource subscriptions. We can make it generic - if GET endpoint returns Last-Modified, we enable subscription, and periodically send HEAD requests. If there's no Last-Modified, return an error.

jirispilka:
[...]

PR #1003, opened 2026-06-17
feat: Storage as resources

RobertCrupa:
Closes the first part of #998 — expose Apify storage (and any Apify API GET endpoint) via MCP `resources/read`.

## What this does

- Implements `resources/read` as a generic proxy over any Apify API GET URL, identified by its real `api.apify.com` URL — no custom scheme, so URLs that Actors and tools return can be read back verbatim.
- The apify-client injects the session token; reads are gated to the configured API origin (`isApifyApiUri`), so the token is never sent to another host.
- Bodies are returned by parsed type: JSON as text, text verbatim (preserving `Content-Type`), binaries as a base64 blob. Binaries over the inline limit (256 KB) link out as a `text/plain` message — for a key-value-store record this is the signed `recordPublicUrl`, fetchable without a token.
- Missing resources, bad tokens, and 5xx soft-fail as `text` rather than throwing.
- `resources/list` and `resources/templates/list` stay minimal (widgets / usage-guide only). Discovery is the prose in the server instructions, which document the URL shapes and that large datasets should be paged with `limit`/`offset`.

## Scope

`resources/read` only — this is the first part of #998. `resources/subscribe` is the deferred v2 item.

## Notes

- **Security / redirects:** reads are gated to the configured Apify API origin and reuse the shared `httpClient`, so they inherit the same redirect behavior as the existing storage tools (`get-dataset-items`, `get-key-value-store-record`) — no new surface. Follow-up: confirm the client strips `Authorization` on cross-origin redirects.
- **Tool integration deferred:** tools don't yet emit `api.apify.com` resource URIs in their responses; the "tool responses often include such URLs" line in the server instructions is forward-looking. Wiring tools to emit resource URIs is a follow-up (the "once we have this, we can…" item in #998).

## Testing

- Unit: `tests/unit/resources.api.test.ts`, `tests/unit/resources.service.test.ts`.
- Integration: `resources/read` for dataset items and a key-value-store record in `tests/integration/suite.ts`.

### End to end tests

Claude Code is able to use the resource interface so here are his results
image

## Related

- #1023 splits out an independent fix (preserve an empty key-value-store record value), kept out of this PR.


RobertCrupa:
Currently, large binaries are inlined in the resource interface. @jirispilka, we will need to decide whether to mirror the same approach with the tools, serving the URI when binaries are > 256Kb.

I am in favor of letting the LLM process the data from a resource link using a script, rather than dumping it into the context.

src/utils/server-instructions/index.ts
```diff
@@ -44,6 +44,11 @@ These tools are called **Actors**. They enable you to extract structured data fr
 ## Storage types
 - **Dataset:** Structured, append-only storage ideal for tabular or list data (e.g., scraped items).
 - **Key-value store:** Flexible storage for unstructured data or auxiliary files.
+
+## Apify API resources
+- Any Apify API GET endpoint can be read as an MCP resource. Pass the full \`https://api.apify.com/v2/...\` URL to resources/read; the server injects authentication and returns the response body.
```
jancurn:
```suggestion
- Any Apify API GET endpoint can be read as an MCP resource. Pass the full \`https://api.apify.com/v2/...\` URL to `resources/read`; the server injects authentication and returns the response body.
```

src/resources/api_resources.ts
```diff
@@ -0,0 +1,209 @@
...
+        for (const dataset of datasets.items) {
+            resources.push({
+                uri: `${base}/v2/datasets/${dataset.id}/items?limit=${DEFAULT_DATASET_ITEMS_LIMIT}`,
+                name: dataset.name ?? dataset.id,
+                description: `Dataset with ${dataset.itemCount} item(s).`,
+                mimeType: JSON_MIME_TYPE,
+            });
+        }
+    } catch {
+        // Ignore: best-effort listing.
```
jancurn: Perhaps at least log the error?

src/utils/server-instructions/index.ts
RobertCrupa: Done [3aa7d69](https://github.com/apify/apify-mcp-server/pull/1003/commits/3aa7d695f7cdc5b9d4738848a490d952b508287a)

src/resources/api_resources.ts
RobertCrupa: Done [35282cd](https://github.com/apify/apify-mcp-server/pull/1003/commits/35282cd179f0f2685d1de5de0a7b9e8dd0595920)

src/mcp/server.ts
```diff
@@ -633,19 +633,39 @@ export class ActorsMcpServer {
...
             paymentProvider: this.options.paymentProvider,
             getMode: () => this.serverMode,
             getAvailableWidgets: () => this.availableWidgets,
         });
 
+        // Resolve the token like the CallTool handler and build a client when one is present.
+        // Only resources/read is token-scoped (the API proxy needs auth); without a token reads soft-fail.
+        const resolveApifyClient = (params: ApifyRequestParams): ApifyClient | undefined => {
+            const token = this.resolveApifyToken(params._meta);
+            return token ? new ApifyClient({ token }) : undefined;
```
RobertCrupa: **[High] `resources/read` is unusable in payment (x402/Skyfire) mode.** `resolveApifyClient` only ever builds `new ApifyClient({ token })`. In payment mode there is no token (`dev_server.ts` sets `apifyToken: undefined` when a provider is present), so `resolveApifyToken` returns `undefined` → this returns `undefined` → `readApiResource` short-circuits to `"Cannot read …: no Apify token in this session."` for **every** Apify-API URL.

It also never forwards the `paymentHeaders` that the CallTool path attaches via `prepareToolCallContext` (`src/payments/helpers.ts:63-67` → `new ApifyClient({ paymentHeaders })`), so even billing-by-headers can't work here. The feature the server-instructions advertise is dead in the exact mode those instructions target. Either build the client the same way CallTool does (provider/payment headers + request headers) or scope/document that resources are token-only.

src/mcp/server.ts
```diff
@@ -633,19 +633,39 @@ export class ActorsMcpServer {
...
         const resourceService = createResourceService({
             paymentProvider: this.options.paymentProvider,
             getMode: () => this.serverMode,
             getAvailableWidgets: () => this.availableWidgets,
         });
 
+        // Resolve the token like the CallTool handler and build a client when one is present.
+        // Only resources/read is token-scoped (the API proxy needs auth); without a token reads soft-fail.
+        const resolveApifyClient = (params: ApifyRequestParams): ApifyClient | undefined => {
+            const token = this.resolveApifyToken(params._meta);
```
RobertCrupa: **[Medium — cross-repo] Hosted `_meta.apifyToken` may not be populated for `resources/read`.** Nothing in this repo writes `_meta.apifyToken` — both transports inject only `mcpSessionId` and pass the token via `options.token`. Per `CLAUDE.md`, per-request token injection lives in `apify-mcp-server-internal`. Until this PR only `tools/call` needed a token, so if the hosted transport injects `_meta.apifyToken` only on the CallTool path, hosted `resources/read` will fall back to `options.token` (often undefined) and soft-fail for authenticated users. Please confirm the internal transport injects the token for `ReadResource` requests too before relying on this in production.

src/resources/api_resources.ts
```diff
@@ -0,0 +1,140 @@
...
+        return {
+            contents: [
+                { uri, ...(mimeType && { mimeType }), blob: data.toString('base64') } satisfies BlobResourceContents,
+            ],
+        };
+    }
+
+    // JSON (already parsed to an object/array) or text/xml (a string). A string is emitted verbatim
+    // with its declared Content-Type; anything else is lossless-serialized as JSON.
+    const text = typeof data === 'string' ? data : JSON.stringify(data);
```
RobertCrupa: **[Medium] JSON primitive bodies lose fidelity.** Because the code branches on the parsed JS type rather than the MIME type, two cases corrupt:

- A body that is literal JSON `null` is parsed by apify-client to JS `null`, hits the `data === null` branch above, and is emitted as **empty text** — the `null` value is silently dropped and looks like an absent record.
- A body that is a bare JSON string (e.g. `"hello"`) is parsed to the JS string `hello`; `typeof data === 'string'` is true, so it's emitted verbatim as `hello` with `mimeType: application/json` — **invalid JSON** (quotes lost), and indistinguishable from a `text/plain` body.

Consider re-serializing with `JSON.stringify` when the declared Content-Type is JSON, so primitives round-trip.

src/resources/api_resources.ts
```diff
@@ -0,0 +1,140 @@
...
+ * `mimeType` to preserve a body's declared Content-Type (e.g. an empty record).
+ */
+function buildTextResult(uri: string, text: string, mimeType: string = TEXT_MIME_TYPE): ReadResourceResult {
+    return { contents: [{ uri, mimeType, text } satisfies TextResourceContents] };
+}
+
+/**
+ * Read any Apify API GET endpoint as an MCP resource.
+ *
+ * A thin proxy: the apify-client injects the session token (and MCP-origin / payment headers),
```
RobertCrupa: **[Medium — doc/code mismatch] This proxy does not inject payment headers.** The client it receives is built by `resolveApifyClient` as `new ApifyClient({ token })` — only the token and the always-on `X-Apify-Request-Origin: MCP` header are attached. Payment headers exist only on the `{ paymentHeaders }` client built in `src/payments/helpers.ts`, which this path never uses. The same claim appears in `src/resources/AGENTS.md`. Please correct both.

```suggestion
 * A thin proxy: the apify-client injects the session token (and the MCP-origin header),
```

src/resources/api_resources.ts
```diff
@@ -0,0 +1,140 @@
...
+        // limit link out instead: an explanatory text block with the URL, size, and type (resources/read
+        // has no resource_link content type), matching the soft-fail contract. For a key-value-store record
+        // the link is the signed public URL, fetchable without a token; other endpoints fall back to the
+        // (token-gated) API URL.
+        if (data.length > KV_RECORD_MAX_INLINE_BYTES) {
+            const downloadUrl = await getRecordDownloadUrl(uri, apifyClient);
+            return buildTextResult(
+                uri,
+                `Content (${mimeType ?? 'binary'}, ${data.length} bytes) is too large to inline. ` +
+                    `Fetch it directly from: ${downloadUrl}`,
```
RobertCrupa: **[Medium] The "Fetch it directly from" link is often not directly fetchable.** `getRecordPublicUrl` returns a **signed** URL only when the store has a `urlSigningSecretKey`; otherwise it returns an unsigned URL that still needs the `Authorization` token. And for any non-KV endpoint, `getRecordDownloadUrl` returns the original **token-gated API URL**. In both cases the message tells the model/user to fetch it directly with no auth, which yields a 401. Either signal that the link may require a token, or only present the "no token needed" wording when a signature is actually present.

src/resources/api_resources.ts
```diff
@@ -0,0 +1,140 @@
...
+    const contentType = typeof contentTypeHeader === 'string' ? contentTypeHeader : undefined;
+    const { data } = response;
+
+    // An empty body (e.g. an Actor that wrote an empty OUTPUT) is legitimate; emit empty text,
+    // preserving the record's declared Content-Type when present.
+    if (data === undefined || data === null) {
+        return buildTextResult(uri, '', contentType);
+    }
+
+    if (Buffer.isBuffer(data)) {
```
RobertCrupa: **[Medium — reuse] This binary branch duplicates `get_key_value_store_record.ts:86-139`.** The MIME-strip (`contentType?.split(';')[0].trim().toLowerCase()`), the `> KV_RECORD_MAX_INLINE_BYTES` link-out via `getRecordPublicUrl`, the base64 inline, and the `...(mimeType && { mimeType })` spread are all re-implemented here. Any change to the inline threshold, MIME normalization, or link-out policy now has to be made in two places and will drift. Extract a shared helper (e.g. in `storage_helpers.ts`) that takes `(contentType, buffer, uri)` and returns either an inline blob descriptor or a link-out, and call it from both.

src/resources/api_resources.ts
```diff
@@ -0,0 +1,140 @@
...
+        const mimeType = contentType?.split(';')[0].trim().toLowerCase();
+        // Inlining a large binary as base64 would blow up the client's context, so above the inline
+        // limit link out instead: an explanatory text block with the URL, size, and type (resources/read
+        // has no resource_link content type), matching the soft-fail contract. For a key-value-store record
+        // the link is the signed public URL, fetchable without a token; other endpoints fall back to the
+        // (token-gated) API URL.
+        if (data.length > KV_RECORD_MAX_INLINE_BYTES) {
```
RobertCrupa: **[Medium] The 256 KB inline cap only guards `Buffer` bodies; text/JSON is uncapped.** A large `datasets/{id}/items` read with no `limit` parses to a JS array and falls through to the text branch, which inlines the entire body — exactly the context blow-up the cap exists to prevent. The server-instructions say "page with limit/offset", but nothing enforces it and the asymmetry with the Buffer path is surprising. Consider applying a size guard to serialized text/JSON too (link-out or truncate-with-notice).

src/resources/api_resources.ts
```diff
@@ -0,0 +1,140 @@
...
+    const { data } = response;
+
+    // An empty body (e.g. an Actor that wrote an empty OUTPUT) is legitimate; emit empty text,
+    // preserving the record's declared Content-Type when present.
+    if (data === undefined || data === null) {
+        return buildTextResult(uri, '', contentType);
+    }
+
+    if (Buffer.isBuffer(data)) {
+        const mimeType = contentType?.split(';')[0].trim().toLowerCase();
```
RobertCrupa: **[Low] Two different normalizations of the same `Content-Type` within one function.** The blob branch strips parameters and lowercases (`contentType?.split(';')[0].trim().toLowerCase()`), but the text/JSON branch below emits the raw header verbatim (charset retained — the unit test asserts `text/plain; charset=utf-8`). Same response header, two output rules, no comment explaining why. Pick one (or document the divergence).

src/resources/api_resources.ts
```diff
@@ -0,0 +1,140 @@
...
+export function isApifyApiUri(uri: string): boolean {
+    try {
+        return new URL(uri).origin === new URL(getApifyAPIBaseUrl()).origin;
+    } catch {
+        return false;
+    }
+}
+
+/** Matches an Apify key-value-store record path, capturing the store id and the record key. */
+const KV_RECORD_PATH_RE = /^\/v2\/key-value-stores\/([^/]+)\/records\/(.+)$/;
```
RobertCrupa: **[Low] Greedy `(.+)$` and `decodeURIComponent` degrade the download link.** The record-key group captures any trailing query string (`…/records/INPUT?disableRedirect=true` → key `INPUT?disableRedirect=true`) and keys containing a literal `%` throw in `decodeURIComponent` — both end up in the `catch` and silently fall back to the token-gated API URL. `getRecordDownloadUrl` also parses `pathname` only, so any query on the original record URI is dropped from the minted URL. Minor (degraded link, not a crash), but worth tightening if record keys with `%`/queries are expected.

src/resources/api_resources.ts
```diff
@@ -0,0 +1,140 @@
...
+
+/**
+ * Download URL for a binary too large to inline. For a key-value-store record URI, returns the
+ * store's signed `recordPublicUrl` — fetchable without an API token when the client can read the
+ * store's URL signing key. Falls back to the original API URL for any other endpoint, or if minting
+ * the signed URL fails (fetching that link then needs a token).
+ */
+async function getRecordDownloadUrl(uri: string, apifyClient: ApifyClient): Promise<string> {
```
RobertCrupa: **[Low — naming] `getRecordDownloadUrl` performs a network call, so the `get` prefix violates the verb rule.** `getRecordPublicUrl` awaits `store.get()` (an HTTP round-trip to fetch store metadata for signing). `CONTRIBUTING.md:91-92`: "`get` — synchronous or cheap lookup from local state…; `fetch` — async retrieval from an external API or network call." Rename to `fetchRecordDownloadUrl`.

src/utils/server-instructions/index.ts
```diff
@@ -44,6 +44,12 @@ These tools are called **Actors**. They enable you to extract structured data fr
 ## Storage types
 - **Dataset:** Structured, append-only storage ideal for tabular or list data (e.g., scraped items).
 - **Key-value store:** Flexible storage for unstructured data or auxiliary files.
+
+## Apify API resources
+- Any Apify API GET endpoint can be read as an MCP resource. Pass the full \`https://api.apify.com/v2/...\` URL to \`resources/read\`; the server injects authentication and returns the response body.
```
RobertCrupa: **[Low — scope] This advertises *any* Apify API GET endpoint, and the gate is origin-only.** `isApifyApiUri` checks only the URL origin, so `/v2/users/me`, `/v2/schedules`, `/v2/webhooks`, billing, etc. are all readable as resources — much broader than "storage as resources". It's not a privilege escalation (the session token already authorizes those endpoints), but it surfaces account-wide data the feature didn't set out to expose. Consider a path allowlist, or confirm the broad surface is intended.

src/mcp/server.ts
RobertCrupa: We should open another issue for this and just leave it documented for now.

@jirispilka do you agree?

src/resources/api_resources.ts
RobertCrupa: Agreed

src/utils/server-instructions/index.ts
RobertCrupa: I think we can leave it as it for now

src/resources/api_resources.ts
RobertCrupa: This is in line with the tool endpoint so we will keep it this way for now

src/resources/api_resources.ts
RobertCrupa: Done [302c521](https://github.com/apify/apify-mcp-server/pull/1003/commits/302c521e1a74df7dbddba6dd2b8f5d20bdae3923)

src/resources/api_resources.ts
RobertCrupa: Done [0ac6550](https://github.com/apify/apify-mcp-server/pull/1003/commits/0ac6550c7cfb3cfe745af653ac447761f65925d8)

src/resources/api_resources.ts
RobertCrupa: Done [41ad83f](https://github.com/apify/apify-mcp-server/pull/1003/commits/41ad83f40ec00b2de6a280b9d7a012a54364638e)

src/resources/api_resources.ts
RobertCrupa: Done [d88a3fe](https://github.com/apify/apify-mcp-server/pull/1003/commits/d88a3feb7ff1ab7a66dec3ca1f8f1637b8ed0896)

src/resources/api_resources.ts
RobertCrupa: Done [af45dfd](https://github.com/apify/apify-mcp-server/pull/1003/commits/af45dfd650103835e843ce8abba115ef086909de)

src/resources/api_resources.ts
RobertCrupa: Not an issue - explained here [af45dfd](https://github.com/apify/apify-mcp-server/pull/1003/commits/af45dfd650103835e843ce8abba115ef086909de)

src/utils/server-instructions/index.ts
```diff
@@ -44,6 +44,12 @@ These tools are called **Actors**. They enable you to extract structured data fr
 ## Storage types
 - **Dataset:** Structured, append-only storage ideal for tabular or list data (e.g., scraped items).
 - **Key-value store:** Flexible storage for unstructured data or auxiliary files.
+
+## Apify API resources
+- Any Apify API GET endpoint can be read as an MCP resource. Pass the full \`https://api.apify.com/v2/...\` URL to \`resources/read\`; the server injects the session's Apify token and returns the response body. Reads require an Apify token — a session without one (e.g. payment-only x402/Skyfire) soft-fails with an explanatory message.
+- Actor and tool responses often include such URLs (e.g. dataset items, key-value store records) — read them directly via \`resources/read\`, no rewriting needed.
```
MQ37: > Actor and tool responses often include such URLs

I think this is not true right now as we are returning their IDs but not the readable URLs directly.

src/resources/api_resources.ts
```diff
@@ -0,0 +1,163 @@
...
+            ],
+        };
+    }
+
+    // A JSON body must be re-serialized so primitives round-trip as valid JSON: a bare JSON string
+    // `"hi"` parses to the JS string `hi`, and emitting it verbatim would drop the quotes; JSON `null`
+    // must stay `null`, not collapse to empty text. Branch on the declared Content-Type, not the parsed
+    // JS type, so the type alone can't misclassify a string body.
+    if (isJsonContentType(contentType)) {
+        return buildTextResult(uri, JSON.stringify(data), contentType ?? JSON_MIME_TYPE);
```
MQ37:
```suggestion
        return buildTextResult(uri, JSON.stringify(data), contentType);
```

I think we don't need the `?? JSON_MIME_TYPE`

src/resources/api_resources.ts
```diff
@@ -0,0 +1,163 @@
...
+            uri,
+            `Cannot read ${uri}: only Apify API URLs (${getApifyAPIBaseUrl()}) are readable as resources.`,
+        );
+    }
+
+    let response: { data: unknown; headers: Record<string, unknown> };
+    try {
+        // Default responseType is `arraybuffer`, which lets the client's parse interceptor decode
+        // the body by Content-Type. Do NOT set `forceBuffer` — that would keep everything as raw bytes.
+        response = await apifyClient.httpClient.call({ url: uri, method: 'GET', responseType: 'arraybuffer' });
```
MQ37: change: I think we should bound this to prevent reading possibly hundreds of MBs and thus possibly crashing the node instance. It can read the whole dataset that can be large, we could technically use `?limit` but it would be quite hacky and I think limiting what we pull is much safer.

src/tools/storage/storage_helpers.ts
```diff
@@ -164,3 +164,27 @@ const NORMALIZE_RECORD_KEY_REGEX = new RegExp(`^[${QUOTE_WRAPPER_CHARS}]+|[${QUO
...
+    | { kind: 'inline'; mimeType?: string; base64: string }
+    | { kind: 'linkOut'; mimeType?: string; bytes: number };
+
+/**
+ * Classify a binary record value for transport. Shared by the get-key-value-store-record tool and the
+ * API-resource proxy so the MIME normalization, inline threshold, and base64 encoding stay in one place.
+ * It does NOT mint the link-out URL — the two callers build different URLs (a signed record URL vs the
+ * API URL) via async calls — so the caller handles `linkOut` by minting its own URL.
+ */
+export function classifyBinaryRecord(contentType: string | undefined, value: Buffer): BinaryRecordDisposition {
```
MQ37: nit: I would rename it to something like `classifyBinaryRecordWeight` or `classifyBinaryRecordSize` so it is clear that it classifies based on the size of we should inline or not.

MQ37 (changes requested):
Just one major issue with not limiting what we are pulling in the `readApiResource`. Otherwise cool 😎 

But I think we should integrate it more into the MCP server, as we are not returning the resource links and not hinting the agent properly in the tool results. If we are introducing this feature I think we should return proper resource URLs in the tool results that return some kind of storage like the `call-actor` tool for example. So I would add also proper resource URL to the IDs and Console URLs that we are already returning.

src/resources/api_resources.ts
RobertCrupa: Agreed, I will have a look at what other options we have. We could possibly mirror the same approach I took to the KV records. So we have more consistency.

jirispilka (commented):
Thanks Robert.

I did a quick pass but I'll need a bit more time to look into that properly!

jancurn:
Just FYI, we'll update the Apify API to set `Last-Modified` only to Get dataset items and Get key-value store record API endpoints. The remaining will not use `Last-Modified` at all, so the subscribe/unsubscribe logic will not work for them, but it's okay imo

src/resources/api_resources.ts
RobertCrupa: Done, we now have a 5 MB limit [6e6d2c4](https://github.com/apify/apify-mcp-server/pull/1003/commits/6e6d2c493e392aee858b0b370821e6e05dbe1663)

src/tools/storage/storage_helpers.ts
RobertCrupa: Done: [35b4767](https://github.com/apify/apify-mcp-server/pull/1003/commits/35b47674eca4423659008de5ceb74543bfb90ce2)

src/resources/api_resources.ts
RobertCrupa: Done

src/utils/server-instructions/index.ts
RobertCrupa: It now reads:

Actor and tool results return storage IDs, not resource URLs — build the URL from the ID (e.g. a \`datasetId\` becomes \`https://api.apify.com/v2/datasets/{datasetId}/items\`) and read it via \`resources/read\`.

src/mcp/server.ts
MQ37: Confirmed — checked apify-mcp-server-internal directly (master @ 2314ed9):

- `enrichRequestBody()` in [`src/server/shared.ts#L224-L262`](https://github.com/apify/apify-mcp-server-internal/blob/2314ed9a4d7188bf13122cfc525bc1099e8366cf/src/server/shared.ts#L224-L262) injects `meta.mcpSessionId` for every request, but `meta.apifyToken` only inside `if (body.method === 'tools/call')` (L254-260). `resources/read` falls outside that branch, so `_meta.apifyToken` is never set for it.
- Both `ActorsMcpServer` construction sites in [`src/server/streamable.ts#L538`](https://github.com/apify/apify-mcp-server-internal/blob/2314ed9a4d7188bf13122cfc525bc1099e8366cf/src/server/streamable.ts#L538) (session-restore path) and [`#L637`](https://github.com/apify/apify-mcp-server-internal/blob/2314ed9a4d7188bf13122cfc525bc1099e8366cf/src/server/streamable.ts#L637) (fresh-session path) never pass a `token` option, so `options.token` is always `undefined` on the hosted server.

Net: with internal in its current state, hosted `resources/read` gets no token from either source and always soft-fails with "no Apify token in this session" — this is needed, not speculative. `enrichRequestBody` needs to inject `apifyToken` for `resources/read` too (or drop the method check and inject unconditionally, matching `mcpSessionId`).

MQ37: Fix opened: apify/apify-mcp-server-internal#652 — extends `enrichRequestBody` to inject `_meta.apifyToken` for `resources/read` too (was `tools/call`-only), with a unit test file for `enrichRequestBody` that didn't exist before.

MQ37 (approved):
Tested end-to-end against real Apify datasets/KVS records on the branch (local dev server + mcpc, see thread above) — dataset items, JSON/text/binary KV records, oversized link-out, 404, bad-origin, no-token all behave as designed. Approving.

Two non-blocking follow-ups from an agentic-UX pass, worth a fast-follow rather than blocking this PR:

**1. `resources/read` failures (404, no-token, bad-origin) are soft-failed as normal `contents` text, never a protocol-level JSON-RPC error.** The negotiated 2025-11-25 spec currently says servers **SHOULD** return `-32002`/`-32603` for these — a recommendation, not a hard violation today. But [SEP-2164](https://modelcontextprotocol.io/seps/2164-resource-not-found-error) (Status: Final) tightens this to **MUST** return `-32602` and **MUST NOT** return success-shaped content for a non-existent resource, once it lands in the next spec revision. Worth deciding deliberately now (and noting it in `src/resources/AGENTS.md`) rather than drifting further from where the spec is headed — a host that doesn't forward resource `contents` to the model (spec's own intended architecture: resource errors are host-caller-side, not model-visible) gets nothing at all from these soft-fails today.

**2. Unify the error text shape, and reuse `getToolCallErrorUserText`.** Catalogued the soft-fail messages in `readApiResource` — 7 distinct shapes, 4 different opening phrases (`Cannot read` / `Failed to read` / `Response body` / `Content (...)`), no shared prefix. The non-2xx branch also collapses 401/403/404/429/500 into one generic `Failed to read {uri}: HTTP {status}: {message}`, while the sibling `tools/call` path (`getToolCallErrorUserText`) already differentiates 401 ("check APIFY_TOKEN") and 403 ("may be private or token lacks access") for the exact same failure class. Since the model's only lever is the text in the response, reusing that helper here (and picking one prefix convention across all seven soft-fail sites) would make resources/read's failures as actionable as tools/call's already are.

jirispilka (commented):
I was discussing this issue with Fable and prompted it badly.
So instead of discussion, Fable went ahead and implemented it.
Here are two new PRs that are fixing the issues found #1094 and #1096

jirispilka (approved):
@RobertCrupa I think it should be fine now.

One more polish and we are good to go :) 

https://github.com/apify/apify-mcp-server/pull/1097

Once merged, please give it one round of testing 🙏🏻 

Pre-approving

RobertCrupa:
Last test before merge done https://claude.ai/code/artifact/c7b666f2-9c22-43ec-83ba-37f7c80054d0

PR #1006, opened 2026-06-19
fix(streamable-http): Correct session lifecycle

### LINKED ISSUE #918: test: Add raw-HTTP streamable-transport protocol-compliance suite
## Context

Our integration tests today all go through the official MCP SDK client. The SDK is a well-behaved client and intentionally hides three things our streamable-HTTP transport contract still needs assertions on. This issue tracks adding a narrow raw-HTTP suite for exactly those things.

Companion issue in `apify-mcp-server-internal`: apify/apify-mcp-server-internal#566 (covers the hosted-only slice — auth 401, Origin, Redis-backed session lookup).

## Why raw HTTP

The SDK hides:

1. **HTTP status codes** (200/202/400/404). The SDK surfaces JSON-RPC errors but not the codes; ops dashboards, load balancers, and the SDK's own session-recovery logic key off them.
2. **HTTP-only headers** (`Mcp-Session-Id`, future `MCP-Protocol-Version` enforcement). The SDK manages these internally; a client can't read them.
3. **Misbehavior the SDK refuses to do** — no session, malformed JSON. These come from broken proxies, fuzzers, browser dev tools, and other-language clients.

Anything inside the SDK's happy path (ping, capability errors, tool-not-found, pagination, structured output) stays SDK-mediated in `tests/integration/suite.ts`.

## Tests to add

**Initialization & session lifecycle**
- [ ] `initialize` returns 200 and an `Mcp-Session-Id` header (visible ASCII per spec)
- [ ] POST without `Mcp-Session-Id` (and not `initialize`) → 400
- [ ] POST with unknown `Mcp-Session-Id` → 404 so SDK recovery starts a new session
- [ ] DELETE on a valid session → 2xx; reusing the same session ID afterwards → 404

**HTTP envelope**
- [ ] Notification posts → HTTP 202 with empty body
- [ ] `notifications/cancelled` for an unknown `requestId` → 202 (silent no-op per spec)
- [ ] Malformed JSON body → 4xx

**Protocol-version negotiation**
- [ ] Subsequent requests with an unsupported `MCP-Protocol-Version` header → 400

## Scope boundary

Tests that depend on hosted-stack behavior (auth 401, `WWW-Authenticate`, foreign-origin 403, Redis-backed cross-node session lookup) live in `apify-mcp-server-internal` — see apify/apify-mcp-server-internal#566.

## Implementation notes

A small (~150 LOC) raw-HTTP helper plus a single new test file alongside `tests/integration/suite.ts`. A drafted client lives at apify/apify-mcp-server-internal#474 (closed) and can be ported as a starting point.

jstar0:
## Summary

- return `400` for non-initialize Streamable HTTP POST requests that omit `Mcp-Session-Id`
- remove Streamable HTTP sessions from the dev server maps when DELETE closes them
- add raw HTTP coverage for session lifecycle, notification envelope, malformed JSON, and protocol-version handling

## Changes

- Split the invalid Streamable HTTP POST branch so missing session IDs and unknown session IDs return the expected status codes.
- Register `onsessionclosed` for Streamable HTTP transports so DELETE removes the session from the dev server's local maps.
- Add a raw HTTP test that starts the Express app directly and verifies session lifecycle, notification, malformed JSON, and protocol-version behavior without going through the SDK client.

## Why

The raw HTTP contract distinguishes a missing session header from an unknown session:

- missing `Mcp-Session-Id` on non-initialize requests should be a bad request
- unknown or terminated sessions should be not found so clients can recover by initializing a new session

The current dev server kept a closed transport in its local session maps after DELETE, so the same session ID could be reused after termination.

## Verification

```bash
pnpm exec vitest run tests/unit/dev_server.streamable_http.test.ts
pnpm run type-check
pnpm run lint
pnpm run test:unit
pnpm run format
pnpm run format:check
pnpm run check:agents
```

Fixes #918


jirispilka:
Pushed a follow-up commit (54f7e41) to land this:

- **Kept** the `dev_server.ts` lifecycle fix — 400 for missing `Mcp-Session-Id`, 404 for unknown session, and `onsessionclosed` cleanup on DELETE.
- **Dropped** the raw-HTTP unit test: the hosted server already implements and tests this same lifecycle, so it's redundant here.
- **Added a `DEV ONLY` marker** to `dev_server.ts` — it's a local standby-emulation, not the production HTTP transport. Production HTTP behavior should be fixed in the hosted (internal) server, not here.

Thanks for the fix! 🙏

jirispilka (approved):

Issue #885, opened 2026-05-20
feat: Improve storage tools description


jirispilka:
## Problem
Storage-tool descriptions (`get-dataset`, `get-dataset-items`, `get-dataset-schema`, `get-user-datasets-list`, `get-key-value-store*`) drifted in style and clarity. Some are terse, some verbose, USAGE/USAGE EXAMPLES blocks inconsistent. Several concrete contradictions exist:

- `get_key_value_store_keys.ts:34` USAGE EXAMPLE says `List first 100 keys in store username~my-store`, but the schema caps `limit` at 10. An agent following the example gets an ajv rejection.
- `dataset_collection.ts:14` description says `Maximum number of array elements to return. The default value (as well as the maximum) is 20.`, but the schema sets `.default(10)`. Either align the description to "Default 10, max 20" or change the default to 20.
- `get_dataset_schema.ts:13` uses `z.number().optional().default(5)` and `z.boolean().optional().default(true)`. The `.optional()` before `.default()` is a no-op in Zod (defaults always fill in), but the generated JSON Schema marks the fields as not-required. Clients reading the schema may omit them expecting the upstream API default; Zod silently substitutes 5/true/'all'. `get_dataset_items.ts` uses genuine `.optional()` for `limit` — two storage tools advertise the same "optional" contract with different fallbacks.

## Proposed solution
Review and rewrite descriptions for the storage category — consistent voice, clear USAGE, cross-references between related tools, and fix the contradictions above.

Parent: #784

PR #1008, opened 2026-06-20
fix: Resolve dataset tool schema/description inconsistencies

surya16122114:
## What
Fixes the still-present schema/description contradictions in the storage tools from #885.

- **get-dataset-list** (`dataset_collection.ts`): the `limit` description said the default is `20`, but the schema is `.max(20).default(10)`. An agent following the description expects 20 and silently gets 10. Corrected the description to "The default value is 10, the maximum is 20."
- **get-dataset-schema** (`get_dataset_schema.ts`): `limit` and `clean` used `.optional().default(...)`. In Zod 4 `.default()` already makes the input optional, so `.optional()` is a no-op. Removed it.

## Why
These descriptions are LLM-facing — a wrong default misleads the model about what it gets back. The no-op `.optional()` is dead modifier noise and made the "optional contract" inconsistent with sibling tools (noted in #885).

## How / risk
- Description fix is text-only — no behavior change.
- Removing `.optional()` is behavior-preserving: with Zod 4 `.default()` already implies optional input, so the generated JSON Schema (`z.toJSONSchema`) and runtime parsing are identical. Added a unit test asserting `get-dataset-schema` still applies `limit=5, clean=true` when called with no params.

## Not included
The third bullet in #885 (`get_key_value_store_keys` USAGE example "100 keys" vs `limit` max 10) is **already consistent on master** — the example reads "10 keys" — so no change was needed. Kept minimal per CONTRIBUTING's scope discipline; the broader stylistic rewrite is intentionally left out to keep this reviewable.

## Verification
`type-check`, `lint`, `test:unit` (947 passed), `format`, `check:agents` — all green.

Refs #885

### DIFF
--- a/src/tools/common/dataset_collection.ts
+++ b/src/tools/common/dataset_collection.ts
@@ -16,7 +16,7 @@ const getUserDatasetsListArgs = z.object({
     limit: z
         .number()
         .max(20)
-        .describe('Maximum number of array elements to return. The default value (as well as the maximum) is 20.')
+        .describe('Maximum number of array elements to return. The default value is 10, the maximum is 20.')
         .default(10),
     desc: z
         .boolean()
--- a/src/tools/common/get_dataset_schema.ts
+++ b/src/tools/common/get_dataset_schema.ts
@@ -14,14 +14,9 @@ import { buildStorageNotFound, buildStorageResponse } from './storage_helpers.js
 
 const getDatasetSchemaArgs = z.object({
     datasetId: z.string().min(1).describe('Dataset ID or username~dataset-name.'),
-    limit: z
-        .number()
-        .optional()
-        .describe('Maximum number of items to use for schema generation. Default is 5.')
-        .default(5),
+    limit: z.number().describe('Maximum number of items to use for schema generation. Default is 5.').default(5),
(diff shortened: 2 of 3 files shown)


jirispilka:
Verified locally: type-check, lint, `test:unit` (947 passed), format, and check:agents all green on the branch. The description fix (`get-dataset-list`: "default 10, max 20") is the real win — it corrects what the model is told it gets back. The `.optional()` removal is a confirmed no-op: under Zod 4, `.optional().default(x)` and `.default(x)` produce the same `parse()` result and a semantically identical JSON Schema (only property key order differs), so runtime and validation are unchanged.

Two non-blocking notes:
- The body says the generated JSON Schema is "identical" — strictly it's *semantically* identical; the property key order differs. Harmless, just slightly overstated.
- The broader `required`-array inconsistency this touches is now tracked in #1016 (published `inputSchema` advertises `.default()` fields as required in 18/19 tools, while the AJV gate doesn't enforce it). #1016 supersedes the idiom-cleanup intent here, but doesn't conflict with this PR.

LGTM to merge as-is.

---

jirispilka (approved):

PR #1010, opened 2026-06-21
fix: Hints and abort-actor-run

### LINKED ISSUE #1007: [Bug]: Fix `abort-actor-run` and `get-dataset-items` from external review
### Server type

Hosted (mcp.apify.com)

### MCP Client

Claude.ai (web)

### Operating System

None

### What happened?

We got this feedback from external testing:

Image

BTW the OpenAI directory submission also didn't like the abort tools:

Image

### Steps to reproduce

N/A

### Node.js version

_No response_

### Apify MCP Server version

_No response_

### MCP server configuration

```json
N/A
```

### Error logs

```text

```

jirispilka:
Fix nextStep hints that named tools the caller may not have — or that don't match how the run was started.

- `get-actor-run` / `abort-actor-run` retry hints no longer name `call-actor`. These tools only have a `runId` and can't know whether the run came from `call-actor` or a native Actor tool, so the hint is now generic ("re-run the Actor").
- `get-dataset-items` names `get-dataset-schema` only when that tool is actually loaded; otherwise it falls back to inspecting the returned items directly.
- `abort-actor-run` now returns a structured `RunResponse` (with `outputSchema`) instead of a raw run dump — stops leaking internal fields (`userId`, `buildId`, `containerUrl`) and adds `summary`/`nextStep`.

Closes #1007

### DIFF
--- a/src/tools/common/abort_actor_run.ts
+++ b/src/tools/common/abort_actor_run.ts
@@ -44,18 +44,13 @@ USAGE EXAMPLES:
         openWorldHint: false,
     },
     call: async (toolArgs: InternalToolArgs) => {
-        const { args, apifyClient: client, apifyMcpServer } = toolArgs;
+        const { args, apifyClient: client } = toolArgs;
         const parsed = abortRunArgs.parse(args);
         const run = await client.run(parsed.runId).abort({ gracefully: parsed.gracefully });
 
         const dataset = run.defaultDatasetId ? { id: run.defaultDatasetId } : undefined;
         const keyValueStore = run.defaultKeyValueStoreId ? { id: run.defaultKeyValueStoreId } : undefined;
-        const { summary, nextStep } = buildStatusSummaryNextStep({
-            run,
-            dataset,
-            keyValueStore,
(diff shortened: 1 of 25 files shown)


src/tools/common/abort_actor_run.ts
```diff
@@ -30,6 +33,7 @@ USAGE EXAMPLES:
 - user_input: Abort run y2h7sK3Wc
 - user_input: Gracefully abort run y2h7sK3Wc`,
     inputSchema: z.toJSONSchema(abortRunArgs) as ToolInputSchema,
+    outputSchema: getActorRunOutputSchema,
```
MQ37: nit: I would rename the `getActorRunOutputSchema` since we are now using that here and not only in the get Actor run.

MQ37 (approved):

src/tools/common/abort_actor_run.ts
jirispilka: Good point, I renamed to `actorRunOutputSchema`, I was not able to figure out a better name

PR #1012, opened 2026-06-22
fix: Handle task expiration gracefully when storing results

### LINKED ISSUE #1011: fix: Expired task on storeTaskResult logs as ERROR, should be softFail
## Problem

Long-running task result storage logs an ERROR when the task's Redis entry has already expired:

```
ERROR executeToolAndUpdateTask failed unexpectedly
{"error":{"message":"Task with ID call-tool-call-actor-... not found or expired",
"name":"Error","stack":"Error: Task with ID ... not found or expired
    at RedisTaskStore.storeTaskResult (.../stores/redis-task-store.js:326:19)"},
"taskId":"call-tool-call-actor-..."}
```

This is an expected, benign condition — the task TTL (default 5 min, max 10 min) elapsed before the long-running actor run finished, so there's no task left to store the result against. It is not an unexpected internal failure, but it logs at ERROR with "failed unexpectedly" wording, which pollutes error monitoring (Mezmo promotes on "error").

## Cause

In `src/mcp/server.ts`, `executeToolAndUpdateTask` stores the result via `this.taskStore.storeTaskResult(...)`. When the task expired, the store throws `Task with ID ... not found or expired` (`storeTaskResult` reads the task back first and throws if missing).

That `Error` is a plain `Error` — no HTTP status, not an `McpError` — so `getToolStatusFromError` classifies it as `FAILED` (`src/utils/tool_status.ts:82-83`) and it escapes to the unconditional top-level handler:

```ts
.catch((error) => log.error('executeToolAndUpdateTask failed unexpectedly', { taskId, error })) // server.ts:1051
```

The `storeTaskResult` in the catch block also re-throws the same "not found or expired" error when storing the failed result, compounding the noise.

## Expected

Treat an expired/missing task on result storage as a soft fail (warn), not an ERROR. The same routing already exists for client-side/quota conditions (#679) — extend it to this case.

## Scope

- Public repo only (`src/mcp/server.ts`, possibly `src/utils/tool_status.ts`). The thrown message comes from `RedisTaskStore` in the internal repo, but the log-level decision is here.
- Add a failing test that reproduces the expired-task path first.

Related: #679, #680, #360.

jirispilka:
When a long-running task's TTL elapses before its result can be stored, the task store rejects the operation with a "task not found" error. This is a benign terminal condition (the client gave up), not an unexpected failure. Previously this escalated to an error log (`Error executing tool for task` + `executeToolAndUpdateTask failed unexpectedly`); now it soft-fails.

> Jun 20 11:23:30 apify-mcp-server-586585b6fb-8njwb apify-mcp-server ERROR {"error":{"message":"Task with ID call-tool-call-actor-195fbee6-39fb-422e-9cbe-f6a6b8ba349b not found or expired","name":"Error","stack":"Error: Task with ID call-tool-call-actor-195fbee6-39fb-422e-9cbe-f6a6b8ba349b not found or expired\n    at RedisTaskStore.storeTaskResult (file:///home/node/dist/stores/redis-task-store.js:326:19)\n    at process.processTicksAndRejections...[truncated]"},"msg":"executeToolAndUpdateTask failed unexpectedly","taskId":"call-tool-call-actor-195fbee6-39fb-422e-9cbe-f6a6b8ba349b"}

## Testing

**End-to-end with mcpc (claude):** verified against the running HTTP server.
- mcpc sends `task: {}` with no TTL, and `InMemoryTaskStore` only schedules deletion when a TTL is set — so the expiry can't be triggered through mcpc against the in-memory store. It's enforced by the hosted Redis store's TTL cap. To exercise the real guard, the hosted TTL was emulated locally (2s, throwaway, reverted) and driven via `mcpc tools-call call-actor ... --detach`.
- Task TTL 2s, actor run 9.7s → result storage hit the expired task. Server log:
  ```
  SFAIL  Task expired before its result could be stored {"taskId":"515ec2...","toolName":"call-actor"}
  INFO   Tool call completed {"taskId":"515ec2...","toolStatus":"SOFT_FAIL","durationMs":9761}
  ```
  Zero ERROR/EXCEPTION lines; server stayed healthy. Client `tasks-result` returned a clean `-32602 Task not found`.

**Note:** the real Redis-backed trigger lives in `apify-mcp-server-internal`; matching e2e coverage against `RedisTaskStore` belongs there, not here.

closes #1011

https://claude.ai/code/session_01EY4mYy88GUH31Ein1tX2XG


src/mcp/server.ts
```diff
@@ -1047,9 +1052,18 @@ export class ActorsMcpServer {
...
+                            // throw "task not found or expired" when the TTL elapsed. That is benign —
+                            // soft-fail instead of escalating to an error.
+                            if (isTaskNotFoundError(error)) {
+                                log.softFail('Task expired before its result could be stored', {
+                                    taskId: task.taskId,
+                                });
+                                return;
+                            }
+                            log.error('executeToolAndUpdateTask failed unexpectedly', { taskId: task.taskId, error });
+                        });
```
MQ37: question: why this guard does not finish the telemetry as the second one does?

MQ37 (approved):
One question otherwise LGTM! Nice!

src/mcp/server.ts
jirispilka: Good point, it would make code a bit more complicated, more refactoring. 
Anyway, I tasked Claude to do it, it is fixed now.

PR #1027, opened 2026-06-25
chore: Use DEFAULT_DATASET_ITEMS_LIMIT in nextStep templates

### LINKED ISSUE #826: chore: Use DEFAULT_DATASET_ITEMS_LIMIT in get-actor-run nextStep templates
## Problem

After #814 lands, `get-dataset-items` will default `limit` to `20` via a file-local `DEFAULT_DATASET_ITEMS_LIMIT` constant. The `nextStep` templates in `src/tools/core/actor_run_response.ts` still hardcode `limit=20` as a magic number in three places (lines ~244, 253, 285). If the default ever changes, the `nextStep` hints drift silently.

## Proposed solution

1. Export `DEFAULT_DATASET_ITEMS_LIMIT` from `src/tools/common/get_dataset_items.ts`.
2. Import it in `actor_run_response.ts` and interpolate into the three `nextStep` templates instead of the literal `20`.

Blocked by: #814 (don't start until it merges).

mvanhorn:
## Summary
The four `get-actor-run` `nextStep` hint templates in `src/tools/core/actor_run_response.ts` hardcoded the literal `20` (the `limit (for example 20)` phrasing). The real default lived in a separate file-local constant, `DEFAULT_DATASET_ITEMS_LIMIT`, in `src/tools/common/get_dataset_items.ts`. The two could drift silently. This points the hints at the single source of truth.

## Why this matters
Per #826, the `get-dataset-items` default (`DEFAULT_DATASET_ITEMS_LIMIT = 20`) and the example value shown in the run-response `nextStep` hints were maintained independently. If the default ever changed, the hints would keep telling callers `for example 20` while the tool actually used a different limit. The blocker named in the issue (#814) has merged, so the constant and `get_dataset_items.ts` already exist.

## Changes
- Export `DEFAULT_DATASET_ITEMS_LIMIT` from `src/tools/common/get_dataset_items.ts` (value unchanged).
- Import it into `src/tools/core/actor_run_response.ts` and interpolate `${DEFAULT_DATASET_ITEMS_LIMIT}` into the four `DATASET_GET_ITEMS` `nextStep` templates in place of the literal `20`.
- Add a drift-guard unit test asserting the dataset `nextStep` reads `limit (for example ${DEFAULT_DATASET_ITEMS_LIMIT})`, so changing the constant updates hint and test together.

No behavioral change: the rendered hints still read `for example 20` because the constant is `20`. No import cycle introduced (`type-check` passes).

## Testing
- `pnpm run type-check` - pass
- `pnpm run lint` - pass (0 warnings, 0 errors)
- `pnpm run format:check` - pass
- `pnpm run check:agents` - pass
- `pnpm run test:unit` - 963 passed, 2 skipped

Fixes #826

### DIFF
--- a/src/tools/actors/actor_run_response.ts
+++ b/src/tools/actors/actor_run_response.ts
@@ -22,6 +22,7 @@ import { logHttpError } from '../../utils/logging.js';
 import { buildMCPResponse } from '../../utils/mcp.js';
 import { formatRunStatusMessage, type ProgressTracker, TERMINAL_RUN_STATUSES } from '../../utils/progress.js';
 import { cleanEmptyProperties } from '../../utils/schema_generation.js';
+import { DEFAULT_DATASET_ITEMS_LIMIT } from '../storage/get_dataset_items.js';
 
 /** Cap on `storages.keyValueStores.default.keys` array length. */
 const KV_KEYS_LIMIT = 50;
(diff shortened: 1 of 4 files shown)


jirispilka (approved):
Thanks, I had to rebase master and also fix one or two issues. All good now!

MQ37 (approved):
Thank you sir 🙇

Issue #978, opened 2026-06-10
refactor: Regroup src/tools by domain (actors/runs/storage/docs/widgets)


jirispilka:
## Problem

`src/tools/` groups files by three inconsistent axes: `common/` (mode-independence), `core/` (shared logic — a synonym of "common" meaning the opposite here), and `default/`+`apps/` (a mode split that mostly doesn't exist — see #975). Helper file names mislead too: `build.ts` fetches Actor definitions, `core/*_common.ts` marks files as shared twice. The user-facing taxonomy already exists: the `toolCategories` registry keys (`actors`, `docs`, `runs`, `storage`). Directories should match it.

## Proposed solution

One mechanical move PR to the domain layout:

```
src/tools/
├── index.ts, registry.ts (was categories.ts)
├── actors/    search_actors, fetch_actor_details, call_actor, add_actor,
│              actor_tools_factory, actor_executor, actor_definition (was build.ts),
│              actor_run_response
├── runs/      get_actor_run, get_actor_run_log, abort_actor_run, get_actor_run_list
├── storage/   dataset + key-value-store tools, storage_helpers
├── docs/      search_apify_docs, fetch_apify_docs
└── widgets/   the 4 *-widget tools (was apps/)
```

- Dissolve `common/`, `core/`, `default/`, `apps/`. Shared-logic files move next to their tools, dropping the `_common` suffix (e.g. `core/call_actor_common.ts` → merges into or sits beside `actors/call_actor.ts`).
- `categories.ts` → `registry.ts`; one directory per category key, same vocabulary everywhere.
- `utils.ts` (431 lines) and `structured_output_schemas.ts` stay top-level; splitting `utils.ts` is out of scope.
- Pure `git mv` + import fixes, no logic changes — the only logic change (call-actor dedup) lands first in #975 so rename detection stays intact.

Blast radius checked: `package.json` exports only `.` and `./internals` — no subpath into `tools/`, so moves are internal-only. `index_internals.ts` import paths must be updated but its export surface stays identical.

**Timing:** this conflicts with every open PR touching `src/tools/` — land right after the in-flight storage PRs (#964, #965, #932) merge. If #972 (per-directory AGENTS.md) is merged by then, update `src/tools/AGENTS.md` in the same PR.

Depends on #975.

## Plan

- [ ] Create domain dirs, `git mv` files per the layout above (incl. `build.ts` → `actors/actor_definition.ts`, drop `_common` suffixes)
[...]

PR #1035, opened 2026-06-27
refactor: Merge each tool's *_common.ts into its tool-entry file

### LINKED ISSUE #1020: refactor: Merge each tool's *_common.ts logic into its tool-entry file
## Problem

After #978 regrouped `src/tools/` by domain, each of four tools is still split across two files in the same domain dir:

- `actors/call_actor.ts` (tool entry) + `actors/call_actor_common.ts` (shared logic)
- `actors/search_actors.ts` + `actors/search_actors_common.ts`
- `actors/fetch_actor_details.ts` + `actors/fetch_actor_details_common.ts`
- `runs/get_actor_run.ts` + `runs/get_actor_run_common.ts`

The `_common` suffix only ever distinguished the shared-logic module from its sibling tool-entry file when they lived in different directories (`default/` vs `core/`). Now that both sit in the same domain dir, the suffix is redundant noise. #978 deliberately left it alone because dropping it there would have been a logic/structure change inside a pure-`git mv` PR — both files would want the bare name (e.g. `actors/search_actors.ts`), so the only clash-free resolution is to merge them.

## Proposed solution

Merge each `*_common.ts` into its tool-entry file so there is one self-contained file per tool (e.g. fold `call_actor_common.ts` into `actors/call_actor.ts`), then delete the `_common` file. Update imports in consumers and tests.

- Behavior, exports actually consumed elsewhere, schemas, and the `HelperTools` enum stay unchanged — this is a structure-only consolidation.
- Some `*_common` exports are imported directly by other modules / tests (e.g. `mcp/server.ts` imports from `actors/call_actor_common.ts`; several unit tests import `*_common` symbols). Those import sites move to the merged file; re-exports may be needed to keep the surface stable.
- Coordinate with #977 (file/export-name normalization) — these touch the same files and likely want to land together or in a deliberate order to avoid moving the same code twice.

## Notes

Split out of #978 (the move PR kept `*_common.ts` as-is by design). Part of the #979 cleanup. Confirm no impact on `apify-mcp-server-internal` before landing, since it consumes this package.

jirispilka:
Closes #1020.

After #978 regrouped `src/tools/` into domain directories, each of four tools still kept a `*_common.ts` sibling next to its tool-entry file. The `_common` suffix only distinguished shared logic from the tool-entry file back when they lived in *different* directories (`default/` vs `core/`). Now that both sit in the same domain dir, the suffix is structural noise.

This is a bit controversial as some files might be quite large but I think it is better than having `_common.ts` suffix.

## How

Fold each `*_common.ts` into its tool-entry file so each tool is one self-contained file, delete the `_common` file, and repoint every import site:

- `runs/get_actor_run_common.ts` → `runs/get_actor_run.ts`
- `actors/search_actors_common.ts` → `actors/search_actors.ts`
- `actors/fetch_actor_details_common.ts` → `actors/fetch_actor_details.ts`
- `actors/call_actor_common.ts` → `actors/call_actor.ts`

Pure structure change — exported symbols keep identical names and signatures, no behavior change. The only substantive edit is one alias (`fetchActorDetails as fetchActorDetailsFromApi`) in the merged `fetch_actor_details.ts`, needed because the utility function and the tool export shared the name `fetchActorDetails`. Consumers updated: the four `-widget` tools, `mcp/server.ts`, `actor_executor.ts`, and four test files. Stale `*_common.ts` references in `src/tools/AGENTS.md` and an `actor_tools_factory.ts` comment were updated in the same change.


RobertCrupa (approved):

MQ37 (approved):
I agree for now 👍  But I think in future if the files get too chunky we should use more hierarchical approach - maybe subdir for each tool and split into multiple files with main content in `index.ts`.

Issue #784, opened 2026-05-04
feat: Storage tools — correctness, coverage, schema


jirispilka:
Umbrella for storage-tool work (datasets + KV stores). Tracks correctness fixes, MCP-spec gaps, test/eval coverage, and the dataset-schema redesign.

## Background

Storage tools (`get-dataset-items`, `get-dataset`, `get-dataset-schema`, `get-user-datasets-list`, `get-key-value-store*`) were added without end-to-end validation. The two original bugs from this issue have landed (`DEFAULT_DATASET_ITEMS_LIMIT = 20`; #731 closed). Remaining work below.

jirispilka:
For `storage` 
- Enable all storage tools and play with
- Try Actor that outputs datasets and records in kv store (website content crawler)
- Add chart to mixpanel with tools (usage, error rate)
- Review unit-tests and fill gaps
- Review integration tests and fill gaps
- Improve evals
- fix mcp.apify.com configuration (internal)
- dataset schema: we need to check how the schema is generated. It is difficult to push dataset schema across Apify. Ad-hoc solution will be required.
- we will need to implement some evals to figure out how much detailed the schema should be. The current schema you return is very small while the dataset_schema is quite bloated on the other hand. We can parse it down but at least I have no idea what amount of tokens we should serve.

https://apify.slack.com/archives/C04URRT3934/p1778871826309349

PR #1038, opened 2026-06-29
chore: Add output-schema conformance tests for storage tools

### LINKED ISSUE #886: feat: Make sure that all storage tools are spec compliant

jirispilka:
## What we're solving

Closes #886 (under umbrella #784). "Spec compliant" for the storage tools means: each tool's emitted `structuredContent` must validate against its declared `outputSchema` — the non-coercing check a spec-compliant MCP client runs on tool results (a result with a schema but non-conforming output is rejected).

All 8 storage tools already declare an `outputSchema` and emit `structuredContent`, and a static audit of every success/error branch found **no mismatch at HEAD**. But only `get-key-value-store-keys` and `get-key-value-store-record` had a test proving conformance. For the other six tools, a future schema/output drift could ship undetected. This is a **tests-only** change.

## MCP conformance suite — cross-check (no code impact)

Ran the official `@modelcontextprotocol/conformance` suite (`server` mode over HTTP) against the local dev server as an independent cross-check. **11 passed, 21 failed**, but the failures are expected and do not indicate a defect:

- The suite tests **protocol conformance**, not tool behavior. To exercise `tools/call`, `prompts/get`, etc. it requires the server to implement its own **fixed test fixtures** (`test_simple_text`, `test_simple_prompt`, …). This server exposes real tools (`search-actors`, `call-actor`, `get-dataset-items`, …), so those scenarios fail with "tool/prompt not found" — a fixture mismatch, not a violation.
- The server **passes every scenario that maps to a capability it declares**: `initialize`, `ping`, `logging/setLevel`, `tools/list`, `resources/list`, `resources/read` (text), `resources/templates`, `prompts/list`, SSE multi-stream, and the valid-Host check.
- Optional capabilities the server does not advertise (`completion/complete`, `resources/subscribe`) are reported as failures by the suite but are not required.
- One genuine finding — the dev server does not validate the `Host`/`Origin` header (DNS-rebinding). **Out of scope here**: `src/dev_server.ts` is the dev/standby server only; the production Streamable-HTTP transport and its security live in `apify-mcp-server-internal`.
- 

https://claude.ai/code/session_016D4Lyju2prcnEiw4hPAZhK

### DIFF
--- a/tests/unit/helpers/tool_context.ts
+++ b/tests/unit/helpers/tool_context.ts
@@ -1,5 +1,6 @@
 import type { CallToolResult } from '@modelcontextprotocol/sdk/types.js';
 import { decode as decodeToon } from '@toon-format/toon';
+import Ajv from 'ajv';
 import { expect } from 'vitest';
 
 import { FAILURE_CATEGORY, HelperTools, TOOL_STATUS } from '../../../src/const.js';
@@ -67,4 +68,16 @@ export function expectSoftFailInvalidInput(result: { isError?: boolean; toolTele
             failureCategory: FAILURE_CATEGORY.INVALID_INPUT,
         }),
     );
+}
+
+/**
+ * Assert that `structuredContent` validates against `schema` under a non-coercing AJV
+ * (`new Ajv({ strict: false })` compiled directly on the schema) — the same shape of check the MCP
+ * SDK client runs on tool results. The repo's lenient `compileSchema` coerces types and would hide a
+ * mismatch, so conformance tests compile here instead. Surfaces AJV errors in the failure message.
+ */
+export function expectStructuredContentMatchesSchema(schema: object, structuredContent: unknown): void {
+    const validate = new Ajv({ strict: false }).compile(schema);
+    const valid = validate(structuredContent);
+    expect(valid, JSON.stringify(validate.errors)).toBe(true);
 }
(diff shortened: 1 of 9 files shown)


RobertCrupa (approved):
Nice one!

tests/unit/helpers/tool_context.ts
```diff
@@ -68,3 +69,15 @@ export function expectSoftFailInvalidInput(result: { isError?: boolean; toolTele
...
     );
 }
+
+/**
+ * Assert that `structuredContent` validates against `schema` under a non-coercing AJV
+ * (`new Ajv({ strict: false })` compiled directly on the schema) — the same shape of check the MCP
+ * SDK client runs on tool results. The repo's lenient `compileSchema` coerces types and would hide a
+ * mismatch, so conformance tests compile here instead. Surfaces AJV errors in the failure message.
+ */
+export function expectStructuredContentMatchesSchema(schema: object, structuredContent: unknown): void {
```
MQ37: change: there is pre-existing `expectSchemaConformingStructuredContent` func for this - I would adapt and use that instead of introducing duplicate func

MQ37 (approved):
Just one one possible duplicate func. Otherwise LGTM!

tests/unit/helpers/tool_context.ts
jirispilka: Nice, thanks!

PR #1029, opened 2026-06-25
feat: cap individual tool response sizes

jmikitova:
Closes #1005 

## What

Several MCP tools returned unbounded output, so a single tool call could blow the AI-chat context budget (e.g. `get-dataset-items` reaching ~236k tokens). This caps per-call output so no single response is unbounded, and every truncated response tells the model how to get the rest (paginate / fetch full record / fetch full schema).

## Changes

| Tool | Cap |
| --- | --- |
| **get-dataset-items** | Clamp effective `limit` to `MAX_DATASET_ITEMS_LIMIT` (100) **pre-fetch**, then a `DATASET_ITEMS_MAX_BYTES` (50 KB) byte cap on the encoded payload, dropping trailing items and steering to `offset` pagination from the **actually-returned** count (lossless resume). |
| **get-key-value-store-record** | Cap inlined text/JSON values at `KV_RECORD_MAX_INLINE_TEXT_BYTES` (50 KB) → truncate + link to the full record (mirrors the existing 256 KB binary cap). |
| **search-apify-docs** | Clip per-snippet `content` to `DOCS_SNIPPET_MAX_LENGTH` (1000), with a pointer to `fetch-apify-docs` for the full page. |
| **call-actor** | Stop dumping the full input schema on input-validation / missing-input errors (could be ~30k chars for complex Actors). Return the AJV validation errors + a `fetch-actor-details` pointer instead. |

Measured on what the **model actually receives**: the chat serializes MCP results via `mcpToModelOutput` (`schemas: "automatic"`), which forwards only `content` and drops `structuredContent`, so the caps measure the `content` payload.

## Tests

New/extended unit tests in `tests/unit/` for each cap (over-cap input stays bounded, truncation/pointer present, pagination offset resumes from the returned count). Full unit suite green (68 files, 973 pass / 2 skip).

## Out of scope (follow-up)

The model paginating through an entire dataset (many capped calls) is a separate prompt/behavior lever, not a per-call size fix — worth its own change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1046, opened 2026-07-01
feat: Cap search-apify-docs output with Algolia snippeting

### LINKED ISSUE #1042: Investigate issue with Apify Docs access via MCP
When Apify Docs are access through MCP certain pages return unformatted output which fills quite a large chunk of context. Apify AI suffers from these issues the most as it has limited context size when calling `search-apify-docs`.

Our goal:
- Find pages that would cause context problems - through MCPC try to acquire the pages and check the size of output.
- Explore possible solutions how to pass them smaller
   - Chunk the content of page
   - Return full content of page in search-apify-docs only for first result.
   - Utilize different service for docs look up - right now Algolia, maybe Recombee has similar solution, or we can do it inhouse
- Add parameter in MCP tool call to specify size of returned docs.

vojtechj-apify:
## Problem
`search-apify-docs` returned the full Algolia `content` attribute per hit. API-reference records inline the entire schema (~34.5K chars each, byte-identical across endpoints), so a `limit: 20` search returned **~173K tokens** — mostly duplicated boilerplate — overrunning limited-context agents (Apify AI hit hardest). Part of Context Bloating problem.

## Fix
Bound each hit at the source with query-time Algolia snippeting in `prepareAlgoliaRequest`:
- `attributesToSnippet: ['content:100']` — a 100-word, match-centered window instead of the full attribute.
- Matched terms come wrapped in highlight markup; we strip it via a private-use sentinel tag (`DOCS_SNIPPET_HIGHLIGHT_TAG`). An empty tag is treated as "unset" and falls back to Algolia's verbose default `<span>`, so a sentinel is required.
- `attributesToHighlight: []` + `attributesToRetrieve: ['url_without_anchor','anchor']` shrink the wire response (1.43 MB → 27 KB).

`processAlgoliaResponse` reads `_snippetResult.content.value` and strips the sentinel.

## Why it's safe (no ranking/relevance change)
The diff only adds display/retrieval params; it never touches the ranking inputs (`query` / `filters` / `typeFilter` / `facetFilters`), so Algolia ranking is unchanged by construction. Verified two ways:
- **Actual `master` build vs this branch, via mcpc → live Algolia** (5 queries): identical URLs + order for every query.
- **Before/after eval** (13 queries): identical results, ~90% content reduction, latency unchanged (~49 → ~48 ms).

Measured reductions (master → branch output): `create webhook` 217,793 → 11,390 B (**−95%**); `actor input schema` 27,496 → 15,362 B. **crawlee-js/py are byte-identical** — their `lvl1` records have no `content`, so the change is a no-op there.

## Tests
`tests/unit/utils.apify_docs.test.ts` (mocked Algolia): the request carries the snippet params; result processing prefers and strips the snippet; hits without a snippet (crawlee) stay url-only.

## Scope
Bounds `search-apify-docs` only. `fetch-apify-docs` is unchanged — it serves the full page on demand (a deliberate single-page pull), which is the right tool for full context.

## Relationship to #1029
This is an **alternative** to #1029's `search-apify-docs` cap (`clipSnippet`, a 1000-char head-slice) — the two are **mutually exclusive for search and should not both be merged**. Snippeting is match-centered (vs a head-slice of the boilerplate top) and also shrinks the Algolia wire payload. #1029's other caps (dataset / key-value-store / call-actor) are independent and unaffected.

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -125,6 +125,22 @@ export const DOCS_SOURCES = [
     },
 ] as const;
 
+/**
+ * Word window for Algolia `attributesToSnippet` in `search-apify-docs`. Bounds each hit to a
+ * match-centered snippet instead of the full indexed `content` attribute — API-reference records
+ * inline the whole OpenAPI schema (~34.5k chars each), so 20 hits returned ~173k tokens. The agent
+ * fetches full pages via `fetch-apify-docs`.
+ */
+export const DOCS_SNIPPET_MAX_WORDS = 100;
+
+/**
+ * Sentinel used as Algolia `highlightPreTag`/`highlightPostTag` for docs snippets, stripped before
+ * returning. An empty-string tag is treated as "unset" and falls back to Algolia's default
+ * `<span class="algolia-docsearch-suggestion--highlight">` markup, so we set a private-use
+ * character (U+E000) and remove it.
+ */
+export const DOCS_SNIPPET_HIGHLIGHT_TAG = '\uE000';
+
 export const ALLOWED_DOC_DOMAINS = ['https://docs.apify.com', 'https://crawlee.dev'] as const;
 
 export const APIFY_STORE_URL = 'https://apify.com';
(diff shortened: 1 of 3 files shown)


MQ37 (approved):
Great, thanks! I was not aware that Algolia started or is returning full content in some cases.

Issue #641, opened 2026-04-07
[Bug]:


pbarsoom:
### Server type

Hosted (mcp.apify.com)

### MCP Client

Claude Desktop

### Operating System

macOS

### What happened?

I have the Apify MCP connector installed and configured in Claude Cowork under Desktop connectors. The token is set and the enabled tools field is configured with actors,docs,apify/rag-web-browser. The connector shows as "Enabled" in the Cowork settings UI.
However, when I start a conversation and ask Claude to use the Apify tools (e.g., scrape Google Maps for local businesses), Claude does not recognize that the Apify MCP tools are available. It searches the MCP registry instead, reports that no Apify connector exists, and suggests workarounds like web search or Chrome browsing.
I confirmed that the Apify tools do not appear in Claude's loaded tools list — they are simply not being surfaced to the model despite the connector being installed and enabled.
This matches the behavior described in the existing issue where Claude "will not connect and say there is no connection" and "tries different workarounds with Chrome."

### Steps to reproduce

Install the Apify MCP connector in Claude Cowork (Desktop section)
Enter a valid Apify API token
Set enabled tools to actors,docs,apify/rag-web-browser
Save and confirm the connector shows as Enabled
Start a new conversation
Ask Claude to use Apify to scrape Google Maps (e.g., "using the Apify connection, scrape Google Maps to find 5 dispensaries near zip code 10014")
Claude does not see the Apify tools and suggests alternative approaches instead

### MCP server configuration

```json
Configured via Claude Cowork Desktop Connectors UI (not a local JSON config).

- Server: apify-mcp-server (Hosted, mcp.apify.com)
- Status: Enabled
- Apify token: (set)
- Enabled tools: actors,docs,apify/rag-web-browser
- Read-only tools (6): Always allow
- Visible tools: Search Actors, Fetch Actor details, Get Actor run
```

### Error logs

```text
2026-04-07T03:07:11.919Z [Apify] [info] Initializing server... { metadata: undefined }
2026-04-07T03:07:11.919Z [Apify] [info] Using built-in Node.js for MCP server: Apify { metadata: undefined }
2026-04-07T03:07:11.921Z [Apify] [info] Server started and connected successfully { metadata: undefined }
2026-04-07T03:07:11.952Z [Apify] [info] Client transport closed { metadata: undefined }
2026-04-07T03:07:11.952Z [Apify] [info] Server transport closed (intentional shutdown) { metadata: undefined }
[...]

PR #1050, opened 2026-07-02
feat: Add report-problem tool for agents to report issues

### LINKED ISSUE #748: feat: Add report-issue tool for agent feedback
## Problem
When an agent gets stuck against our MCP server, we only see the failing tool call — never the higher-level "I was trying to do X but your tool's response didn't tell me how to proceed." We have no direct channel for agent feedback.

Related: #673 (Improve tool call accuracy).

## Proposed solution
Ship a standalone helper tool (`report-issue`) that the agent calls when blocked.

- Add `HelperTools.REPORT_ISSUE = 'report-issue'` to `src/const.ts:25`.
- New tool file: `src/tools/common/report_issue.ts` (following the pattern of `src/tools/common/search_apify_docs.ts` — internal tool, Zod schema, AJV validate, `buildMCPResponse` return).
- Zod schema:
  - `goal: z.string()` — what the agent was trying to accomplish
  - `attempted: z.string()` — what it tried (tools, inputs)
  - `blocker: z.string()` — where it got stuck / what was confusing
  - `tool_name: z.string().optional()` — tool tied to the issue
- Register it in the appropriate category in `src/tools/categories.ts` (likely enabled by default like docs tools).
- Pipe submissions through telemetry: emit a new Segment event from `src/telemetry.ts` (e.g. `'MCP Agent Feedback'`) carrying the four fields plus `mcp_session_id`, `mcp_client_name`, etc.
- The Slack/GitHub fan-out lives in `apify-mcp-server-internal` (consumes the Segment event) — keep delivery out of the public repo.
- Mention the tool in the description footers of tools where agents commonly get stuck (`call-actor`, `search-actors`, `fetch-actor-details`) so the agent knows the channel exists. One sentence: *"If you're blocked or our response is unclear, call `report-issue` so we can fix it."*

## Plan
- [ ] Add `HelperTools.REPORT_ISSUE` enum + tool implementation
- [ ] Add Segment event in `src/telemetry.ts`; extend types in `src/types.ts`
- [ ] Register in tool categories (default-enabled)
- [ ] Add tool-mention sentence to `call-actor`, `search-actors`, `fetch-actor-details` descriptions
- [ ] Unit test + `tests/integration/suite.ts` case
- [ ] Coordinate with `apify-mcp-server-internal` for Slack + GitHub fan-out

## Alternatives considered
- **Pure log-based feedback (softFail)** — not visible to product, only to oncall.
- **Bundling feedback into existing tools as an optional field** — agents won't reach for an optional field; a dedicated tool framed as "get unstuck" has stronger pull.

RobertCrupa:
Implements the `report-problem` MCP tool from #748 — a channel for agents to report problems with Apify's tools or Actors, so we get signal on where agents get stuck.

## What's included

- **New tool `report-problem`**: returns a plain-text acknowledgement, always success. `readOnlyHint: true`, not payment-gated.
  Schema:
  - `message` — required string, 1–2000 chars. What happened.
  - `actorId` — optional string, ≤200 chars.
  - `actorRunId` — optional string, ≤200 chars.
  - `relatedTools` — optional string[], up to 20 items, each ≤100 chars.
- **Default-enabled `report-problem` tool category** in the registry.
- **Telemetry**: a successful call emits a dedicated `MCP Agent Feedback` Segment event (fields mapped to snake_case, absent optionals dropped) alongside the standard tool-call event. Identity handling mirrors the tool-call event.

## Gating — served only when all three hold

1. **Telemetry enabled.** The tool's only function is forwarding submissions via telemetry, so it is never served when telemetry is off (it would just fake an acknowledgement into the void).
2. **Client allowed.** Hidden from Anthropic surfaces (Claude.ai / Claude Desktop / Claude Code) pending the directory review, via a substring blocklist (`claude`, `anthropic`) matched against `clientInfo.name`. The tool cannot be judged until the client is known, so it is withheld until initialize and added on the flush if the client allows.
3. **Category selected.** The `report-problem` category is enabled (default on).

All other tools are unconditionally servable, so recovery loads compose them eagerly and they survive a load that never sees an initialize; only `report-problem` is withheld until the client is known.

## Discovery

- **Error nudge**: failed (`isError`) tool results get a report-problem nudge appended to their text, prompting the agent to report at the moment it decides what to do next. Full nudge for internal/unknown errors; a softer nudge for `INVALID_INPUT` (likely the agent's own mistake); suppressed for expected user-resolvable states (auth, payment/approval required). Also applied to failed long-running task results. Never appended when the tool isn't served or when `report-problem` itself is the failing tool.
- **Server instructions**: a gentle line pointing agents at `report-problem`, emitted only when the tool is actually served.

## Testing

- **Workflow eval** `report-problem-on-tool-error`: injects a `call-actor` failure and checks the agent proactively reports the blocker (rather than only telling the user it failed). Required adding tool-failure injection to the eval harness.
- **Unit tests**: tool behavior/schema, gating, client blocklist, server instructions, telemetry event shaping.
- **Integration suite**: `report-problem` listing and gating.
- Manually verified with mcpc: schema, valid/invalid submissions, and the Anthropic block (filtered from the tool list for a `claude`-family client while remaining available to non-Anthropic clients).

## TODO

- [x] Get access to Segment so I can check the reported problems.

Closes #748


src/utils/server-instructions/index.ts
```diff
@@ -83,5 +86,14 @@ ${
   \`${HelperTools.STORE_SEARCH}\` finds robust and reliable Actors for specific websites; ${RAG_WEB_BROWSER} is a general and versatile web scraping tool.
 - **Dedicated Actor tools (e.g. ${RAG_WEB_BROWSER}) vs \`${HelperTools.ACTOR_CALL}\`:**
   Prefer dedicated tools when available; use \`${HelperTools.ACTOR_CALL}\` only when no specialized tool exists in the Apify store.
-`;
+${
+    feedbackAvailable
+        ? `
+## Reporting problems and feedback
+- **Before you tell the user you cannot complete a request — because a tool or Actor is missing, errored, or returned a confusing, wrong, or empty result — you MUST first call \`${HelperTools.FEEDBACK_SHARE}\`**, describing what you were trying to do and what blocked you. Do this yourself, without being asked; it is a required step, not optional.
```
RobertCrupa: This might be a bit too harsh, I will tweak it to see if it still passes the test

RobertCrupa: done [75128c4](https://github.com/apify/apify-mcp-server/pull/1050/commits/75128c4f50f93f0de749e2523475517c3da9b29d)

jirispilka (commented):
Thanks Robert,  the architecture is right: dedicated tool, dedicated Segment event. Before merge we need to discuss it a bit more Summary first, details below.

With Claude

## To discuss

- **Rename `share-feedback` to `report-problem`.** The tool only gets called at failure moments, so it is a problem-report channel — name it that. This also makes it plausibly compatible with the Anthropic directory review: a descriptive diagnostic tool, no MUST instructions, only a gentle nudge.
- **Drop `npsRating`.** Failure-conditioned sampling makes the score junk data, and a rating field makes no sense on a problem-report tool.
- **Solicitation: remove MUST, keep one gentle nudge.** Scope the nudge so it never fires on payment/approval responses or the agent's own input errors.
- **Move client gating out of `upsertTools`.** It has real bypasses there. Client-based filtering belongs in the same compose step that already resolves the tool set per client (initialize handler), behind a generic tool→client blocklist, not an Anthropic-specific function.
- **Confirm with Anthropic instead of guessing.** Keep the block on Claude surfaces in v1, email with the exact design, remove the block when they OK it.
- Smaller fixes: don't serve the tool when telemetry is off, cap input sizes, per-session submission cap, update docs, coordinate the internal fan-out.

## Product

**npsRating is wrong.** Every discovery mechanism (nudge, footers, instructions) triggers on failure. Nothing prompts a rating when things work. So the score samples almost only broken sessions — it measures "how annoyed are agents when things break", not Actor quality. LLM ratings also cluster in a narrow band, so NPS math is meaningless even with balanced sampling. Drop the field. For Actor quality, use run telemetry we already have (success rate, failure categories, call-actor → get-dataset-items completion) — unbiased, covers every session, can't be gamed by prompt wording.

**Rename to `report-problem`.** One tool, one honest job: "Report a problem with an Apify tool or Actor: what you were doing, what failed." This matches what the sampling can actually deliver, and it changes the Anthropic story. Their rejection criteria target vendor-serving behavioral instructions ("call this before giving up", "you MUST"), not diagnostic tools with descriptive wording — our existing nextStep guidance is fine for the same reason: it serves the user's task. A neutral report-problem tool with a gentle nudge is a defensible submission. Note: positive feedback / Actor ratings are explicitly out of scope for v1; if we ever want them, that's a separate design with success-moment sampling.

**Other product points:**
- Privacy is currently prompt-only ("do not include personal data"). The fan-out side (internal) must treat report text as untrusted input — escape markdown, no auto-actions — because this is an unauthenticated free-text write path.
- Document the tool and how to disable it (`tools=` without the category) in README.

## Code

1. **Client gating is in the wrong layer and has bypasses.** The `upsertTools` filter reads `initializeRequestData`, which is only set during `initialize`. With explicit `--ui default|apps` or `UI_MODE`, stdio composes tools before initialize and the block never applies. The footers are static constants in three tool descriptions, so Claude surfaces see "call share-feedback before giving up" regardless — the exact pattern in the exact place the directory reviews. And internal's multi-node recovery restores tools on a node that never saw initialize.

   Fix: make "client known" (not "mode resolved") the trigger for composing helper tools — the pending-queue mechanism already exists, the condition is just wrong for explicit modes. MCP guarantees no tools/list before initialize, so deferring composition is always safe. Then apply a generic policy inside that one compose step:

   ```ts
   const TOOL_CLIENT_BLOCKLIST: Record<string, string[]> = {
       [HELPER_TOOLS.PROBLEM_REPORT]: ['claude', 'anthropic'],
   };
Same step decides the footer variants of the three tool descriptions (append only when report-problem is served). upsertTools stays policy-free. Blocklist is data, not code, because we'll extend it: OpenAI's directory has near-identical criteria, and localAgent (seen in our telemetry, matches neither substring) may need adding — verify in Mixpanel first whether it's a Claude surface or an API/Agent-SDK path we actually want feedback from. Longer term, put a vendor field in mcp-client-capabilities (its registry only knows claude-ai and claude-code today) and match exactly instead of by substring.

2. Scope the nudge. It currently fires on every isError: true result, including payment-required and approval responses — telling the agent to "report to Apify" when the correct next step is the payment flow — and on the agent's own invalid-input errors. Exclude those categories, and gate it with the same predicate as the tool itself.

3. Instructions: replace MUST with one line. "If a tool or Actor fails and you cannot resolve it, you can report it with report-problem." The MUST version delays failure reports to users and invites junk submissions.

4. Don't serve the tool when telemetry is disabled. With --telemetry-enabled false it's still listed and returns "Feedback submitted" while the event goes nowhere. The tool has no other function.

5. Cap inputs. message max ~2000 chars, relatedTools max ~20 items, "keep it to a few sentences" in the description.

6. Per-session submission cap (in-memory is fine — it's telemetry, not critical data). Return "already recorded" past the cap.

7. Rename mechanics. Decide the Segment event name now (MCP Problem Report vs keeping MCP Agent Feedback) — renaming events later fragments Mixpanel history. The feedback category name can stay.

MQ37:
@jirispilka I think it was JC's proposal to make this `share-feedback` tool explicitly with the `npsRating` so the agents can also report what is possibly working well along with what is broken. But on the other hand I think right now we are mostly in interested in what is not working well so let's do the `report-issue` (problem sounds bad IMHO).

src/mcp/server.ts
```diff
@@ -345,7 +347,10 @@ export class ActorsMcpServer {
             await this.resolveWidgets();
 
             const result = await sdkInitHandler(request);
-            result.instructions = getServerInstructions(this.serverMode);
+            // Tools are final here (updateToolsAfterServerModeResolved ran above, after the
+            // Anthropic hard-block applied), so tool presence is the ground truth for whether to
+            // advertise share-feedback in the instructions.
+            result.instructions = getServerInstructions(this.serverMode, this.tools.has(HELPER_TOOLS.FEEDBACK_SHARE));
```
MQ37: nit: what about we pass `server` instance instead and the `getServerInstructions` can then check whatever it wants without adding additional arg?

MQ37 (commented):
I agree with Jirka, let's make sure it is disabled when telemetry is off.

Also I noticed that the long running task path may be missing the report feedback nudge - but I would honestly remove the nudge and just state that there is this tool available, but not tell agent must/should use it, just state the fact that the tool is available. 

Also I agree that we should focus on the negative cases and I would rename the tool to `report-issue` and remove the score and just keep `message: string` as an input to keep it simple.

jancurn:
Fair enough, I think it makes sense to have this to report problems/issues only, and drop NPS in that case. we'll see how it works later

jancurn:
BTW, once we have this, we should gently nudge the agent to use this tool on problems. Most tools are not loaded into context, so otherwise the agents might not know about it at all

MQ37:
@RobertCrupa Follow-up review found one edge-case bug in `report-problem` gating on the hosted multi-node path. Line-cited against internal HEAD `8e65ef4` and this branch.

## `report-problem` can silently vanish after cross-node session recovery

**Severity: High (edge-case-triggered), not a regression** — it's a gap in the new feature. Normal recovery is fine; only a narrow clientInfo-erasure window exposes it.

### Fact chain (all verified)

1. **Gated tools are withheld until the client is known** — `apify-mcp-server/src/mcp/server.ts:394`:
   ```ts
   if (isClientGatedTool(toolName) && !this.clientKnown) return false;
   ```
   `clientKnown` = `initializeRequestData != null` (`:367`).

2. **On restore, "client known" depends entirely on `clientInfo` in Redis** — `apify-mcp-server-internal/src/server/streamable.ts:519`:
   ```ts
   let initializeRequestData;
   if (sessionState.clientInfo) { initializeRequestData = {...}; }
   // else: undefined → clientKnown === false
   ```

3. **Restore never runs the initialize handler, so the pending-queue flush never fires** — your own comment at `streamable.ts:535`: *"initialize won't fire during session restore (transport is `_initialized=true`)."* The drain `composePendingToolsForClient()` is called only from the initialize handler (`server.ts:351`). On restore, `loadToolsByName` (`streamable.ts:553`) withholds the gated tool and parks it in `pendingToolsUntilClientKnown` (`server.ts:521`) — a queue nothing drains on this transport.

4. **One path erases clientInfo.** Two `storeSessionState` call sites: `streamable.ts:671` (initial connect) passes clientInfo ✅; `streamable.ts:429` (orphaned-session fallback) omits it ❌:
   ```ts
   await stateStore.storeSessionState(sessionId, mcpServer.listAllToolNames()); // no clientInfo
   ```
   Missing arg → empty write (`redis-state-store.ts:51`: `clientInfo ? JSON.stringify(clientInfo) : ''`) → read back as `null` in `getSessionState`. A `updateSessionTools` that *preserves* clientInfo exists (`redis-state-store.ts:64`) but isn't used here.

### Reproduction (multi-node, shared Redis, round-robin LB)

1. Non-Anthropic client (e.g. `cursor`) connects on node A → `report-problem` served, state saved **with** clientInfo (`:671`).
2. Session's Redis state expires/evicts — the `:429` comment names this: *"might happen for long-running sessions."*
3. Request hits node B (session in memory, not Redis) → fallback at `:429` re-writes Redis with `clientInfo = ''`.
4. Later request round-robins to node C (session only in Redis) → `handleSessionRestore`: `clientInfo` null → `clientKnown` false.
5. `loadToolsByName` restoring `report-problem` → `isToolServable` false → filtered out of the committed set, queued into the undrainable queue.

**Result:** Redis `tools` still lists `report-problem`; node C doesn't serve it; the failure-nudge also disappears (all three sites gate on `this.tools.has(PROBLEM_REPORT)` — `server.ts:900`, `1732`, `1884`). No error, no warning, no metric.

### Fix
1. **Primary:** `streamable.ts:429` — use `updateSessionTools(...)` instead of `storeSessionState(...)` so clientInfo is never erased.
2. **Defense in depth:** in `loadToolsByName`, when a client-gated tool is explicitly requested but withheld due to `!clientKnown`, `log.warning` (tool + sessionId) — turns any silent recurrence into a greppable signal.

### Regression test
Store a session in Redis with `clientInfo = ''` and `tools` incl. `report-problem`, restore via `handleSessionRestore`, assert it's served (or a warning is emitted) — not a silent drop.

MQ37:
@RobertCrupa Fix for the cross-node \`report-problem\` gating bug ([comment above](https://github.com/apify/apify-mcp-server/pull/1050#issuecomment-4914260770)) is ready: apify/apify-mcp-server-internal#641.

Root cause was in \`apify-mcp-server-internal\`, not here — \`handleExistingSession\`'s orphaned-session fallback (\`streamable.ts\`) rewrote Redis session state with tools only, dropping \`clientInfo\` whenever a node still held the session in memory but Redis had evicted it. A later restore on a different node then read \`clientInfo: null\` and could never recover it, silently disabling any client-gated tool for the rest of that session.

The internal PR includes a real two-node regression test (connect → confirm \`report-problem\` listed → delete session from Redis → hit the same node again → hit a second node that never saw the session → confirm still listed).

It's flagged do-not-merge-yet: it depends on this PR merging and releasing, then \`apify-mcp-server-internal\`'s \`@apify/actors-mcp-server\` dependency bumping past \`0.11.5\` to pick up \`report-problem\`. Once that lands, apify-mcp-server-internal#641 is ready to go in right behind it.

src/mcp/server.ts
```diff
@@ -340,30 +348,68 @@ export class ActorsMcpServer {
...
+    /**
+     * Whether a tool may be served on this connection right now:
+     * - report-problem's only function is forwarding submissions via telemetry, so it is never
+     *   servable when telemetry is disabled (it would just fake an acknowledgement into the void).
+     * - A client-gated tool (has a {@link TOOL_CLIENT_BLOCKLIST} rule) cannot be judged until the
+     *   client is known, so it is withheld until then — the initialize flush re-composes and adds it
+     *   if the client allows. Client-agnostic tools are always servable, so they compose eagerly and
+     *   survive a recovery load that never sees an initialize.
+     */
+    private isToolServable(toolName: string): boolean {
```
MQ37: nit: I think if we are introducing generic function like this it should be truly generic/abstracted or focused only on the single tool. I would either focus this truly on the report tools and simplify the logic and rename to something like `isReportProblemToolServable()` or remove this focused logic and make it more generic:
```
if (toolName === HELPER_TOOLS.PROBLEM_REPORT && !this.telemetryEnabled) return false;
```

MQ37 (approved):
Cool, thank you! Pre-approving.

But please before merging please let's nudge/inform the agent about the report tool existence also in the INVALID INPUT tool failure as there can be some funky bugs causing input issues that should be reported and agent might not report that - if we keep getting lots of spam, we can then rework the nudging.

src/mcp/server.ts
RobertCrupa: Done, I went with focusing it on just the report tool to keep it simple. [4fdf542](https://github.com/apify/apify-mcp-server/pull/1050/commits/4fdf542cf441ebf88bd039b8de530d11f026d4ca)

src/mcp/server.ts
```diff
@@ -340,30 +348,70 @@ export class ActorsMcpServer {
...
+     * other tool composes eagerly, so a recovery/rehydration load without an initialize still
+     * restores them. report-problem is withheld until the client is known and re-added by the
+     * initialize flush. Used by every input-driven load path and the flush. (loadActorsAsTools
+     * upserts actor tools directly; actor tools are never gated, so they need no filtering.)
+     */
+    private composeToolsForClient(input: Input, actorTools: ToolEntry[]): ToolEntry[] {
+        return getToolsForServerMode(input, actorTools, this.serverMode).filter(
+            (tool) => tool.name !== HELPER_TOOLS.PROBLEM_REPORT || this.isReportProblemServable(),
+        );
+    }
```
jirispilka: Honestly, this condition reads like a puzzle A != B || C

There is a much simpler and cleaner way to put it: 
```ts
  if (this.isReportProblemServable()) return tools;
    return tools.filter((tool) => tool.name !== HELPER_TOOLS.PROBLEM_REPORT);
```

src/mcp/server.ts
```diff
@@ -837,8 +896,16 @@ export class ActorsMcpServer {
             // Captured by `captureResult` so the `finally` block can measure response size for telemetry.
             let toolResult: unknown = null;
             const captureResult = <T>(r: T): T => {
-                toolResult = r;
-                return r;
+                // On a failed result, nudge the agent to report the blocker via report-problem at the
+                // moment it decides what to do next. Gated on the tool actually being served (see
+                // isReportProblemServable), so clients where it is blocklisted or telemetry is off never see it.
+                const augmented = appendReportProblemNudge(r, {
```
jirispilka: This nudge is repeated three times in the server.ts, can we simplify it?

```ts
private withReportProblemNudge<T>(result: T, toolName: string | undefined, cat?: string): T {
    return appendReportProblemNudge(result, {
        failingToolName: toolName,
        available: this.tools.has(HELPER_TOOLS.PROBLEM_REPORT),
        failureCategory: cat,
    });
}
```

src/tools/registry.ts
```diff
@@ -72,6 +73,7 @@ export const toolCategories = {
         { default: callActorDefault, apps: callActorApps },
     ],
     docs: [searchApifyDocs, fetchApifyDocs],
+    'report-problem': [reportProblem],
```
jirispilka: I would not add separate category just for report tool, I would add it to experimental or `dev` (I incline to `dev`, @MQ37  what do you think?)

src/tools/dev/report_problem.ts
jirispilka, line 1: Again, I would move this tool to dev or experimental, this placement reads badly

src/tools/report_problem/report_problem.ts
```diff
@@ -0,0 +1,114 @@
...
+    type: TOOL_TYPE.INTERNAL,
+    name: HELPER_TOOLS.PROBLEM_REPORT,
+    title: 'Report a problem',
+    description: REPORT_PROBLEM_DESCRIPTION,
+    inputSchema: reportProblemInputSchema,
+    ajvValidate: compileSchema(reportProblemInputSchema),
+    paymentRequired: false,
+    annotations: {
+        title: 'Report a problem',
+        readOnlyHint: true,
```
jirispilka: I understand why you want to keep it read-only so the client doesn't prompt the user for approval). 

I would prefer to be honest here. Anyway, any agent checking tool description and comparing it to our annotations will flag it. Like if we want to Claude or ChatGPT, they will complain.

So I would keep it as false, I know that @MQ37  wanted to have it readOnlyHint true.

src/telemetry.ts
```diff
@@ -21,6 +25,7 @@ const SEGMENT_FLUSH_INTERVAL_MS = 5_000;
 // Event names following apify-core naming convention (Title Case)
 const SEGMENT_EVENTS = {
     TOOL_CALL: 'MCP Tool Call',
+    AGENT_FEEDBACK: 'MCP Agent Feedback',
```
jirispilka: We need to change this event to: `MCP Reported problem`

src/const.ts
```diff
@@ -54,9 +54,19 @@ export const HELPER_TOOLS = {
...
 export type HelperToolName = (typeof HELPER_TOOLS)[keyof typeof HELPER_TOOLS];
 
+/**
+ * Client-name substrings (lowercased, matched against `clientInfo.name`) that `report-problem` is
+ * hidden from. Applied once per connection in the compose step, where the client is known.
+ * `report-problem` is hidden from Anthropic surfaces (Claude.ai / Claude Desktop / Claude Code)
+ * pending the directory review. Substring matching covers new client builds without a maintained
+ * allowlist; over-matching only hides an optional tool, which is the safe failure mode.
+ */
+export const REPORT_PROBLEM_BLOCKED_CLIENTS: string[] = ['claude', 'anthropic'];
```
```suggestion
export const REPORT_PROBLEM_BLOCKED_CLIENTS: string[] = ['claude', 'anthropic'] as const;
```

jirispilka: Btw. as I said earlier, these clients are also from Anthropic, please update the list

Image

jirispilka (changes requested):
Thanks, looks better now. Although there are multiple issues that we should fix.

Please also the README with `report-problem` tool

Before merging, I would love to see that we receive the event is segment. If you don't have access yet, lets ask @mr-rajce 

And we need to decide on the category to which this tool will belong.

Further, claude complains here:

server.ts — payment responses are nudged inconsistently, and the comment at 1884–1886 is inaccurate.
A 402/x402 payment error is classified INVALID_INPUT (:1328, :1796), which is not in NON_NUDGE_FAILURE_CATEGORIES (that's only AUTH + PERMISSION_APPROVAL_REQUIRED), so it receives the softer nudge — not suppression. Two consequences:

The sync path (:1325 → captureResult → nudge) appends the softer nudge to a paywall response, while the task path (:1773) returns at :1802 before any nudge site — so the same 402 behaves differently by execution mode. The comment at :1884 claims the paths "behave identically"; they don't.
That same comment says INVALID_INPUT is self-suppressed and "only real defects get it" — both false.
Suggest: decide whether payment should be nudged (I'd say no — it's a billing state, not a defect), make both paths match, and fix the comment. Note you can't blanket-suppress INVALID_INPUT (genuine input bugs should still get the softer nudge), so gate payment via its 402 status or by isolating the payment/approval branches from the nudge.

src/const.ts
jirispilka: Sorry for the cut, it is `local-agent-mode-apify` and 'local-agent-mode-Apify`

src/tools/registry.ts
RobertCrupa: If we move it to dev or experimental, it won't be on by default (which is what the issue asked for), but I can make it an exception. This way I am also inclined to move it to dev as experimental + on by default doesn't feel right :)

src/tools/report_problem/report_problem.ts
RobertCrupa: Good point, we need to ensure maximum compliance with providers, so I will revert this

src/telemetry.ts
RobertCrupa: Done, but I will capitalize the P in problem

src/tools/report_problem/report_problem.ts
RobertCrupa: Done [1e4fdbd](https://github.com/apify/apify-mcp-server/pull/1050/commits/1e4fdbd2a29542fda88dbfd71258d86e587f1412)

src/tools/dev/report_problem.ts
RobertCrupa: Done [4ed6501](https://github.com/apify/apify-mcp-server/pull/1050/commits/4ed65012c56a1238815af49672c5f66378adeb71)

src/tools/registry.ts
RobertCrupa: moved to dev [4ed6501](https://github.com/apify/apify-mcp-server/pull/1050/commits/4ed65012c56a1238815af49672c5f66378adeb71)

src/const.ts
RobertCrupa: Done [e71c66e](https://github.com/apify/apify-mcp-server/pull/1050/commits/e71c66e4f7f3fb401589d6e0be344b7d3b634399)

We turn the string to lowercase before comparing, so I only had to add one

src/telemetry.ts
RobertCrupa: [7cdda63](https://github.com/apify/apify-mcp-server/pull/1050/commits/7cdda63da2482e5d37f58e402d71293e6f8d55e1)

src/mcp/server.ts
RobertCrupa: done [7b5190e](https://github.com/apify/apify-mcp-server/pull/1050/commits/7b5190e469e87a845ca699d52f8a28702373fb14)

src/mcp/server.ts
RobertCrupa: Done [4ed6501](https://github.com/apify/apify-mcp-server/pull/1050/commits/4ed65012c56a1238815af49672c5f66378adeb71)

RobertCrupa:
I now have access to segment and I can see the old tool name, but it hasn't picked up the new one yet:
image

jirispilka:
> I now have access to segment and I can see the old tool name, but it hasn't picked up the new one yet

It depends on how you run it, it might be in the dev env only?

jirispilka (changes requested):
I've verified the tool output in the segments. It works!

Re-review of the latest push thanks, this addressed the earlier round well.

I think internal #641 is now stale

Cleanup before merge:

- Move src/tools/report_problem/ → src/tools/dev/ (git mv; keep report_problem.ts; update [src/tools/AGENTS.md](https://github.com/apify/apify-mcp-server/blob/master/src/tools/AGENTS.md)) so the folder matches its category — but settle whether dev is the final category name first; dev is an odd name for a default-served user-facing tool.

- Wrap REPORT_PROBLEM_NUDGE and REPORT_PROBLEM_INVALID_INPUT_NUDGE across lines using backslash-continuation (like REPORT_PROBLEM_DESCRIPTION in the same file) — keep them a single runtime string, don't insert real newlines.
- Dead filter + stale comment: [tests/integration/suite.ts:36-41](https://github.com/apify/apify-mcp-server/blob/master/tests/integration/suite.ts#L36-L41) filters report-problem out of getDefaultTools(), which no longer contains it → no-op + misleading comment.
- Stale comment: [src/utils/server-instructions/index.ts:20](https://github.com/apify/apify-mcp-server/blob/master/src/utils/server-instructions/index.ts#L20) still references the removed report-problem category.
- README footnote omits local-agent-mode-apify from the withheld-clients list.
- `readOnlyHint:false` may prompt for per-report approval on annotation-honoring non-blocked clients — inherent consent/call-rate tradeoff (Claude surfaces are blocklisted anyway). Let's make it `false` now and observer if we are able to gather some feedback. Let's be honest with that. We agreed on this with @MQ37 
- Tests — trim the ~1 file of over-build (no coverage lost): collapse mcp.server.report_problem_nudge.test.ts (93 lines, module mock + private-handler reflection, 2 assertions) to a single "dispatch wires the nudge" case and move the toolTelemetry-stripping assertion to utils.tool_status.test.ts; drop the tools=dev/tools=storage selection cases from the gating test (already covered purely in utils.tools_loader.test.ts); drop one duplicate server-instructions default case. Gap to add: relatedTools per-item ≤100 and actorRunId ≤200 caps are untested.

jirispilka (approved):

PR #1051, opened 2026-07-02
fix: Flush Segment telemetry on stdio shutdown

RobertCrupa:
## Problem

The Segment analytics client batches events (`flushAt: 50`, `flushInterval: 5s`). On stdio shutdown the process can exit before the interval timer fires, dropping any events produced in the final window — e.g. `MCP Tool Call`.

## Fix

- **`src/telemetry.ts`** — new `closeAnalyticsClient()`: a terminal `closeAndFlush()`, guarded to no-op when no client exists, nulling the singleton (idempotent).
- **`src/stdio.ts`** — runs it on `process.on('beforeExit')`, mirroring the existing Sentry cleanup in `instrument.ts`.
- **`tests/unit/telemetry.test.ts`** — flushes when a client exists; no-op when none.

Deliberately kept out of `ActorsMcpServer.close()` and the SIGINT handler — those are shared with HTTP mode, where the process-level Segment client is shared across sessions and a terminal flush would break other sessions.

## Known gaps

- **SIGINT (Ctrl-C) not covered** — that path `process.exit(0)`s and skips `beforeExit`, same as Sentry's cleanup already does. Covering it needs a broader change to the shared handler; out of scope here.
- **Not exercised end-to-end** — unit tests cover the flush logic; real Segment delivery on a live shutdown isn't observable without a write key/network.

### DIFF
--- a/src/stdio.ts
+++ b/src/stdio.ts
@@ -34,7 +34,7 @@ import { ApifyClient } from './apify_client.js';
 import { DEFAULT_TELEMETRY_ENV, TELEMETRY_ENV } from './const.js';
 import { processInput } from './input.js';
 import { ActorsMcpServer } from './mcp/server.js';
-import { getTelemetryEnv } from './telemetry.js';
+import { closeAnalyticsClient, getTelemetryEnv } from './telemetry.js';
 import type { ApifyRequestParams, Input, ServerModeOption, TelemetryEnv, ToolSelector } from './types.js';
 import { isApiTokenRequired } from './utils/auth.js';
 import { parseCommaSeparatedList } from './utils/generic.js';
@@ -234,6 +234,13 @@ async function main() {
     const mcpSessionId = randomUUID();
 
     await mcpServer.connect(transport);
+
+    // Flush buffered Segment events before the process exits (dominant path: the client closes
+    // stdin, the transport ends, the loop empties). Mirrors the Sentry beforeExit cleanup in
+    // instrument.ts. No-op when telemetry is off (no client was ever created).
+    process.on('beforeExit', async () => {
+        await closeAnalyticsClient();
+    });
 
     const sdkOnMessage = transport.onmessage;
     transport.onmessage = (message) => {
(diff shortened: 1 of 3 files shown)


jirispilka (commented):
Thanks Robert. I looked into this with Claude and I don't think the fix actually does anything.

### Claude:
The premise is that the process exits before the 5s flush timer fires. But Segment's flush timer is a ref'd setTimeout — it keeps the event loop alive, so on a normal stdin-close shutdown the process can't exit until that timer fires and sends the batch. beforeExit only runs once the loop is empty, i.e. after the batch already went out. So on that path closeAnalyticsClient() has nothing left to flush.

On the paths where events can actually be lost — SIGINT, SIGTERM, and the
process.exit(1) in main().catch — Node doesn't emit beforeExit at all, so the hook
never runs there either.

So as it stands it doesn't change behavior on any real shutdown path. The tests don't catch this because they mock closeAndFlush and only assert it was called — revert the stdio.ts change and they stay green.

Before we spend more time here, can we confirm we're actually dropping events, and on which path?
- stdin closes normally → the ref'd timer already covers us, nothing to fix.
- SIGINT/SIGTERM → we need to flush in the stdio exit path (with a timeout), not beforeExit.
- clients SIGKILL → no shutdown hook can help; we'd flush eagerly instead (lower flushAt or flush per tool call).

One note: the shared-singleton reasoning is right — keeping this out of close()/the SIGINT handler is correct, since that client is shared across HTTP sessions.

MQ37 (approved):
Approving, I think this change is low-risk and it's better than possibly missing some important events. But we can at least add some log message so human user would know that we are flushing events and it does not seem frozen.

But I think it would be better to go through the Jirka's comment and verify it.

RobertCrupa:
Good catch, @jirispilka. I still don't have access to Segment, so I cannot verify yet. I can see it not being an issue, since it wasn't reported before.

I will verify and close it if it's not an issue.

RobertCrupa:
After some more testing, it looks like this doesn't do anything, as @jirispilka mentioned, the segment timer is referenced, so this code does nothing.

This also means that the sentry cleanup that was mirrored by the PR is a no-op.

jirispilka:
ok, thanks Robert!

Issue #1053, opened 2026-07-04
feat(tools): add read-only lookup tools so agents stop reaching for raw curl


DaveHanns:
## Summary

Several very common read-only lookups an agent performs while iterating on an Actor do not exist as MCP tools today. When agents (Claude / Cursor / other LLM harnesses) are given both this MCP server and shell access, they consistently bypass MCP and default to raw `curl https://api.apify.com/...` or `apify` CLI calls for these lookups — because the equivalent MCP tool either does not exist, or requires a chain of two-to-three calls to accomplish what the REST API does in one hop.

This issue proposes four new read-only tools plus two small quality fixes to existing tools.

### 1. `list-my-actors`

Wraps `GET /v2/acts?my=1`. There is `get-actor-run-list` and `get-dataset-list` for the user's own runs and datasets, but no equivalent for the user's own Actors. Every agent workflow that touches an existing Actor starts by asking \"what Actors do I have?\", and today the only MCP answer is \"call `search-actors` and hope it finds mine in the public store\" (which fails for private Actors).

Suggested shape:

- Inputs: `offset` (default 0), `limit` (max 20, default 10), `desc` (default false), `my` implicit true.
- Output: paginated list mirroring `get-dataset-list` — one JSON blob with `total`, `items[]`, each item minimally `{id, name, username, title, createdAt, modifiedAt, isPublic}`.

### 2. `check-actor-name-available`

Wraps a lookup for `username/actor-name`. Agents scaffolding a new Actor need to know whether the intended technical name collides with an existing one under the same user *before* they run `apify push`. Today this requires either a full `list-my-actors` scan or a `GET /v2/acts/{username}~{name}` and interpreting 404 vs 200 — which the current tool surface does not expose cleanly.

Suggested shape:

- Inputs: `name` (required), `username` (optional; default to authenticated user).
- Output: `{available: boolean, existingActorId?: string}`.

### 3. `list-previous-runs`

Wraps `GET /v2/acts/{actorId}/runs` with the same filters as the existing `get-actor-run-list` (`status`, `desc`, `limit`, `offset`). The existing tool lists runs across the *entire user account*; there is no scoped-to-one-Actor variant. To answer \"what's the latest SUCCEEDED run of *this* Actor?\", agents currently list up to 10 recent runs, then filter client-side by `actId`, which frequently misses if the Actor hasn't been the most active one.

Suggested shape:

[...]

PR #1054, opened 2026-07-04
fix(tools): return structured {found:false} envelope for missing key-value store records

DaveHanns:
## Summary

`get-key-value-store-record` currently returns `isError: true` with a text-only `Record 'X' not found in key-value store 'Y'.` envelope (and no `structuredContent`) when the store exists but the requested key does not. Callers that treat "key missing" as a legitimate branch (probing whether a checkpoint has been cleared, checking for optional metadata, feature-flag records) have to regex-match that text to distinguish it from a real error.

This PR splits the not-found path:

- **Store missing** (`.get()` also returns `undefined`) — unchanged. Still soft-fails with `isError: true` + `INVALID_INPUT` telemetry, because that is a bad input.
- **Key missing in an existing store** — now returns a structured envelope with no `isError`:
  ```json
  { "found": false, "keyValueStoreId": "kv-1", "recordKey": "CHECKPOINT" }
  ```
  The human-readable line ("`Record 'X' not found in key-value store 'Y'.`") is preserved as the tool `summary`, so clients that only render text see the same message.
- **Hit** — unchanged shape.

## Rationale

Consider a common agent pattern: "check whether a checkpoint exists; if it does, resume from there; otherwise start fresh." Under the current behavior, the checkpoint being absent (which is the healthy fast path when work was completed and the checkpoint cleared) surfaces as `isError: true`. Downstream orchestration that fails on tool errors then treats a successful clear as a fault, and text-envelope callers must regex the message to recover.

A boolean `found` in `structuredContent` lets programmatic consumers branch without touching prose, and keeps the "real bad input" (missing store) as a genuine error.

## Evidence

Before this change, a call for a missing key in an existing store yielded (`kv-1` exists, `CHECKPOINT` does not):

```
isError: true
content: [{ type: 'text', text: "Record 'CHECKPOINT' not found in key-value store 'kv-1'." }]
```

with no `structuredContent`. After:

```
isError: (unset)
structuredContent: {
  found: false,
  keyValueStoreId: 'kv-1',
  recordKey: 'CHECKPOINT',
  summary: "Record 'CHECKPOINT' not found in key-value store 'kv-1'."
}
```

Missing-store still returns `isError: true` (see the `returns isError "store not found"` unit test).

## Output schema change

`keyValueStoreRecordOutputSchema` gains an optional `found: boolean` and an optional `recordKey: string`. `key` and `value` are dropped from `required` (they remain present on a hit; they are absent on a miss). This is additive for hit consumers and only removes required-ness — no existing hit response shape changes.

## Test plan

- [x] Existing "record on a hit" tests pass unchanged (JSON, text, empty, binary/image/audio/PDF, over-inline-limit link).
- [x] Renamed unit test asserts the structured miss envelope: `found: false`, `keyValueStoreId`, `recordKey`, no `key`/`value`, summary preserved.
- [x] Existing store-missing soft-fail test remains untouched.
- [x] `pnpm run type-check`, `pnpm run lint`, `pnpm run test:unit` all green.

_Surfaced during an evaluation of Apify surfaces for agent-driven Actor development._

### DIFF
--- a/src/tools/storage/get_key_value_store_record.ts
+++ b/src/tools/storage/get_key_value_store_record.ts
@@ -59,10 +59,27 @@ export const getKeyValueStoreRecord: ToolEntry = Object.freeze({
         if (record === undefined) {
             // getRecord returns undefined for both missing-store and missing-key; disambiguate.
             const storeInfo = await store.get();
-            const text = storeInfo
-                ? `Record '${recordKey}' not found in key-value store '${keyValueStoreId}'.`
-                : `Key-value store '${keyValueStoreId}' not found.`;
-            return buildStorageNotFound(text);
+            if (!storeInfo) {
+                // Missing store is a real bad input — keep the soft-fail so telemetry flags it.
+                return buildStorageNotFound(`Key-value store '${keyValueStoreId}' not found.`);
+            }
+            // Store exists but the key doesn't. This is a legitimate branch case (e.g. a checkpoint
+            // that has been cleared on success); returning `isError: true` forces callers to regex
+            // the text envelope to distinguish it from a real error. Emit a structured
+            // `{found: false, ...}` envelope so agents can branch on the boolean instead.
+            const apifyConsoleUrl = buildConsoleKeyValueStoreUrl(
+                await getConsoleLinkContext(apifyToken, client),
+                keyValueStoreId,
+            );
+            return buildStorageResponse({
+                structuredContent: {
+                    found: false,
+                    keyValueStoreId,
+                    recordKey,
+                },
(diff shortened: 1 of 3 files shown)


jirispilka:
Problem is real — `get_key_value_store_record.ts:62-68` soft-fails for both a missing
key and a missing store, and probing mcp.apify.com returns both on the error channel.

Closing anyway:

- Stale: patches `buildStorageNotFound`, now `respondUserError`. Branch conflicts, lint red.
- `found` is omitted on a hit, so `if (result.found)` is falsy on a successful read.
- The error-channel convention is shared with the other storage tools — decide it on
  #1053, not here. `respondUserError` already takes `structuredContent`, so a structured
  miss doesn't require the `isError` flip.

PR #1045, opened 2026-07-01
refactor: Adopt respond* response constructors across tools

jirispilka:
I'm sorry for a bulky PR but there was no other way.
This PR helps to create a boundary between tool response and MCP response. 

> 📄 **Visual summary:** [The `respond*` response contract](https://claude.ai/code/artifact/384caccf-53c8-413e-8a2b-3ae9031e344c) — a one-screen brief.

Tool handlers built MCP responses by hand, so failure classification was opt-in and telemetry unreliable: `get_dataset_schema` failures carried no `failureCategory` (P2), and the `SOFT_FAIL + INVALID_INPUT` telemetry literal was duplicated across handlers (P3), alongside redundant domain builders and hand-rolled JSON code fences.

## How

A `respond*` family — `respondOk`, `respondJson`, `respondUserError`, `respondServerError`, plus a `respondErrorNoTelemetry` helper for framework paths — is now the only public way to build a tool response. `buildMCPResponse` is un-exported (private to `utils/mcp.ts`); an exported `ToolResponse` alias replaces the `ReturnType` annotations. `respondServerError` derives `toolStatus` / `failureCategory` / `httpStatus` from the caught error; `respondJson` owns the ` ```json ` fence via the existing `wrapJsonText` helper. `buildPermissionApprovalErrorResponse` and `buildStorageNotFound` are deleted and the single-call-site `buildSearchActorsEmptyResponse` is inlined; three helpers with two call sites each (base tool + widget) are kept but now route through `respond*`.

**Client-facing wire output is byte-identical** except the intended telemetry corrections:
- **P2** — `get_dataset_schema` failure is now classified `INTERNAL_ERROR` (was category-less).
- **P3** — the duplicated `SOFT_FAIL + INVALID_INPUT` literal collapses into `respondUserError`.
- Two coherence fixes to previously incoherent status/category pairings: `fetch-apify-docs` thrown-fetch (`SOFT_FAIL` → `FAILED`) and `get-actor-run` error (now carries `INVALID_INPUT`).

## Alternatives considered

Deleting all six domain builders and inlining every call site (the issue's literal wording) would duplicate multi-branch logic across each base+widget pair; keeping the multi-call-site helpers routed through `respond*` preserves DRY while still removing the hand-rolled response bodies. The brand + `ToolEntry.call` narrowing (the compile-time lock) is deferred to a separate sub-issue — `call` is unchanged here.

## Note for `apify-mcp-server-internal`

[...]

Issue #937, opened 2026-06-01
refactor: Adopt respond* response constructors across tools


jirispilka:
Part of #935. One mechanical PR — independent of #936.

## Goal

Introduce the `respond*` family and route every tool response through it. Mechanical substitution; wire output stays byte-identical so `apify-mcp-server-internal` sees no change.

## Scope

- Add to `src/utils/mcp.ts`:
  - `respondOk(text, opts?)` — success with text (`{ structuredContent?, meta? }`)
  - `respondJson(value, opts?)` — success carrying JSON; **owns the ```json fence**
  - `respondEmpty(text)` — success, intentionally no data (thin alias of `respondOk` for now)
  - `respondUserError(text, opts?)` — `SOFT_FAIL`; `category` defaults to `INVALID_INPUT` (`Exclude<…, 'INTERNAL_ERROR'>`), `httpStatus?`, `detail?`, `structuredContent?`
  - `respondServerError(text, opts?)` — `FAILED`; pass the caught `error`, category/httpStatus derived via existing `classifyFailureCategory` / `getHttpStatusCode`
- Migrate all `src/tools` handlers + the inline error paths in `src/mcp/server.ts`.
- Delete the 6 domain builders: `buildActorNotFoundResponse`, `buildSearchActorsEmptyResponse`, `buildPermissionApprovalResponse`, `buildPermissionApprovalErrorResponse`, `buildGetActorRunError`, `buildCallActorErrorResponse`, plus the inline storage not-found literals.
- Un-export `buildMCPResponse` (make it private to `utils/mcp.ts`); shrink `extractToolTelemetry` accordingly.
- Collapse the ~8 hand-rolled `` ```json `` fences into `respondJson`.

## Fixes within this PR

- **P2** — `respondServerError` fills `failureCategory` by default → no more "category unknown".
- **P3** — the 10 duplicated `SOFT_FAIL + INVALID_INPUT` literals collapse into `respondUserError`.

## Out of scope

- The brand + narrowing `ToolEntry.call` (the compile-time lock) — separate sub-issue. `call` stays `Promise<object>` here.
- A real `NO_DATA` telemetry status (P5).

## Acceptance

- No remaining call sites build raw response shapes or inline JSON fences.
- `pnpm run type-check`, `pnpm run lint`, `pnpm run test:unit` clean; assertions migrated to the `respond*` shapes.

PR #1056, opened 2026-07-05
refactor: Brand ToolResponse to lock tool-response contract (#938)

### LINKED ISSUE #938: refactor: Brand ToolResponse and narrow ToolEntry.call (P4 lock)
Part of #935. Depends on #937 (all handlers must route through `respond*` first).

## Goal

Make an unclassified error a **compile error** — closes P4. This is the structural guarantee that distinguishes Option F from a plain helper.

## Scope

- Brand the constructor output in `src/utils/mcp.ts`:

  ```ts
  declare const wireBrand: unique symbol;
  export type ToolResponse = MCPWireShape & { readonly [wireBrand]?: never };
  ```

  The `respond*` functions return `ToolResponse`; raw object literals don't carry the brand.
- Narrow the handler signature in `src/types.ts` (currently `call: (args) => Promise<object>`):

  ```ts
  call: (args: InternalToolArgs) => Promise<ToolResponse>;
  ```

- Fix whatever the compiler flags — those are the spots #937 missed (hand-rolled shapes or stray `isError`).

## Acceptance

- A hand-rolled `{ content: [...] }` or bare `isError: true` in a handler does not compile.
- `pnpm run type-check`, `pnpm run lint`, `pnpm run test:unit` clean.

jirispilka:
## What we're solving

The `respond*` constructors landed in #937, but nothing forced handlers to use them: a hand-rolled `return { content, isError: true }` was an unclassified, telemetry-less response that compiled silently. This is the P4 lock — make that a **compile error**, so the tool-response contract is enforced by the type system instead of review discipline.

## How

`ToolResponse` now carries a module-private **required** brand (a `unique symbol`, never emitted), and `HelperTool.call` returns `Promise<ToolResponse>`. The only way to produce a `ToolResponse` is a `respond*` constructor. `respondRaw` applies the brand to pre-shaped remote/binary content as a pure identity pass-through; `respondAborted` covers cancellation returns. `content` is widened to the MCP `ContentBlock` union so non-text handlers (key-value-store record) keep compiling.

**Wire output is byte-identical** — the brand is a phantom symbol, never serialized; `respondRaw` returns its argument unchanged and the tightened return-type annotations are type-only.

## Alternatives considered

- **Optional brand** (`{ [brand]?: never }`, the issue's original snippet): rejected — `tsc` proves an optional symbol does not reject raw literals, so it fails its own acceptance criterion. A required brand is necessary.
- **A rebuilding constructor for binary blocks**: rejected — it would inject `isError` and reorder keys, changing serialized bytes. `respondRaw` preserves them exactly.
- `ToolResponse` is written as an explicit transcription of the wire shape rather than `ReturnType<typeof buildMCPResponse>`, because the builder now casts to `ToolResponse`, making that derivation circular.

#937 (PR #1045) has since merged to `master`; this PR is rebased onto `master` and contains only the #938 lock. Part of #935. Closes #938.

https://claude.ai/code/session_01PLMFGqUuBGat7FPxWGhsRD

### DIFF
--- a/src/utils/mcp.ts
+++ b/src/utils/mcp.ts
@@ -1,3 +1,5 @@
+import type { CallToolResult, ContentBlock } from '@modelcontextprotocol/sdk/types.js';
+
 import { FAILURE_CATEGORY, TOOL_STATUS } from '../const.js';
 import type { AjvErrorDetails, FailureCategory, ToolTelemetryContext } from '../types.js';
 import { ACTOR_RUN_LIMIT_MESSAGE, isActorRunLimitError } from './apify_errors.js';
@@ -50,14 +52,28 @@ function buildMCPResponse(options: {
         ...(telemetry && { toolTelemetry: telemetry }),
         ...(structuredContent !== undefined && { structuredContent }),
         ...(_meta !== undefined && { _meta }),
-    };
+    } as unknown as ToolResponse;
 }
 
 /**
- * Shared return type of the `respond*` constructors. Replaces the
- * `ReturnType<typeof buildMCPResponse>` annotations now that `buildMCPResponse` is private.
+ * Module-private brand. Phantom `declare`d symbol — never assigned or emitted at runtime, so it adds
+ * nothing to the wire. Its only job is to make `ToolResponse` unforgeable at compile time.
  */
-export type ToolResponse = ReturnType<typeof buildMCPResponse>;
+declare const toolResponseBrand: unique symbol;
+
+/**
+ * Shared return type of the `respond*` constructors and of every `HelperTool.call`. The required brand
+ * means a raw `{ content, isError }` literal (or a bare `{}`) fails to compile — the only way to produce a
+ * `ToolResponse` is a constructor here. `content` is the full MCP `ContentBlock` union so image/audio/
+ * resource returns type-check.
+ */
+export type ToolResponse = {
+    content: ContentBlock[];
+    isError: boolean;
+    toolTelemetry?: ToolTelemetryContext;
+    structuredContent?: unknown;
+    _meta?: Record<string, unknown>;
+} & { readonly [toolResponseBrand]: never };
 
 /** Normalise a `string | string[]` text argument to the array `buildMCPResponse` expects. */
 function toTexts(texts: string | string[]): string[] {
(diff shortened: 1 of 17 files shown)


vojtechj-apify:
I will look into it right now

src/utils/mcp.ts
```diff
@@ -168,6 +184,24 @@ export function respondErrorNoTelemetry(
...
+    return result as unknown as ToolResponse;
+}
+
+/**
+ * Empty response for MCP cancellation paths — per spec, receivers SHOULD NOT reply to a cancelled request.
+ * Returns runtime `{}`, branded.
+ */
+export function respondAborted(): ToolResponse {
+    return {} as unknown as ToolResponse;
+}
```
vojtechj-apify, line 207: **low:** The branded type over-promises here — `ToolResponse` declares `content: ContentBlock[]` and `isError: boolean` as required, but both escape hatches violate that at runtime:

- `respondAborted()` returns `{}` — both fields missing.
- `respondRaw()` accepts `CallToolResult`, where `isError` is optional — and the live call sites (`get_actor_run_log.ts:48`, `get_actor_run_list.ts:67`, `get_key_value_store_record.ts:96/112/118/126`) omit it, so `.isError` is `undefined` behind a type that says `boolean`.

Nothing breaks today — every current consumer either erases the type (`as Record<string, unknown>` in server.ts) or guards defensively, and the wire bytes are identical to master. But it's exactly the contract the PR title claims to lock: the first future consumer that trusts `res.content.map(...)` or `res.isError === false` compiles clean and then throws on the abort path or silently mis-branches (`undefined === false`).

Suggested fix, either:
1. make `content`/`isError` optional on `ToolResponse`, documenting that only the abort/raw paths omit them, or
2. give `respondAborted` a narrower branded `AbortedResponse` type unioned into `HelperTool.call`'s return.

tests/unit/tools.call_actor_common.test.ts
```diff
@@ -67,7 +67,7 @@ describe('call_actor_common', () => {
             });
 
             expect(response.isError).toBe(true);
-            const allText = response.content.map((c) => c.text).join('\n');
+            const allText = response.content.map((c) => ('text' in c ? c.text : '')).join('\n');
```
vojtechj-apify: **nit:** `('text' in c ? c.text : '')` is repeated 6× in this file (plus the `firstBlock && 'text' in firstBlock ? ...` variant at L282), while `utils.mcp.test.ts:541` in this same PR introduces a `textOf()` helper for exactly this. A local one-line `textOf` here would remove the noise.

vojtechj-apify (commented):
Ran a review pass over the diff — type-check, lint, and unit tests all pass, and the brand pattern itself is correctly implemented (phantom `declare`d symbol, nothing emitted at runtime, locked both ways by the tests). Two small findings below, one **low** and one **nit**. Also verified some things that are *not* issues: the `as ContentBlock[]` cast in `call_actor.ts` is genuinely required (the SDK's `callTool` return type widens `content` to `unknown`), and `respondRaw` call sites are byte-identical to master's output.

vojtechj-apify (approved):
Looks good, just check the low issue with ToolRespponse declaration please.

src/utils/mcp.ts
jirispilka: Good catch! Fixed

tests/unit/tools.call_actor_common.test.ts
jirispilka: Also good one! Fixed

MQ37 (approved):
Great PR!

PR #1060, opened 2026-07-07
chore: Remove TOON format

### LINKED ISSUE #1059: Remove TOON format
Following our TOON evaluation we have decided to remove it from our codebase.

RobertCrupa:
## Why
Following the TOON evaluation, remove the TOON encoding from the codebase (closes #1059). The measured token/byte cost was a net loss or wash for real Apify storage payloads, so the format is not worth maintaining.

## What

The five list tools (get-dataset-items, get-dataset-list, get-key-value-store-list, get-key-value-store-keys, get-actor-run-list) now ship raw JSON.stringify in content[].text instead of a toon fence — the same shape the sibling single-object storage tools already use. structuredContent (the authoritative, outputSchema-validated channel) is unchanged, so programmatic consumers are unaffected.

- Delete src/utils/encode_text.ts (encodeToon, dotFlatten, wrapJsonText, FENCES; no non-test consumers remained)
- Drop the toon? param from buildStorageResponse; always JSON.stringify
- Remove @toon-format/toon dependency
- Delete utils.encode_text.test.ts and the now-redundant nested-items dot-flatten test; gut the fenced-text test helpers in favor of JSON.parse

## Testing

image

### Note

Note for apify-mcp-server-internal: content[].text format changed for these five tools (TOON fence -> raw JSON). structuredContent is unchanged; verify the contract/smoke suite does not assert on the fenced shape.


MQ37 (approved):
Thank you sir! 😎

jirispilka (approved):
Nice!

Only three nits that would be nice to fix:

Stale comments (minor)  three leftover comments still say "fenced data" in [tests/unit/tools.get_dataset_items.test.ts](https://github.com/apify/apify-mcp-server/blob/master/tests/unit/tools.get_dataset_items.test.ts) (~L190, ~L208) and [tests/unit/tools.get_key_value_store_keys.test.ts](https://github.com/apify/apify-mcp-server/blob/master/tests/unit/tools.get_key_value_store_keys.test.ts) (~L184), even though the code now emits plain JSON.

Issue #1066, opened 2026-07-08
chore: Quick-win cleanups from codebase sweep


jirispilka:
Small, independent cleanups from the 2026-07-08 sweep (full report: `res/refactoring-sweep-2026-07.md`). Each checkbox = one small PR (15–60 min). They share no state — pick off in any order.

- [x] **`getTaskOrThrow` helper.** Three task handlers copy-paste the `getTask` → `softFail` 404 → `throw McpError(InvalidParams, 'Task ... not found')` block (`server.ts:714-726`, `:729-756`, `:759-790`), differing only in log tag.
- [x] **Dedup the mode-resolution gate** in `loadToolsByName` / `loadToolsFromUrl` / `loadToolsFromInput` (`server.ts:450-535`): identical buffer-or-compose block with three divergent `notify` booleans to `upsertTools` (the inconsistency needs the explanatory comment at `:362-367`). Extract one helper with an explicit `notify` param. — #1078
- [x] **Memoize `resolveWidgets()`** (`server.ts:1862-1904`): dynamic imports + fs scan of `web/dist` run on every `initialize` AND every `connect` — per HTTP session. Widgets are static build artifacts; resolve once. — #1076
- [x] **`catchNotFound(promise)` helper** in storage helpers: the same 8-line 404→null `.catch` block + near-verbatim SDK-quirk comment in `get_dataset_items.ts:120-128`, `get_dataset_schema.ts:63-72`, `get_key_value_store_keys.ts:62-70`. — #1072
- [x] **Hoist search-actors guidance strings**: broaden-your-search text written twice inside `search_actors.ts` (`:184-191` structured `instructions`, `:205-212` text footer) and the empty-results block duplicated verbatim in `widgets/search_actors_widget.ts:76-80`. These strings get tuned often — one const each. — #1075
- [x] **Shared `injectMcpSessionId(params, sessionId)` helper**: same `params._meta ??= {}; params._meta.mcpSessionId = ...` injection in `stdio.ts:238-250` and `dev_server.ts:211-214`; the "always create params" invariant is commented in only one of them. — #1074
- [x] **Split the `widget` boolean param**: `buildStartRunResponse(..., widget?)` (`actor_run_response.ts:715`) and `buildGetActorRunSuccessResponse(..., widget)` (`get_actor_run.ts:92`) branch internally on a flag callers always pass as a literal; the "widget only from `*-widget` tools" invariant is comment-enforced. Split into base builder + widget wrapper. — #1077
[...]

PR #1076, opened 2026-07-10
feat: Memoize widget resolution to scan web/dist once

jirispilka:
Part of #1066 (checkbox: memoize `resolveWidgets()`).

`ActorsMcpServer.resolveWidgets()` ran two dynamic imports and an `fs.existsSync` scan of `web/dist` (via `resolveAvailableWidgets`) on **every** `initialize` **and** every `connect()` — i.e. once per HTTP session. Widgets are static build artifacts and the scan depends only on a constant `dirName` (from `import.meta.url`), so the result never changes within a process.

https://claude.ai/code/session_016pXZvhEoE6Lz3DpBSL3n1Y

---
_Generated by [Claude Code](https://claude.ai/code/session_016pXZvhEoE6Lz3DpBSL3n1Y)_

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -167,6 +167,60 @@ export async function emitTaskStatusNotification(
     }
 }
 
+/** Module-level cache for resolved widgets. Shared across all instances in the process. */
+let cachedWidgetsPromise: Promise<Map<string, AvailableWidget>> | undefined;
+
+/**
+ * Load and cache available widgets once per process. Concurrent callers share the
+ * in-flight promise. On rejection, clears the cache to allow retry.
+ */
+async function loadAvailableWidgetsOnce(): Promise<Map<string, AvailableWidget>> {
+    if (cachedWidgetsPromise) {
+        return cachedWidgetsPromise;
+    }
+
+    cachedWidgetsPromise = (async () => {
+        const { fileURLToPath } = await import('node:url');
+        const path = await import('node:path');
+
+        const filename = fileURLToPath(import.meta.url);
+        const dirName = path.dirname(filename);
+
+        const resolved = await resolveAvailableWidgets(dirName);
+
+        const readyWidgets: string[] = [];
+        const missingWidgets: string[] = [];
+
+        for (const [uri, widget] of resolved.entries()) {
+            if (widget.exists) {
+                readyWidgets.push(widget.name);
+            } else {
+                missingWidgets.push(widget.name);
+                log.softFail(`Widget file not found: ${widget.jsPath} (widget: ${uri})`);
+            }
+        }
+
+        if (readyWidgets.length > 0) {
+            log.debug('Ready widgets', { widgets: readyWidgets });
+        }
+
+        if (missingWidgets.length > 0) {
+            log.softFail('Some widgets are not ready', {
+                widgets: missingWidgets,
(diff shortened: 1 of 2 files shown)


MQ37:
I think this PR is not worth it as the node already does the caching of the modules.

**Clanker:**

Looked into this closely, including a benchmark against the real `web/dist` in this repo — sharing what I found in case it changes the scope.

**What's actually being cached:** `resolveAvailableWidgets()`'s scan of `WIDGET_REGISTRY` (currently 2 entries) — i.e. 2 `fs.existsSync()` calls. Measured steady-state (Node 24, 10k iterations, warm cache): **~8.1µs total**. For comparison, on the same machine:

| Operation | Per-call cost |
|---|---|
| The fs scan this PR caches | 8.14 µs |
| A single `log.debug()` write (real `@apify/log`, level DEBUG) | 8.41 µs |
| `crypto.randomUUID()` (already runs per-session, uncached, elsewhere in `server.ts`) | 2.44 µs |
| `JSON.stringify`+`parse` of a representative `initialize` payload | 1.19 µs |

The scan costs about the same as one debug-log line already written unconditionally a few lines above it in the same handler. It's below the noise floor of anything else `initialize`/`connect` does (TLS handshake, JSON-RPC parsing). Also worth noting: Node caches `import()` by resolved specifier automatically (confirmed via MDN) — the dynamic `import('node:fs')`/`node:path`/`node:url` calls this PR's description blames were never actually expensive, cached or not.

A few concrete issues on top of the proportionality question:

1. **Dead retry branch.** The `.catch()` at the end of `loadAvailableWidgetsOnce()` clears the cache "to allow retry on next session" — but nothing in the call chain can actually reject. `fs.existsSync()` never throws (returns `false` on any input), and dynamic imports of builtins can't fail in a live process. This path is also untested.
2. **Self-healing regression.** Pre-PR, every session re-scanned `web/dist`, so if the build artifacts land after process boot (deploy race, `postbuild`'s `|| :` swallowing copy failures), the next session would just pick them up. Post-PR, a "missing" result is cached for the life of the process — recovery needs a restart. Not mentioned in the PR description.
3. **Module-global cache scope.** `cachedWidgetsPromise` is shared by every `ActorsMcpServer` instance in the process (previously `availableWidgets` was instance-local). No reset seam — the new test only stays isolated because Vitest defaults to per-file module isolation, not because anything in the code enforces it.
4. **Log-cadence change is the one real win, but it's scoped too large.** Moving "ready/missing widgets" logging from per-session to per-process is legitimate (real log-volume reduction on a hosted server). That's a 3-6 line boolean guard around the logging block alone — it doesn't need a promise cache, retry logic, or a 104-line test to get there.

Given the actual cost being optimized doesn't show up against a single log write, I'd suggest closing this and, if the log-volume reduction is still wanted, replacing it with just that: a `let hasLoggedWidgetStatus = false` guard around the existing logging loop in `resolveWidgets()`, leaving the scan itself unchanged (cheap, self-healing, per-session).

Happy to be wrong here — if there's a session-churn number or profile that shows this mattering in practice, that changes the calculus. Didn't see one attached to #1066 or this PR.

MQ37 (changes requested):
⏫

jirispilka:
Thanks! Good point.

I've re-checked it and there is indeed no benefit. It was claude's hallucination.

Closing this.

PR #1077, opened 2026-07-10
refactor: Split widget flag into base + widget response builders

jirispilka:
Part of #1066 (checkbox: split the `widget` boolean param).

`buildStartRunResponse` and `buildGetActorRunSuccessResponse` each took a `widget` boolean and branched internally. Every caller passed a literal — widget tools `true`, everything else `false`/omitted — so the "widget only from `*-widget` tools" rule was enforced by a comment, not the type system.

https://claude.ai/code/session_016pXZvhEoE6Lz3DpBSL3n1Y

---
_Generated by [Claude Code](https://claude.ai/code/session_016pXZvhEoE6Lz3DpBSL3n1Y)_

### DIFF
--- a/src/tools/actors/actor_run_response.ts
+++ b/src/tools/actors/actor_run_response.ts
@@ -4,7 +4,7 @@ import log from '@apify/log';
 
 import type { ApifyClient } from '../../apify_client.js';
 import { DATASET_SIZE_HINT_BYTES, HELPER_TOOLS, NARROW_OUTPUT_HINT } from '../../const.js';
-import { getWidgetConfig, WIDGET_URIS } from '../../resources/widgets.js';
+import { buildActorRunWidgetMeta } from '../../resources/widgets.js';
 import type { ConsoleLinkContext } from '../../types.js';
 import {
     buildConsoleDatasetUrl,
@@ -703,13 +703,10 @@ async function waitForRunWithProgress(opts: {
 
 /**
  * Shared construction for the immediate start response, used by both the base and widget
- * builders below. Returns `base` (structuredContent minus `nextStep`), the computed `summary`,
- * and `computedNextStep` (the non-widget nextStep) — each builder picks its own `nextStep`/meta.
+ * builders below. Returns the full `RunResponse` with the computed (non-widget) `nextStep`;
+ * `buildStartRunWidgetResponse` overrides `nextStep` on its own copy.
  */
-function buildStartRunSharedContent(
-    actorName: string,
-    actorRun: ActorRun,
-): { base: Omit<RunResponse, 'nextStep'>; summary: string; computedNextStep: string } {
+function buildStartRunSharedContent(actorName: string, actorRun: ActorRun): RunResponse {
     // Start path returns before any metadata fetch, so every entry — default and aliases — is id-only.
     const datasetIds = buildStorageAliasIds(actorRun.storageIds?.datasets, actorRun.defaultDatasetId ?? undefined);
     const kvIds = buildStorageAliasIds(
@@ -723,13 +720,13 @@ function buildStartRunSharedContent(
         Object.fromEntries(Object.entries(kvIds).map(([alias, id]) => [alias, { id }])),
     );
 
-    const { summary, nextStep: computedNextStep } = buildStatusSummaryNextStep({
+    const { summary, nextStep } = buildStatusSummaryNextStep({
         run: actorRun,
         dataset: datasets?.default,
         keyValueStore: keyValueStores?.default,
     });
 
-    const base: Omit<RunResponse, 'nextStep'> = {
+    return {
         runId: actorRun.id,
         actorId: actorRun.actId,
         actorName,
@@ -740,9 +737,8 @@ function buildStartRunSharedContent(
             ...(keyValueStores && { keyValueStores }),
         },
         summary,
+        nextStep,
     };
-
-    return { base, summary, computedNextStep };
 }
 
 /**
@@ -757,16 +753,16 @@ export function buildStartRunResponse(params: {
 }): ToolResponse {
     const { actorName, actorRun, linkContext } = params;
 
-    const { base, summary, computedNextStep } = buildStartRunSharedContent(actorName, actorRun);
-    const nextStep = computedNextStep;
-
-    const structuredContent: RunResponse = { ...base, nextStep };
+    const structuredContent = buildStartRunSharedContent(actorName, actorRun);
     const consoleLinks = applyConsoleLinks(structuredContent, linkContext);
 
-    return respondOk([JSON.stringify(structuredContent), `${summary}\n${nextStep}${consoleLinks}`], {
-        structuredContent,
-        meta: undefined,
-    });
+    return respondOk(
+        [
+            JSON.stringify(structuredContent),
+            `${structuredContent.summary}\n${structuredContent.nextStep}${consoleLinks}`,
+        ],
+        { structuredContent },
+    );
 }
 
 /**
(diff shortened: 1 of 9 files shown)


src/tools/actors/actor_run_response.ts
```diff
@@ -752,16 +740,57 @@ export function buildStartRunResponse(params: {
...
+    actorRun: ActorRun;
+    linkContext?: ConsoleLinkContext;
+}): ToolResponse {
+    const { actorName, actorRun, linkContext } = params;
+
+    const { base, summary } = buildStartRunSharedContent(actorName, actorRun);
+    const nextStep = WIDGET_NO_POLL_NEXT_STEP;
+
+    const structuredContent: RunResponse = { ...base, nextStep };
     const consoleLinks = applyConsoleLinks(structuredContent, linkContext);
```
vojtechj-apify: `linkContext` and the `applyConsoleLinks` call here are dead code:

- The only caller, `call-actor-widget` (`call_actor_widget.ts:123`), never passes `linkContext`. With it undefined, `applyConsoleLinks` returns `''` at its early return (line 319) and mints no `apifyConsoleUrl`.
- The widget UI never reads `apifyConsoleUrl` — it builds the Console URL client-side from `runId` (`ActorRun.tsx:578`: `https://console.apify.com/actors/runs/${runData.runId}`).

So `linkContext`, the `applyConsoleLinks` call, and the `${consoleLinks}` interpolation can be dropped — output is byte-identical, and it lines up with `buildGetActorRunWidgetResponse`, which already omits console links.

```ts
export function buildStartRunWidgetResponse(params: { actorName: string; actorRun: ActorRun }): ToolResponse {
    const { actorName, actorRun } = params;

    const { base, summary } = buildStartRunSharedContent(actorName, actorRun);
    const nextStep = WIDGET_NO_POLL_NEXT_STEP;
    const structuredContent: RunResponse = { ...base, nextStep };

    const widgetMeta = {
        ...(getWidgetConfig(WIDGET_URIS.ACTOR_RUN)?.meta ?? {}),
        'openai/widgetDescription': `Actor run progress for ${actorName}`,
    };

    return respondOk([JSON.stringify(structuredContent), `${summary}\n${nextStep}`], {
        structuredContent,
        meta: widgetMeta,
    });
}
```

---

jirispilka: Good catch! confirmed dead fixed. Thanks

src/tools/actors/actor_run_response.ts
```diff
@@ -752,18 +740,54 @@ export function buildStartRunResponse(params: {
...
     };
-    const consoleLinks = applyConsoleLinks(structuredContent, linkContext);
 
-    const widgetMeta = widget
-        ? {
-              ...(getWidgetConfig(WIDGET_URIS.ACTOR_RUN)?.meta ?? {}),
-              'openai/widgetDescription': `Actor run progress for ${actorName}`,
-          }
-        : undefined;
+    return { base, summary, computedNextStep };
```
vojtechj-apify: Nit (optional): the tri-tuple return leaves some refactor residue. `summary` is already `base.summary`, so it's returned twice, and both builders then re-assemble `{ ...base, nextStep }` and alias `const nextStep = computedNextStep` (line 761).

Cleaner: have this helper return a finished `RunResponse` with `nextStep = computedNextStep`. Then `buildStartRunResponse` uses it verbatim and `buildStartRunWidgetResponse` spreads-and-overrides — the exact one-line pattern already used in `get_actor_run.ts` (`{ ...structuredContent, nextStep: WIDGET_NO_POLL_NEXT_STEP }`). That removes the `Omit<>`, the duplicate `summary`, and both `nextStep` aliases in one go.

---

src/tools/actors/actor_run_response.ts
```diff
@@ -752,18 +740,54 @@ export function buildStartRunResponse(params: {
...
+ */
+export function buildStartRunWidgetResponse(params: { actorName: string; actorRun: ActorRun }): ToolResponse {
+    const { actorName, actorRun } = params;
+
+    const { base, summary } = buildStartRunSharedContent(actorName, actorRun);
+    const nextStep = WIDGET_NO_POLL_NEXT_STEP;
+
+    const structuredContent: RunResponse = { ...base, nextStep };
+
+    const widgetMeta = {
```
vojtechj-apify: Nit (optional): this widget `_meta` block is now duplicated in two files — here and `get_actor_run.ts` (`buildGetActorRunWidgetResponse`). Both build the ACTOR_RUN widget config + `openai/widgetDescription`, and they already differ (the get-run copy also merges `buildUsageMeta` and sources the name as `structuredContent.actorName ?? runId` vs `actorName` here).

Consider a shared `buildActorRunWidgetMeta(...)` (natural home: `src/resources/widgets.ts`). Otherwise a change to the description format or an `openai/*` key has to be made in both places, and the two ACTOR_RUN widgets can drift silently.

---

tests/unit/tools.get_actor_run.response.test.ts
```diff
@@ -731,11 +732,10 @@ describe('buildStartRunResponse()', () => {
         expect(content[1].text).toContain(VERBATIM_LINKS_NUDGE);
     });
 
-    it('includes widget metadata and no-poll nextStep when widget=true', () => {
-        const result = buildStartRunResponse({
+    it('buildStartRunWidgetResponse includes widget metadata and a no-poll nextStep', () => {
```
vojtechj-apify: Convention: this tests `buildStartRunWidgetResponse` but lives under `describe('buildStartRunResponse()')` with a function-name-prefixed `it()` name.

CLAUDE.md → Testing: *"`describe('fnName()')`, plain-verb `it()` names (no `should` prefix). Group with nested `describe()` per method…"*. Move it into its own `describe('buildStartRunWidgetResponse()')` block and drop the prefix, e.g. `it('includes widget metadata and a no-poll nextStep')`.

---

src/tools/actors/actor_run_response.ts
```diff
@@ -752,18 +740,54 @@ export function buildStartRunResponse(params: {
...
+
+    const { base, summary, computedNextStep } = buildStartRunSharedContent(actorName, actorRun);
+    const nextStep = computedNextStep;
+
+    const structuredContent: RunResponse = { ...base, nextStep };
+    const consoleLinks = applyConsoleLinks(structuredContent, linkContext);
 
     return respondOk([JSON.stringify(structuredContent), `${summary}\n${nextStep}${consoleLinks}`], {
+        structuredContent,
+        meta: undefined,
```
vojtechj-apify: Nit (optional): `meta: undefined` is redundant here. `respondOk`'s `meta` is optional and `buildMCPResponse` omits `_meta` when it's `undefined`, so this is equivalent to just `{ structuredContent }`. It's a leftover of the old conditional `meta: widgetMeta` — can drop it.

---

src/tools/runs/get_actor_run.ts
```diff
@@ -86,25 +86,30 @@ export function buildGetActorRunError(runId: string, error: unknown): ToolRespon
...
+        ],
+        { structuredContent, meta: buildUsageMeta(run) },
+    );
+}
+
+/**
+ * Build the widget success response. `content[1]` carries a short pointer instead of the
+ * summary/nextStep narrative. Used only by `*-widget` tools; does not apply console links.
+ */
+export function buildGetActorRunWidgetResponse(params: FetchActorRunResult): ToolResponse {
```
vojtechj-apify, line 111: Naming (optional / possibly out of scope): the get-run pair mixes two axes — `buildGetActorRunSuccessResponse` keeps `Success` while its new widget sibling is `buildGetActorRunWidgetResponse`. The start pair uses the cleaner `buildStartRunResponse`/`buildStartRunWidgetResponse` axis. Aligning them (`buildGetActorRunResponse`/`buildGetActorRunWidgetResponse`) reads more consistently, but it renames a pre-existing function, so probably a separate change.

---

vojtechj-apify (commented):
Reviewed as a behavior-preserving refactor — no correctness issues. Traced all four builders against the pre-split code: identical wire output (`structuredContent`, `content[0]/[1]`, `_meta`, key order, `nextStep` overrides). The dropped `applyConsoleLinks` call in the widget start builder is a genuine no-op (the widget caller never passed `linkContext`, and the fn returns `''` without mutating when it's undefined). The "widget only from `*-widget` tools" rule is now type-enforced rather than comment-enforced, which is the nice win here.

Left 5 inline comments — all optional cleanup/convention nits, nothing blocking:
1. `buildStartRunSharedContent` tri-tuple return leaves residue (dup `summary`, `nextStep` aliases) — could return a finished `RunResponse`.
2. Widget `_meta` block duplicated across the two widget builders — candidate for a shared helper.
3. Widget test placed under the wrong `describe()` with a non-plain-verb name (CLAUDE.md testing convention).
4. `meta: undefined` is redundant.
5. Get-run builder naming mixes axes (`…SuccessResponse` vs `…WidgetResponse`) — minor, likely separate.

LGTM to merge as-is.

---

vojtechj-apify (approved):
All I find were just nits, approaved!

src/tools/actors/actor_run_response.ts
jirispilka: Nice, fixed, the code looks way better now!

tests/unit/tools.get_actor_run.response.test.ts
jirispilka: Fixed

src/tools/runs/get_actor_run.ts
jirispilka: Fixed as well!

jirispilka:
> All I find were just nits, approaved!

Thanks for all the comments! Good points

src/tools/actors/actor_run_response.ts
jirispilka: Fixed

Issue #1063, opened 2026-07-08
refactor: Extract dispatchToolCall with exhaustive TOOL_TYPE switch


jirispilka:
Part of #658 (step 4, final). Depends on #1062 (error mapper) and #684 (pre-flight short-circuit — removes the standby/402 re-checks from the task path first, shrinking this extraction).

## Plan

- [ ] Extract the tool-type dispatch as a **private method** on `ActorsMcpServer` — no `ctx` bag, `this` provides `apifyMcpServer`/`mcpServer`:

  ```
  dispatchToolCall(params: {
      tool, toolArgs,
      progressTracker,          // constructed by caller
      abortSignal,              // sync: extra.signal / task: cancelWatcher.signal
      forwardNotifications,     // sync: true / task: false
  }) → { result, toolStatus, callDiagnostics }
  ```

- [ ] **No mode flag.** Parameterize by capability, not by mode: caller constructs the progress tracker (sync opt-in for `call-actor`/`get-actor-run`, task always), passes its abort signal, sets `forwardNotifications` (task false — the original request is already answered, so forwarding ACTOR_MCP proxy notifications against its `progressToken` is invalid). Dispatch never knows whether it runs sync or task, so a future third mode cannot fork it.
- [ ] **Exhaustive `switch` on TOOL_TYPE with a `never` default** — folds ACTOR_MCP into the shared core and makes any future missing branch a compile error, not a silent `{}` result.
- [ ] **Returns, never throws.** Execution errors normalized via the #1062 mapper; validation `McpError`s stay pre-dispatch in the sync handler.
- [ ] Rewire both callers. #1061 suite stays green; flip the ACTOR_MCP-gap test to expect real dispatch.
- [ ] Optional, same PR: rename `executeToolAndUpdateTask` → `runToolCallAsTask`, sync post-validation → `runToolCallSync`.

## Verification

Human: `test:integration` + mcpc smoke (sync + task calls). Flag for the internal contract suite — result shapes/`_meta` must not drift.

## Fallback

If this stalls: stop after #1062 and add exhaustive switches to both existing ladders instead — the ACTOR_MCP trap is closed at compile time, two ladders survive. Acceptable steady state per #658.

Issue #1061, opened 2026-07-08
test: Pin sync/task tool-call contracts before dispatch dedup


jirispilka:
Part of #658 (step 1). Behavior-pinning tests before any extraction. Line refs as of 2026-07-08.

## Already covered at helper level — don't duplicate

- Abort isolation (request abort must not cancel task): `tests/unit/mcp.utils.test.ts` (cancel watcher ignores unrelated signals)
- Sync progress-tracker opt-in (`call-actor`/`get-actor-run` only): `tests/unit/mcp.server.progress_wiring.test.ts`
- Outer-catch wire shape on abort: `tests/unit/mcp.server.error_handling.test.ts`
- Classification primitives: `tests/unit/utils.tool_status.test.ts`, `tests/unit/telemetry.test.ts`

## Missing — pin these, handler-level, table-driven

- [ ] Task terminal-status mapping per failure class: 402 and permission-approval stored as `completed`; generic execution errors as `failed` + `internalToolStatus` (task catch in `executeToolAndUpdateTask`, `server.ts:1684-1748` vs `server.ts:1750-1815`)
- [ ] Telemetry property shapes per failure class. No CI guard exists anywhere for these: the internal repo consumes none of the tool-call telemetry shapes in code (verified 2026-07-08) — only analytics dashboards do, so drift is silent. This suite is the only guard.
- [ ] Sync handler 402 / permission-approval result shapes (sync outer catch, `server.ts:1258-1284`)
- [ ] taskSupport gate: task-augmented call to a tool without `taskSupport` is rejected pre-dispatch (`server.ts:986-998`)
- [ ] Task path with an ACTOR_MCP-typed tool stores empty `{}` — documents the latent gap the dispatch extraction closes; flip this test to expect real dispatch when it lands

Vitest, scoped to `src/mcp/**`. Success paths break loudly — cover last or skip.

## Related

#907 is the integration-level counterpart (same input through both sync and task mode end-to-end); this issue pins the unit-level handler contracts.

Issue #658, opened 2026-04-10
[spec] Split src/mcp/server.ts into smaller modules


jirispilka:
## Problem

- [`src/mcp/server.ts`](https://github.com/apify/apify-mcp-server/blob/master/src/mcp/server.ts) is a single ~1.4k-line `ActorsMcpServer` class: constructor setup, telemetry, logging, **`CallToolRequestSchema`** (largest block), prompts, resources, task protocol handlers, and **`executeToolAndUpdateTask`**.

## Proposed solution

**Phase 1 — mechanical split only (behavior and public API unchanged)**

- Keep [`src/index.ts`](https://github.com/apify/apify-mcp-server/blob/master/src/index.ts) exporting **`ActorsMcpServer`** as today; no new package surface.
- Add a small directory under `src/mcp/` (e.g. `src/mcp/handlers/`) for wiring helpers.
- Move bodies of existing **`setupTaskHandlers`**, **`setupPromptHandlers`**, **`setupResourceHandlers`** into modules that register the same `setRequestHandler` calls (same schemas, same logic).
- Move **`CallToolRequestSchema`** handling into a dedicated module; pass an explicit **context** object (e.g. `Server` instance, `options`, `tools` map, `taskStore`, `actorExecutor`, telemetry helpers) instead of growing `this` usage across files.
- Follow the same layering idea as the SDK: low-level [`Server`](https://github.com/modelcontextprotocol/typescript-sdk) + `setRequestHandler`; **do not** migrate to high-level `McpServer.registerTool` (dynamic `ToolEntry` / AJV / payments / Actor MCP do not map cleanly).

**Phase 2 — separate PR (optional)**

- Extract **small shared helpers** used by both sync `tools/call` and **`executeToolAndUpdateTask`** (e.g. internal / actor execution slices). Avoid one mega-merge of control flow; preserve task store transitions and cancellation behavior.

## Plan

**Phase 1**

- [ ] Create `src/mcp/server/` (or `handlers/`) and move `setupTaskHandlers` / `setupPromptHandlers` / `setupResourceHandlers` implementations behind `register*`-style functions; call them from `ActorsMcpServer` constructor.
- [ ] Extract `CallToolRequestSchema` handler (+ tightly coupled private helpers if moved together) into e.g. `call_tool_handler.ts` with explicit `ctx`.
- [ ] Run `npm run type-check`, `npm run lint`, `npm run test:unit`.
[...]

PR #1081, opened 2026-07-11
refactor: Extract shared tool-call error mapper

### LINKED ISSUE #1062: refactor: Extract shared tool-call error mapper (dedup twin catch blocks)
Part of #658 (step 3). First of two extraction steps; near-zero behavioral risk. Depends on #1061 (tests). Line refs as of 2026-07-08.

The twin catch blocks are the largest textual duplication in `server.ts`: sync `server.ts:1252-1322`, task `server.ts:1664-1815`. Both implement the same three-way mapping: 402 → x402 payload, permission-approval → approval result, generic → `classifyFailureCategory` + isError result.

## Plan

- [ ] Extract `mapToolCallError(error) → { kind: 'payment' | 'approval' | 'execution', result, failureCategory }`. Pure function, never throws.
- [ ] Both catches call it; each keeps its own sink. Sync returns the result; task maps kind → terminal status: `payment`/`approval` → `completed`, `execution` → `failed` + `internalToolStatus`.
- [ ] `McpError` re-throw (sync validation → protocol error) stays in the sync handler, before the mapper.
- [ ] #1061 suite stays green; telemetry property shapes unchanged.

Do NOT touch the dispatch ladders in this PR — that is the next step.

## Follow-up (separate PR, after this lands)

Payment seam closure: once the mapper exists, move the x402 knowledge behind `PaymentProvider` (a `buildPaymentRequiredResult(error)` hook) so `server.ts` stops branching on `isX402PaymentRequiredError` directly, and unify the duplicated `payment-required` header constant + decode (`payments/x402.ts` vs `utils/payment_errors.ts`). Details: `res/refactoring-sweep-2026-07.md` → "Payment seam closure".

jirispilka:
## Why

Closes #1062. The sync (`CallToolRequestSchema`) and task (`executeToolAndUpdateTask`) catch blocks in `server.ts` classified tool-call errors — 402 payment, permission-approval, generic execution — with the same logic written out twice, so any classification change had to be hand-synced across both paths. Step 3 of #658; unblocks #1063.

## What changed

Both catches now call one pure `buildToolCallErrorResult()` that returns the classification (kind + tool status + diagnostics + either a ready-to-send response or the user-facing text). Each catch keeps its own sink, so the deliberate sync/task asymmetries are preserved unchanged: sync spread-merges diagnostics onto pre-existing fields while task flat-overwrites; sync's 402 branch logs nothing while task's does; the McpError re-throw stays sync-only; terminal task-status mapping and the wire fields (`toolTelemetry.toolStatus` vs `internalToolStatus`) stay in the sinks. No wire, telemetry, or task-status behavior changes.

One non-obvious point worth a close look: the McpError re-throw was hoisted above the payment/approval checks. That reorder is safe only because no McpError with an HTTP-range `.code` reaches the sync catch — every remote-McpError route is sealed by an inner catch, and in-repo McpErrors use negative `ErrorCode.*` values. The invariant is stated at the re-throw and pinned by a mapper test.

## Proof it works

Behavior-preservation refactor — there is no new user-visible behavior to screenshot; the claim is that the error-path behavior is byte-identical before and after, so the proof is invariance:

- The #1061 behavior-contract suite (landed just before this) pins the exact result shapes, telemetry property sets, and terminal task statuses per error class on both the sync and task paths. It passes here **with zero assertion edits** — the wire/telemetry contract is provably unchanged by the extraction, which is the actual claim of the refactor (not merely "tests pass").
- Cell-by-cell equivalence review — error class × path × aspect (status, diagnostics, response/userText, logging, store writes, cancel guards, wire wrapping, terminal status, abort source) — diffed the mapper and rewritten sinks against both pre-change catch blocks: behavior-identical for every reachable error. The McpError-hoist containment invariant was independently re-traced across every route into the sync catch (ACTOR_MCP branch, call-actor passthrough, `connectMCPClient`, `getMcpToolsInfoForActor`) and holds.
- New table-driven mapper unit test pins the exact response objects (including the full x402 payload in both `content[0].text` and `structuredContent`), abort handling, and both McpError cases.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,17 @@
+<!-- A PR that is hard to review is not ready for review. One concern per PR, clean diff (no debug logs, formatting noise, commented-out code). -->
+
+<!-- AI agents: fill in Why, What changed, and Proof it works. Leave "Notes for reviewers" empty — it belongs to the human author. No file lists, no restating the diff, no praising the change. -->
+
+<!-- Human author: an AI draft is a draft, not a description. Before requesting review: (1) self-review your own diff in the GitHub UI, (2) check every claim below against the diff and cut what you wouldn't say yourself, (3) write "Notes for reviewers" in your own words. -->
+
+## Why
+<!-- "Closes #123". No issue? Ad-hoc is the exception, not the default — one line on the problem and why the change is needed. Evidence beats adjectives: a measurement, a failing run, a probe result. -->
+
+## What changed
+<!-- For a small fix, "Before: … / Now: …" is enough. Don't list files — the diff shows that. Explain the approach only when it's non-obvious; otherwise it's noise. -->
+
+## Notes for reviewers (human-written)
+<!-- Your own words — two sentences beat none. Where should review look hardest? What are you unsure about? If AI wrote parts you don't fully follow, say which — the reviewer must not be the first human to read them. Any impact on apify-mcp-server-internal (internals.js exports, _meta, structuredContent, clientInfo, ?ui=/?payment=)? -->
+
+## Proof it works
+<!-- Not "unit tests pass" — CI shows that. An end-to-end MCP probe (mcpc transcript), a screenshot, or a link to an Actor run on the Apify platform. "Works on my machine" / "the AI said it works" is not proof. -->
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -37,7 +37,7 @@ Zero tolerance for errors — fix before proceeding, don't defer.
 
 When the user says "test with mcpc", **use mcpc** — do not invent a substitute (no curl, no ad-hoc Node/Python scripts, no unit tests in place of an e2e probe). Use the **apify CLI** (`apify datasets`, `apify key-value-stores`, `apify actors`, …) for ground-truth data — never curl the Apify API.
 
-After `pnpm run build`, run `mcpc` (no args) to check sessions: if `@stdio` (default) / `@stdio-full` (non-default tools) is listed, `mcpc @stdio restart`; otherwise `mcpc --config .mcp.json stdio connect @stdio`. Use the `mcpc-tester` subagent for systematic spec/edge-case coverage; call mcpc directly for quick checks.
+After `pnpm run build`, run `mcpc` (no args) to check sessions: if `@stdio` (default) / `@stdio-full` (non-default tools) is listed, `mcpc @stdio restart`; otherwise `mcpc connect .mcp.json:stdio @stdio` (non-default tools: `mcpc connect .mcp.json:stdio-full @stdio-full`). Use the `mcpc-tester` subagent for systematic spec/edge-case coverage; call mcpc directly for quick checks.
 
 ## Testing
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,17 @@
 
 All notable changes to this project will be documented in this file.
 
-<!-- git-cliff-unreleased-start -->
-## 0.11.6 - **not yet released**
+## [0.11.6](https://github.com/apify/apify-mcp-server/releases/tag/v0.11.6) (2026-07-16)
+
+### 🚀 Features
+
+- Add report-problem tool for agents to report issues ([#1050](https://github.com/apify/apify-mcp-server/pull/1050)) ([76586f4](https://github.com/apify/apify-mcp-server/commit/76586f4dd9f3c3c5cdbbde02c38e4d0ad562a360)) by [@RobertCrupa](https://github.com/RobertCrupa), closes [#748](https://github.com/apify/apify-mcp-server/issues/748)
+
+### 🐛 Bug Fixes
+
+- Resolve pre-flight task failures synchronously ([#1080](https://github.com/apify/apify-mcp-server/pull/1080)) ([64905c7](https://github.com/apify/apify-mcp-server/commit/64905c7627a476b04d3a8ce96d30d3c046c1e7b5)) by [@jirispilka](https://github.com/jirispilka), closes [#684](https://github.com/apify/apify-mcp-server/issues/684)
 
 
-<!-- git-cliff-unreleased-end -->
 ## [0.11.5](https://github.com/apify/apify-mcp-server/releases/tag/v0.11.5) (2026-07-08)
 
 
--- a/README.md
+++ b/README.md
@@ -243,6 +243,7 @@ Here is an overview list of all the tools provided by the Apify MCP Server.
 Legend for the **Enabled by default** column:
 - ✅ — in the default tool set.
 - ⚡ — auto-injected when `call-actor`, `add-actor`, an Actor tool, or `get-actor-run` is present (which is true in the default configuration).
+- ✅¹ — served by default, but only when telemetry is enabled and the client is not withheld: Anthropic surfaces (Claude.ai / Claude Desktop / Claude Code) or `local-agent-mode-apify`. To disable, pass an explicit `tools=` list that omits it.
 
 | Tool name | Category | Description | Enabled by default |
 | :--- | :--- | :--- | :---: |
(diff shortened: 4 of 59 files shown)


vojtechj-apify (approved):
## Approving — sound, behavior-preserving refactor

Verified: `type-check` clean, the contract + mapper test files pass (32 tests, zero assertion edits), and observable behavior is byte-identical to `master` in every branch **except** the deliberate one — hoisting the `McpError` re-throw above the payment/approval checks in the sync catch. That reorder is currently **sealed**: all in-repo `McpError`s use negative `ErrorCode.*`, and every remote-`McpError` route into the sync catch is caught by an inner handler. No live correctness bugs found.

None of the below blocks merge — they're fragility, test-coverage, cleanup, docs, and one convention.

### Findings (ranked)

**1. `src/mcp/server.ts:1430` — the hoist's safety is real but unenforced and under-tested.**
Behavior-preserving only while no `McpError` with a positive `.code` in `[100,600)` reaches the sync catch (`getHttpStatusCode` falls through to `.code`, so `McpError(402)` would otherwise satisfy the x402 predicate). Holds today, but the safety net is thin:
- The mapper classifies `McpError(402)` as `PAYMENT` (its own test pins this). Consider intercepting `error instanceof McpError` *inside* the mapper so "an `McpError` is never payment/approval" lives in one enforced place.
- No handler-level test guards the guarantee; the mapper test pins the *hazard* (402→payment), not the invariant (no HTTP-coded `McpError` reaches the sync catch). If an inner catch later stops sealing a route, unit tests stay green while the wire contract flips from x402 result to a thrown JSON-RPC error.
- The re-throw comment claims "every remote-McpError route" but omits `getMcpToolsInfoForActor` → `client.listTools()` (via `fetch-actor-details`). Sealed, so no bug — but the comment misleads a future auditor.

**2. `tests/unit/tool_call_error_mapper.test.ts` / contracts `FAILURE_CLASSES` — execution `failure_http_status` inclusion is untested.**
The only EXECUTION test error is `new Error('boom')` (no HTTP status), so the mapper's `...(httpStatus !== undefined ? { failure_http_status } : {})` arm is never exercised with a status present. A regression dropping `failure_http_status` for a genuine 4xx/5xx execution error (e.g. a 500) would keep every test green while blanking a telemetry field the dashboards consume. Add an execution case carrying an HTTP status.

**3. `src/mcp/server.ts:1854` — task catch builds the response before the cancel guard.**
`buildToolCallErrorResult` runs at the top of the task catch, so the payment/approval response (incl. `JSON.stringify`) is built *before* the `skipIfTaskCancelled` guards (1866/1891); the old code built it *after*. On a cancel-vs-error race the result is built then discarded. Builders are pure, so non-observable — but a step-order deviation from the byte-identical claim. Build after the guard, or split classification from response construction.

**4. `src/mcp/server.ts:1347` — the ACTOR_MCP inner catch is a divergent third copy of the execution classification.**
It hand-rolls `getToolStatusFromError` + `classifyFailureCategory` + `.slice(0,200)` + `buildActorFields` — byte-for-byte the mapper's EXECUTION arm — but omits `failure_http_status`, which the mapper includes. Same execution failure → different telemetry depending on path. The PR consolidated the two *outer* catches but left this drifted copy. (Can't call the mapper wholesale — that would reclassify remote `McpError(402)` as payment — but the diagnostics construction could be a shared builder.)

**5. `src/mcp/tool_call_error_mapper.ts:21` — `TOOL_CALL_ERROR_KIND` has no same-named type alias.**
`CONTRIBUTING.md` §Enumerations: *"Create a TypeScript type with the same name as the enumeration object."* Every comparable const-object (`TOOL_TYPE`, `SERVER_MODE`, `TOOL_STATUS`, `FAILURE_CATEGORY`) has the alias; here the test re-derives the type inline. Add `export type TOOL_CALL_ERROR_KIND = ...`.

**6. `src/mcp/tool_call_error_mapper.ts:2` — a "pure classifier" imports a response builder from the tools layer.**
It imports `buildPermissionApprovalResponse` from `../tools/actors/call_actor.js`, which imports back into `../../mcp/client.js` — an mcp→tools→mcp chain that drags call_actor's heavy graph into a module `AGENTS.md` calls a leaf classifier. Its twin `buildPaymentRequiredResponse` already lives in a leaf util (`utils/payment_errors.ts`); co-locating the approval builder there keeps the classifier lightweight. (No import *cycle* — verified — just layering weight.)

**7. `tests/unit/tool_call_error_mapper.test.ts:14` — `X402_PAYMENT_DATA` duplicated across two test files.**
The fabricators were centralized into `helpers/mcp_server.ts`, but the x402 payload literal (and derived expected response) are copy-pasted identically in `tool_call_error_mapper.test.ts:14` and `mcp.server.tool_call_contracts.test.ts:34`. Hoist it next to `makePaymentRequiredError`, mirroring `PERMISSION_HTTP_STATUS`.

**8. `src/mcp/AGENTS.md` — "the catch blocks own … wire wrapping" is inaccurate.**
For payment/approval the mapper returns the complete ready-to-send `response` (content + `isError` + `structuredContent`); the catch passes it straight through. Wire wrapping is catch-owned only for the execution kind. Worth also noting in the mapper's JSDoc that an HTTP-range-coded error (incl. `McpError(402)`) classifies as PAYMENT.

**9. `tests/unit/helpers/mcp_server.ts:68` — `makePaymentRequiredError()` no-arg contract silently inverted.**
Old no-arg attached the x402 payload; the new optional-arg version with no arg produces the *bare* error (the old `makeBarePaymentRequiredError`'s job). All current callers were updated, but a future `makePaymentRequiredError()` expecting a payload will silently exercise the fallback branch.

**10. `src/mcp/server.ts:1446 / 1862` — kind-dispatch duplicated across both sinks, no exhaustiveness guard.**
Neither catch uses `switch`/`assertNever`; EXECUTION is reached by fall-through, in two files. A future 4th union kind carrying `userText` would silently fall into the execution sink in both catches rather than trip a compile error. The task PAYMENT/APPROVAL branches are near-identical (only two strings differ) and collapsible.

**11. `tests/unit/utils.payment_errors.test.ts` — still inlines a 402 fabricator (low).**
Could import the now-shared `makePaymentRequiredError()`. Keeping the low-level predicate test self-contained is defensible; low priority.

Minor, folded in: the `PAYMENT | APPROVAL` arms share a shape and could be a two-arm union; `httpStatus` is computed eagerly at the mapper top but read only by the EXECUTION arm (not a regression — the old inline code did the same).

---

jirispilka:
Triage of the findings — thanks for the detailed pass. (Finding numbers backticked so they don't autolink.)

**Addressed** (latest commits): `#2` execution `failure_http_status` test · `#5` `TOOL_CALL_ERROR_KIND` type alias · `#6` moved the approval builder to `utils/payment_errors.ts` (mapper no longer imports the tools layer; its suite moved too) · `#7` shared `X402_PAYMENT_DATA` fixture · `#8` AGENTS.md + JSDoc wording · `#9` `makePaymentRequiredError` no-arg contract · `#1` hoist comment reworded (no longer implies an exhaustive route list).

**Deferred → #1104**: `#1` handler-level containment test · `#4` ACTOR_MCP third-copy telemetry (pre-existing) · `#10` dispatch exhaustiveness. All real, but separate from a mechanical extraction.

**Left as-is**: `#3` non-observable (builders are pure) · `#11` self-containment fits a low-level predicate test.

One note on `#5`: the premise that all comparable const-objects have a same-named alias isn't quite right — `TOOL_STATUS`/`FAILURE_CATEGORY` use PascalCase aliases — but added the same-named one here regardless.

---

MQ37 (approved):
Approved. Clean extraction, behavior preserved (tests + tsc pass, x402/Skyfire flows unaffected — verified independently).

One follow-up worth filing: `server.ts`'s `ACTOR_MCP` catch (~line 1341) still duplicates the classification arithmetic (`getToolStatusFromError` + `classifyFailureCategory` + `buildActorFields`) that `buildToolCallErrorResult` now owns. Not this PR's scope, but we should unify it.

jirispilka:
> Approved. Clean extraction, behavior preserved (tests + tsc pass, x402/Skyfire flows unaffected — verified independently).
> 
> One follow-up worth filing: `server.ts`'s `ACTOR_MCP` catch (~line 1341) still duplicates the classification arithmetic (`getToolStatusFromError` + `classifyFailureCategory` + `buildActorFields`) that `buildToolCallErrorResult` now owns. Not this PR's scope, but we should unify it.

Thanks. 
This was already partly covered in the follow up issue: https://github.com/apify/apify-mcp-server/issues/1104

PR #1082, opened 2026-07-11
refactor: Extract dispatchToolCall with exhaustive TOOL_TYPE switch

### LINKED ISSUE #1063: refactor: Extract dispatchToolCall with exhaustive TOOL_TYPE switch

jirispilka:
## Why

Closes #1063. Step 4 (final) of #658.

The `tool.type` dispatch ladder (INTERNAL / ACTOR_MCP / ACTOR) was written twice — in the sync `CallToolRequest` handler and in the task path — and had drifted: the task copy had no ACTOR_MCP branch, so an actor-mcp tool run as a task silently stored an empty `{}` "completed" result instead of the tool's real output.

## What changed

Both call sites now go through one private `dispatchToolCall`: a single exhaustive `switch (tool.type)` with a `satisfies never` default, so a future 4th tool type is a compile error rather than a silent fall-through. Callers are parameterized by capability, not mode — they construct the progress tracker, set notification forwarding, and pass the abort source. A `completeTask` helper folds the store→notify→track triple that was repeated at all four task outcomes.

One intended behavior change, unreachable by convention (no proxied actor-mcp tool sets `execution.taskSupport`, and the pre-dispatch gate rejects task calls without it; reachable only via `upsertTools`):

- **Before:** an ACTOR_MCP tool run as a task stored `{}` as `completed`.
- **Now:** it dispatches for real — success stores the proxied result; a connect/exec failure stores `completed` + `isError` (soft-fail, matching the sync path).

Contract note for `apify-mcp-server-internal`: reachable-path result shapes, `_meta`, `structuredContent`, notification timing, and telemetry are unchanged; only the above (unreachable-by-convention) ACTOR_MCP-as-task shape differs.

This is the last dedup work, after that we can continue with the next issue and simplify divide the `server.ts`
Will follow with this issue: https://github.com/apify/apify-mcp-server/issues/1084

## Proof it works

**mcpc live probe** (real `APIFY_TOKEN`, 10/10 checks):
- INTERNAL sync dispatch — `search-actors`, `fetch-actor-details`, `get-actor-run` returned real results.
- ACTOR sync dispatch — `call-actor apify/hello-world` returned a real `runId`.
- **Task lifecycle** (`executeToolAndUpdateTask` → `dispatchToolCall` → `completeTask`): `--detach` → `tasks-get` (`working` → `completed`) → `tasks-result` returned the full stored output, **not an empty `{}`**. Task `_meta` carries `taskId` + related-task / ActorRun fields; sync has none.
- Error paths — tool-not-found and invalid input both return clean MCP errors, no crash.

**Recommended before merge:** a human `test:integration` run (needs `APIFY_TOKEN`).


src/mcp/server.ts
```diff
@@ -1242,182 +1242,35 @@ export class ActorsMcpServer {
...
+                // for ACTOR tools. ACTOR_MCP forwards notifications directly, not via a tracker.
+                const progressTrackerOptIn =
+                    tool.type === TOOL_TYPE.ACTOR ||
+                    (tool.type === TOOL_TYPE.INTERNAL &&
+                        (tool.name === HELPER_TOOLS.ACTOR_CALL || tool.name === HELPER_TOOLS.ACTOR_RUNS_GET));
+                const progressTracker = progressTrackerOptIn
+                    ? createProgressTracker(progressToken, extra.sendNotification)
+                    : null;
+
+                const dispatchResult = await this.dispatchToolCall({
```
jirispilka, line 1274: I know that this function is awful with plenty of arguments, but I don't see a better solution right now.

I would merge it and deal with it later on.

MQ37 (approved):
Approved. Dispatch consolidation looks correct — verified sync/task-path behavior preservation and the one intentional behavior change, tests/tsc/lint clean.

Follow-ups worth addressing (not blocking):

- `shouldForwardNotifications`/`taskMode` are exact inverses at both call sites — one bit encoded as two params with no type-level guarantee they stay paired. Derive one from the other.
- `completeTask`'s 4 positional params (incl. two adjacent status-flavored args, `storeStatus`/`status`) should be an object param per our >3-param convention — transposition risk as-is.
- `ACTOR_MCP` case in `dispatchToolCall` nests 5-6 levels (switch→case→try→if→for→callback). Worth extracting into its own method (guard clause for the connect-failure branch, notification-wiring loop pulled into its own helper) so the dispatch switch stays flat and the nesting isn't buried inside a 700-line function.

jirispilka:
> Approved. Dispatch consolidation looks correct — verified sync/task-path behavior preservation and the one intentional behavior change, tests/tsc/lint clean.
> 
> Follow-ups worth addressing (not blocking):
> 
> * `shouldForwardNotifications`/`taskMode` are exact inverses at both call sites — one bit encoded as two params with no type-level guarantee they stay paired. Derive one from the other.
> * `completeTask`'s 4 positional params (incl. two adjacent status-flavored args, `storeStatus`/`status`) should be an object param per our >3-param convention — transposition risk as-is.
> * `ACTOR_MCP` case in `dispatchToolCall` nests 5-6 levels (switch→case→try→if→for→callback). Worth extracting into its own method (guard clause for the connect-failure branch, notification-wiring loop pulled into its own helper) so the dispatch switch stays flat and the nesting isn't buried inside a 700-line function.

Thanks, follow up filled: https://github.com/apify/apify-mcp-server/issues/1116

jirispilka:
@vojtechj-apify will you have time to review it?

PR #1083, opened 2026-07-11
fix(mcp): strip server-internal toolTelemetry at the wire boundary

### LINKED ISSUE #1052: fix: Uncaught tool errors leak toolTelemetry to the wire
## Problem

An uncaught tool error reaches `classifyToolCallError` (`src/mcp/tool_call_engine.ts:400`), which attaches the server-internal `toolTelemetry` field to the result (`:456`). Nothing strips it — the SDK's `CallToolResultSchema` is a `z.looseObject`, and the stateless era's `projectCallToolResult` only appends a text fallback — so the field reaches the client.

Reproduced on `master`:

```
$ mcpc @stdio-full tools-call abort-actor-run '{"runId":"bogusRunId12345"}' --json
{
  "content": [{ "type": "text", "text": "Error calling tool \"abort-actor-run\": Actor run was not found. Verify the tool name and input parameters." }],
  "isError": true,
  "toolTelemetry": { "toolStatus": "SOFT_FAIL" }
}
```

Two effects on an uncaught error: a server-internal field goes onto the wire, and the user gets the generic `getToolCallErrorUserText` message instead of a tool-specific one (compare `get-actor-run` with a bogus id: `Run with ID 'x' not found.`).

## Scope

- **Both protocol eras** — `classifyToolCallError` is called from the outer catch in `src/mcp/legacy_server.ts:606` and `src/mcp/stateless_server.ts:279`.
- Hits any tool whose throw escapes to that catch. Confirmed: `abort-actor-run`. Structural candidate: `search-apify-docs` (no catch around the Algolia call). Storage tools are **not** affected — they route 404s through `catchNotFound` + `respondUserError` (`src/tools/storage/storage_helpers.ts:16`), which strips.
- `tests/unit/mcp.server.error_handling.test.ts:69-76` asserts the leaked field, so it is currently pinned by tests.
- Removing the field changes client-facing wire output. `apify-mcp-server-internal` reads `toolTelemetry` nowhere in its `src`, so nothing on the hosted side depends on it.
- **Pre-existing** — verified byte-identical before/after the `respond*` refactor (#937), so not a regression from PR #1045.

Related: #935 (tool-response contract), #929 (analysis — P4: `isError` without enforced classification).

vjsai:
Closes #1052

## Problem

INTERNAL tools with no `try/catch` of their own let a thrown error propagate to the framework's outer `catch` in the `CallToolRequestSchema` handler. That path returns via `captureResult` and never goes through `extractToolTelemetry()` — the only place `toolTelemetry` is stripped. Because the MCP SDK's `CallToolResultSchema` is a `z.looseObject` (passthrough), the server-internal `toolTelemetry: { toolStatus }` key is forwarded verbatim **to the client**.

Confirmed end-to-end, not just by reading the code: driving a real MCP `Client` over an in-memory transport against a throwing INTERNAL tool, the client-visible result carries `toolTelemetry` on `master`.

## Fix — option 1 from the issue (single framework boundary)

`captureResult` is the one point every tool-call response leaves the handler through, so the strip goes there. That closes the leak on **every** return path — including future ones — rather than only the paths that happen to read the field.

- **`src/utils/tool_status.ts`** — new `stripToolTelemetry()`; `extractToolTelemetry()` now delegates its `delete` to it, so there is a single definition of "remove the internal field".
- **`src/mcp/server.ts`** — `captureResult()` strips `toolTelemetry` before the response is captured and returned.
- **`src/mcp/server.ts`** — the outer catch no longer attaches `toolTelemetry` at all. The ABORTED-aware `toolStatus` it carried is already reported to telemetry from the handler's local state, so **no classification accuracy is lost**. (The comment there previously explained the field was kept for byte-identity during #937; that rationale is what this PR retires.)

~10 lines of production code.

I scoped this to the **leak** only — the issue's second effect (generic `getToolCallErrorUserText` instead of tool-specific messages, i.e. option 2's per-tool `try/catch`) is a separate, larger change across several tools and is better as its own PR. Happy to follow up if you want it.

## Wire-shape change (please read)

This **changes client-facing wire output**: the leaked `toolTelemetry` key is gone from error responses. As you flagged in the issue, `apify-mcp-server-internal` consumes the stripped wire shape — it should need no change (it now simply always sees the shape it already expected), but it's worth a matching assertion in its contract suite. I can't touch that repo, so flagging it here.

## Tests

- **New wire-level test** (`tests/unit/mcp.server.error_handling.test.ts`): real `Client` + `InMemoryTransport` linked pair against a throwing INTERNAL tool, asserting the client-visible result has no `toolTelemetry`. **Fails on `master`, passes here** — this is the before/after wire snapshot the issue asked for.
- The two existing outer-catch tests pinned the *leaked* shape (`expect(result.toolTelemetry).toEqual({ toolStatus })`). They now assert the field is **absent** from the response while the ABORTED / FAILED status still reaches telemetry (spying on `logToolCallAndTelemetry`) — so the strip is proven not to cost classification.
- **New unit tests** for `stripToolTelemetry()` (`tests/unit/utils.tool_status.test.ts`): strips in place, leaves clean responses untouched, passes through non-objects.

Local gates, all green:

```
pnpm run format:check   All matched files use the correct format. (223 files)
pnpm run lint           Found 0 warnings and 0 errors. (190 files)
pnpm run type-check     tsc -p tsconfig.json --noEmit — clean
pnpm run build:core     tsc -b src — clean
pnpm run test:unit      Test Files 70 passed (70)
                        Tests 963 passed | 2 skipped (965)
```

CHANGELOG is git-cliff-generated from the conventional commit, so it is not hand-edited.

---

Prepared with AI assistance (Claude Code); every line reviewed and the behaviour verified locally against the repo's own harness. Opened as a **draft** — happy to adjust the approach (e.g. option 2 instead) before you spend review time on it.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -100,6 +100,7 @@ import {
     extractAjvErrorDetails,
     extractToolTelemetry,
     getToolStatusFromError,
+    stripToolTelemetry,
 } from '../utils/tool_status.js';
 import {
     buildActorFields,
@@ -836,9 +837,15 @@ export class ActorsMcpServer {
             let resolvedToolName = name;
             // Captured by `captureResult` so the `finally` block can measure response size for telemetry.
             let toolResult: unknown = null;
+            // Single wire boundary: every tool-call response leaves the handler through here, so this
+            // is where the server-internal `toolTelemetry` field is guaranteed to be stripped. The
+            // tool-dispatch paths read it via `extractToolTelemetry()` (which also strips it), but the
+            // framework paths that bypass that helper — notably the outer `catch` — would otherwise
+            // pass it straight to the client, since `CallToolResultSchema` is a passthrough schema.
             const captureResult = <T>(r: T): T => {
-                toolResult = r;
-                return r;
+                const result = stripToolTelemetry(r);
+                toolResult = result;
+                return result;
             };
             const failInvalidParams = async (
                 message: string,
@@ -1311,14 +1318,13 @@ export class ActorsMcpServer {
                     validationMissingProperty: callDiagnostics.validation_missing_property,
                     validationAdditionalProperty: callDiagnostics.validation_additional_property,
                 });
-                // This framework outer-catch path bypasses extractToolTelemetry (returned via
-                // captureResult), so preserve the pre-existing wire shape { toolStatus } exactly:
-                // reuse the local ABORTED-aware toolStatus, do NOT re-derive from the error (which
-                // would drop ABORTED and leak failureCategory/failureHttpStatus onto the wire).
+                // The ABORTED-aware `toolStatus` computed above is already what the `finally` block
+                // reports to telemetry, so it does not need to ride along on the response. Attaching
+                // it here as `toolTelemetry` is what leaked it to the client (#1052): this path
+                // bypasses `extractToolTelemetry()`, the only other place the field is stripped.
                 return captureResult({
                     ...respondOk(getToolCallErrorUserText(name, error)),
                     isError: true,
-                    toolTelemetry: { toolStatus },
                 });
             } finally {
                 if (shouldTrackTelemetry) {
--- a/src/utils/tool_status.ts
+++ b/src/utils/tool_status.ts
@@ -148,6 +148,23 @@ export function extractAjvErrorDetails(errors: ErrorObject[] | null | undefined)
 }
 
 /**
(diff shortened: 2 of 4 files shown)


jirispilka:
@vjsai I'm sorry for the delayed response, it looks good, can you please fix the conflicts?

jirispilka:
I'm closing this PR as it is stale

PR #1094, opened 2026-07-16
refactor: Stream resources/read under one inline cap, align with spec

jirispilka:
## Why

Review follow-up to #1003: the buffered proxy held up to 5 MB in RAM per read, the wrong-origin refusal was dead code, errors lacked the spec's `data: { uri }`, and `user@api.apify.com` URLs silently read unauthenticated.

## What changed

The read is now a stream capped at 256 KB — axios aborts the download mid-flight. One limit, one link-out site, body returned verbatim by Content-Type (the JSON re-serialization logic is deleted). `readApiResource` owns the single origin gate and rejects userinfo URLs. Errors throw `McpError` with `data: { uri }` and are logged.

Raw axios instead of `httpClient.call()` on purpose: with a stream body, `call()` turns non-2xx into junk messages and strands a socket per retry. One attempt, as before.

Also: `buildBinaryRecordDisposition` rename, `resolveApifyClient` as a private method, exactly-at-limit boundary test, AGENTS.md trimmed with the `https://`-scheme deviation documented. Templates follow-up: #1095 (stacked as #1096).

## Tools vs resources

Same data as the storage tools, on purpose: tools are the guided, model-driven surface — and the only one in payment mode or for clients without resources support. Resources are the raw data plane. Shared size/MIME helpers prevent drift; no tool is removed. Follow-up: tools return resource links (#998, #587).

I've refactored the code with couple of cosmetic changes, private method vs const, shortened AGENTS.md significanlty.

The biggest change is the removal of 5MB. We need only single threshold. 
I'm not sure about the actually value, no strong preference.

The important things is templates follow up in #1095 , please read rationale there

## Proof it works

Live mcpc against the real API:

```
$ mcpc --json @stdio resources-read ".../v2/store?limit=1000"   # ~1 MB body
→ "Response body exceeds the 262144-byte inline limit. Download it from … re-read with a smaller limit/offset range."
```

Also probed live: 404 keeps the API's message (`HTTP 404: Dataset was not found`, `-32602`); non-Apify and `evil.com@api.apify.com` URLs get the origin refusal; pretty-printed JSON comes back byte-exact; `?limit=1&offset=0` vs `offset=1` return successive items.

https://claude.ai/code/session_01Wv2ptD8dSxeeVK9QnDK1k2

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -11,13 +11,10 @@ export const ACTOR_MAX_MEMORY_MBYTES = 4_096; // If the Actor requires 8GB of me
 /**
  * Content larger than this is linked out instead of inlined, since inlining it would blow up the context
  * window (base64 inflates a binary payload ~33%, and a large text/JSON body overflows it just as easily).
- * The key-value-store-record tool and the API-resource proxy both cap binaries here (link to a fetchable
- * URL); the proxy additionally links oversized text/JSON to a download URL.
+ * The key-value-store-record tool caps binaries here (link to a fetchable URL); the API-resource proxy
+ * caps every body here — its download is also aborted mid-flight at this limit via axios `maxContentLength`.
  */
 export const MAX_INLINE_BYTES = 256 * 1024;
-
-/** Hard cap on bytes the API-resource proxy downloads; enforced mid-flight by axios `maxContentLength`. */
-export const MAX_DOWNLOAD_BYTES = 5 * 1024 * 1024;
 
 /**
  * Advisory threshold (uncompressed bytes) above which dataset tools append a size hint steering the
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -642,21 +642,22 @@ export class ActorsMcpServer {
         return meta?.apifyToken || this.options.token;
     }
 
+    /**
+     * Token-scoped client for resources/read (the API proxy needs auth). Deliberately token-only:
+     * unlike the CallTool path it does NOT forward provider/payment headers, so a payment-only
+     * session (x402/Skyfire, no Apify token) has no client and every read fails by design.
+     */
+    private resolveApifyClient(params: ApifyRequestParams): ApifyClient | undefined {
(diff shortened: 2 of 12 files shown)


RobertCrupa (approved):
This is great, thank you for the alignment.

When I discussed this with @MQ37, we agreed to use Mixpanel data to set a new limit once we understand the sizes we are working with. A limit that would fit the majority of requests while being reasonable.

MQ37 (changes requested):
`readApiResource()` hard-decodes every textual body as UTF-8, ignoring the declared Content-Type charset. Old `arraybuffer` path went through apify-client's charset-aware `body_parser.js`; the new stream path bypasses it.

Reproduced against the live API: PUT a KV record with real ISO-8859-1 bytes (`Content-Type: text/plain; charset=iso-8859-1`), confirmed byte-identical via raw curl, then read it through this PR's `readApiResource()`:

```
Ground truth: Café à côté de la crème brûlée
Proxy returns (mimeType: charset=iso-8859-1): Caf� � c�t� de la cr�me br�l�e
```

Silent, irreversible corruption. Real-world reachable — any Actor storing scraped non-UTF-8 text.

Fix: parse charset from Content-Type, decode with it if `Buffer.isEncoding()`, else fall through to the blob branch instead of lying about the encoding.

jirispilka:
@MQ37 Fixed in 8b95f27, exactly as you suggested: charset parsed from Content-Type, decoded via `body.toString(charset)` when `Buffer.isEncoding()` knows it, otherwise fall-through to the lossless base64 blob branch — same rule as apify-client's `body_parser`. Tests written first (both reproduced the corruption before the fix): latin1 decodes correctly as text; your exact `charset=iso-8859-1` case now returns a byte-intact blob (`iso-8859-1` is not a Buffer encoding label — the old arraybuffer path kept it as a Buffer too, so this restores parity rather than adding decoding).

---

MQ37 (approved):

PR #1096, opened 2026-07-16
feat: Advertise URI templates in resources/templates/list

### LINKED ISSUE #1095: feat: Advertise URI templates for API resources in templates/list
## Problem

`resources/read` is a generic proxy over Apify API GET URLs (#998, #1003), but individual resources carry no metadata: `resources/templates/list` returns `[]` and discovery lives only in the server-instructions prose. A client that does not inject `initialize.instructions` gives the model no way to learn which URL shapes exist or which take paging parameters — e.g. that `datasets/{id}/items` pages with `limit`/`offset` while a key-value-store record cannot be shrunk. The only in-band teacher today is the 256 KB too-large notice (#1094).

## Proposed solution

Advertise a small set of RFC 6570 URI templates with descriptions in `listResourceTemplates` (`src/resources/resource_service.ts`). Templates are advertisement only — the read path stays a generic proxy, no per-type handlers. Candidate set:

- `https://api.apify.com/v2/datasets/{datasetId}/items{?limit,offset,clean,fields,format}` — pages with `limit`/`offset`; inline cap 256 KB
- `https://api.apify.com/v2/key-value-stores/{storeId}/records/{recordKey}` — returned whole in its stored Content-Type; not pageable, oversized returns a download URL
- `https://api.apify.com/v2/key-value-stores/{storeId}/keys{?limit,exclusiveStartKey}`
- `https://api.apify.com/v2/actor-runs/{runId}` and `https://api.apify.com/v2/logs/{runId}`

Descriptions must not imply the list is exhaustive — any Apify API GET URL stays readable. Base the URLs on `getApifyAPIBaseUrl()`, not a hardcoded origin.

Note: this is the first place the server hands clients `https://` resource URIs, where the MCP spec says clients may fetch directly (token-gated → 401). Descriptions should say reads go through `resources/read` (the server injects the token). See the deviation note in `src/resources/AGENTS.md` (#1094).

## Plan

- [ ] Template definitions + descriptions in `resource_service.ts`
- [ ] Unit tests (`resources.service.test.ts`) + `templates/list` integration assertion in `suite.ts`
- [ ] Update server instructions and the `templates/list` facts in `src/resources/AGENTS.md`

Related: #998 (parent), #1003 (feature PR), #1094 (streaming rewrite), #587 (roadmap: resource links).

jirispilka:
## Why

Closes #1095. Stacked on #1094. `resources/templates/list` returned `[]`, so a client that doesn't surface server instructions can't tell any API resources exist — or which URLs page.

## What changed

`listResourceTemplates` now advertises five RFC 6570 templates (dataset items, KV record, KV keys, run, run log). Descriptions carry the paging semantics (`{?limit,offset,…}` vs `exclusiveStartKey` vs not pageable), the 256 KB cap, and that reads go through `resources/read` with the server injecting the token. Advertisement only — the read path stays a generic proxy; any other Apify API GET URL remains readable. URLs derive from `getApifyAPIBaseUrl()`.

## Tools vs resources

The templates advertise data the storage tools already serve, on purpose: tools are the guided, model-driven surface — and the only one in payment mode or for clients without resources support. Resources are the raw data plane. No tool is removed. Follow-up: tools return resource links (#998, #587).

None

## Proof it works

```
$ mcpc --json @stdio resources-templates-list
[ { "name": "dataset-items",
    "uriTemplate": "https://api.apify.com/v2/datasets/{datasetId}/items{?limit,offset,clean,fields,format}", … },
  … 5 templates ]
```

https://claude.ai/code/session_01Wv2ptD8dSxeeVK9QnDK1k2

### DIFF
--- a/src/resources/AGENTS.md
+++ b/src/resources/AGENTS.md
@@ -45,7 +45,9 @@ JSON-RPC error, never success-shaped content): 3xx/4xx except 429 → `InvalidPa
 status (network, mid-stream drop) → `InternalError`. 401/403 append a hint via `getHttpErrorHint()`
 (shared with `tools/call`); failures are logged via `logHttpError` (5xx → exception). Size
 link-outs are **successful** reads returning a download pointer, not failures. Discovery is the
-server-instructions prose; `resources/templates/list` returns nothing.
+server-instructions prose plus `resources/templates/list`, which advertises the common URL shapes
+(dataset items, KVS record/keys, run, log) as RFC 6570 templates with paging params — advertisement
+only, not a route table: the read path stays generic and accepts any Apify API GET URL.
 
 ## Gotcha
 
--- a/src/resources/resource_service.ts
+++ b/src/resources/resource_service.ts
@@ -13,6 +13,7 @@ import { ErrorCode, McpError } from '@modelcontextprotocol/sdk/types.js';
 import log from '@apify/log';
 
 import type { ApifyClient } from '../apify_client.js';
+import { getApifyAPIBaseUrl } from '../apify_client.js';
 import type { PaymentProvider } from '../payments/types.js';
 import { SERVER_MODE } from '../types.js';
 import { readApiResource } from './api_resources.js';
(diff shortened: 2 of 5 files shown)


RobertCrupa (approved):
Thank you for handling this!

MQ37 (approved):
Spec-checked `resources/templates/list` against the MCP schema (`2025-11-25`, matches the installed `@modelcontextprotocol/sdk@1.29.0`): all 5 templates have valid required fields (`name`, `uriTemplate`), correct RFC 6570 syntax, and every URI is a real, verified Apify API endpoint (including `key-value-store-keys` correctly using `exclusiveStartKey` instead of `offset` — a real API distinction, not copy-paste).

One non-blocking issue worth a follow-up: the `dataset-items` template sets `mimeType: 'application/json'` while also advertising `format` as a query parameter (`{?limit,offset,clean,fields,format}`). The real dataset-items endpoint's `format` accepts `json | jsonl | xml | html | csv | xlsx | rss` — 7 different response Content-Types, not just JSON. Per spec, `ResourceTemplate.mimeType` "should only be included if all resources matching this template have the same type," which isn't true here once `format` is in play.

Suggest one of:
- drop `mimeType` from this template (it's optional; the actual Content-Type is already returned correctly per-read via #1094's Content-Type-driven branching), or
- drop `format` from the advertised query params if you want to keep asserting a fixed `application/json` type.

Don't advertise both a format-varying param and a fixed mimeType on the same template — approving as-is since it's non-blocking, but flag for a quick fix.

jirispilka:
@MQ37 Fixed in bb1a394 — dropped `mimeType` from `dataset-items` (kept `format`; the per-read Content-Type comes from the proxy's branching), with a test pinning the absence.

---

PR #1097, opened 2026-07-16
refactor: Polish resources/read proxy from review

jirispilka:
## Why

Follow-up polish for the `resources/read` proxy in #1003, targeted at that branch (not `master`). The substantive review findings (dead-code origin gate, duplicated link-out, missing error logging, userinfo) were already fixed on #1003; these are the remaining low-risk cleanups. No user-visible behavior change except that the server instructions now honor an `APIFY_API_BASE_URL` override.

## What changed

- **Server instructions** — the `resources/read` example URLs now derive from `getApifyAPIBaseUrl()` instead of a hardcoded `https://api.apify.com`. Before: examples pointed at prod even under a staging/self-hosted `APIFY_API_BASE_URL`. Now: they match the origin gate and `resources/templates/list`.
- **`throwReadFailure()`** — the identical `McpError` construction in the request-catch and non-2xx paths is collapsed into one helper (byte-identical message/code/`data`, logging unchanged); `statusToErrorCode` → `getErrorCodeForStatus` for the verb convention.
- **Proxy note** — documented that reads go through the raw axios instance and so skip `ensureNodeInit()`, meaning a resource read does not honor `HTTPS_PROXY` (auth + MCP-origin headers still apply — not a token leak). Flagged for proxy-mandatory deployments; a real fix needs an apify-client API to run node-init without `call()`.
- **Tests** — added the `429 → InternalError` mapping and the exactly-`MAX_INLINE_BYTES` inline boundary (previously only `+1`/link-out was covered).

## Proof it works

- Oracle on the merged tree: `type-check`, `lint`, `check:agents`, `format` all clean; `test:unit` **1094 passed / 1 skipped** (includes the two new cases).
- `throwReadFailure` was verified behavior-preserving against both original throw sites (same code, message incl. `HTTP <status>:` prefix + hint, and `{ uri }` data); the underlying proxy was previously exercised end-to-end via `mcpc` against the live Apify API on #1003.

---
_Generated by [Claude Code](https://claude.ai/code/session_0127S1UPvFhgfD9DYpCNyPxK)_

### DIFF
--- a/src/resources/api_resources.ts
+++ b/src/resources/api_resources.ts
@@ -15,11 +15,7 @@ import { getHttpErrorHint } from '../utils/mcp.js';
 
 const TEXT_MIME_TYPE = 'text/plain';
 
-/**
- * Base MIME types emitted as `text` contents; everything else (including a missing
- * Content-Type) is emitted as a base64 `blob`. JSON and XML are text so the raw body
- * stays readable and byte-exact — the proxy never parses or re-serializes it.
- */
+/** Textual base MIME types (returned as `text`); everything else becomes a base64 `blob`. */
 function isTextualMimeType(baseMimeType: string | undefined): boolean {
     if (!baseMimeType) return false;
     return (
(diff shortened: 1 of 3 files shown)


MQ37 (approved):
Nice, I like that

RobertCrupa (approved):
Great stuff, thank you!

Issue #337, opened 2025-11-19
Revisit errors returned: Protocol Errors vs Tool Execution Errors


jirispilka:
https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling

- Protocol errors indicate the request is malformed or invalid; the client should not retry with the same request.
- Tool execution errors indicate the request was valid but execution failed; the client may retry or handle the error based on the content.

Other issues:

Error #1 - search-actors with invalid limit:

When limit exceeds 100, tool fails with generic "Tool execution failed"
Recommendation: Return error like: "Invalid limit parameter: 150. Maximum allowed limit is 100."

Error #3 - search-apify-docs with empty query:

Empty query fails with generic "Tool execution failed"
Recommendation: Return error like: "Query parameter is required and cannot be empty."

Error #4 - apify-slash-rag-web-browser missing required param:

Missing required query parameter fails with generic "Tool execution failed"
Recommendation: Return validation error like: "Missing required parameter 'query'. Please provide a search term or URL."

jirispilka:
Verified on current `master` with mcpc probes (fresh build). The classification now matches the spec: invalid arguments reject as JSON-RPC `-32602 InvalidParams` from the shared validation in `src/mcp/tool_call_engine.ts:215` (both protocol eras), while valid-but-failed execution returns an `isError: true` tool result with the real message.

All three cited cases return a specific message, not a generic failure:

| Call | Result |
|---|---|
| `search-actors {"limit":150}` | `-32602` — `/limit: must be <= 10` |
| `search-apify-docs {"query":""}` | `-32602` — `/query: must NOT have fewer than 1 characters` |
| `apify--rag-web-browser {}` | `-32602` — `root: must have required property 'query'` |
| `get-actor-run {"runId":"nonexistentRun123"}` | `isError: true` — `Run with ID 'nonexistentRun123' not found.` |

Note the `limit` ceiling is 10, not 100 — the API rejects higher values when input schemas are included (`MAX_LIMIT_WITH_INPUT_SCHEMA`).

One leftover nit split out into its own issue: the validation message points at `fetch-actor-details` even for non-Actor tools.

Closing.

PR #1102, opened 2026-07-17
fix: Propagate auth errors instead of reporting Actor not found

MQ37:
## What
`fetchActorDetails()` (used by both `fetch-actor-details` and its widget variant) collapsed every error — 401, 403, 5xx, network — to `null`, identical to a genuine 404. A stale/invalid `APIFY_TOKEN` surfaced as "Actor ... was not found", nudging the caller toward `search-actors` instead of fixing the credential. Now only 404/400 return `null`; everything else propagates.

## Why
Mirrors the pattern `getActorDefinition()` (`actor_definition.ts`) already uses in this same codebase. Once the error propagates, the existing 401/403-aware catch-all (`getToolCallErrorUserText`) produces the correct auth message — no new user-facing text needed, no downstream changes required.

## Testing
- Added failing-first unit tests: `utils.actor_details.test.ts` (404 → null, 401 → throws with statusCode preserved), plus regression tests in `tools.fetch_actor_details.response.test.ts` / `.widget.response.test.ts` asserting the tool handlers don't swallow a rejected `fetchActorDetails`.
- `pnpm run type-check`, `lint`, `test:unit` (1058 passed), `format`, `check:agents` all clean.
- Live-verified via `mcpc` against a built `dist/stdio.js`, real Actor `apify/instagram-scraper`:
  - invalid token → `Authentication failed — check APIFY_TOKEN is set and valid.`
  - valid token, real Actor → normal Actor card.
  - valid token, nonexistent Actor → unchanged not-found message.

### DIFF
--- a/src/utils/actor_details.ts
+++ b/src/utils/actor_details.ts
@@ -108,12 +108,14 @@ export async function fetchActorDetails(
             readmeSummary: actorInfo.readmeSummary,
         };
     } catch (error) {
-        logHttpError(error, `Failed to fetch actor details for '${actorName}'`, { actorName });
         // 404/400 is a genuine "Actor doesn't exist" — same treatment as the not-found check
         // above. Anything else (401/403/5xx) must propagate: it's a different failure class
         // (e.g. invalid token) the caller should surface as such, not as "not found".
         const statusCode = getHttpStatusCode(error);
-        if (statusCode === 404 || statusCode === 400) return null;
+        if (statusCode === 404 || statusCode === 400) {
+            logHttpError(error, `Failed to fetch actor details for '${actorName}'`, { actorName });
+            return null;
+        }
         throw error;
     }
 }
--- a/src/utils/actor_details.ts
+++ b/src/utils/actor_details.ts
@@ -8,7 +8,7 @@ import type { Actor, ActorCardOptions, ActorInputSchema, ActorStoreList, Structu
 import { getActorMcpUrlCached } from './actor.js';
 import { formatActorForWidget, formatActorToActorCard, formatActorToStructuredCard } from './actor_card.js';
 import { searchActorsByKeywords } from './actor_search.js';
-import { logHttpError } from './logging.js';
+import { getHttpStatusCode, logHttpError } from './logging.js';
 import type { PricingTier } from './pricing_info.js';
 
 const ACTOR_DETAILS_PICTURE_SEARCH_LIMIT = 5;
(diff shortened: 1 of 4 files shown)


RobertCrupa (approved):

src/utils/actor_details.ts
```diff
@@ -109,7 +109,12 @@ export async function fetchActorDetails(
         };
     } catch (error) {
         logHttpError(error, `Failed to fetch actor details for '${actorName}'`, { actorName });
```
jirispilka:
```suggestion
```

src/utils/actor_details.ts
```diff
@@ -109,7 +109,12 @@ export async function fetchActorDetails(
...
-        return null;
+        // 404/400 is a genuine "Actor doesn't exist" — same treatment as the not-found check
+        // above. Anything else (401/403/5xx) must propagate: it's a different failure class
+        // (e.g. invalid token) the caller should surface as such, not as "not found".
+        const statusCode = getHttpStatusCode(error);
+        if (statusCode === 404 || statusCode === 400) return null;
```
jirispilka:
```suggestion
   if (statusCode === 404 || statusCode === 400) {
        logHttpError(error, `Failed to fetch actor details for '${actorName}'`, { actorName });
        return null;
    }
```

logHttpError fires unconditionally, then non-404/400 errors are re-thrown. The throw reaches the outer catch in [src/mcp/server.ts:1475](https://github.com/apify/apify-mcp-server/blob/master/src/mcp/server.ts#L1475) (logHttpError(error, 'Error occurred while calling tool', …)) — so every 401/403/5xx logs twice

Moving the `logHttpError` below fixes that

jirispilka (approved):
Looks good, there was a small issue with double logging.

I've created two follow up issues:

[#1114 — refactor: Unify getActorDefinition not-found check onto getHttpStatusCode](https://github.com/apify/apify-mcp-server/issues/1114) (labels: t-ai, debt)
[#1115 — feat: Dedicated user messages for 429 and 5xx in getToolCallErrorUserText](https://github.com/apify/apify-mcp-server/issues/1115) (label: t-ai; cross-references [#337](https://github.com/apify/apify-mcp-server/issues/337))

src/utils/actor_details.ts
MQ37: Fixed in 31345cd0 — moved `logHttpError` inside the 404/400 branch so it only logs there; propagated errors are now logged exactly once (by the outer catch in server.ts).

src/utils/actor_details.ts
MQ37: Removed — logging moved into the 404/400 branch below (see reply on the other thread) in 31345cd0.

PR #1103, opened 2026-07-17
refactor: Move dispatchToolCall out of server.ts to tool_dispatch.ts

jirispilka:
## Why

Closes #1107 (part of #1084). After the #658 stack, `src/mcp/server.ts` is 2121 lines; the `dispatchToolCall` switch (~245 lines) is unrelated to server lifecycle, registration, or the handshake, and makes the class hard to navigate. Stacked on #1082 (`refactor/dispatch-tool-call`).

## What changed

Before: `dispatchToolCall` was a private method on `ActorsMcpServer`. Now: a plain exported function in `src/mcp/tool_dispatch.ts` (same pattern as `tool_call_error_mapper.ts`), taking `apifyMcpServer: ActorsMcpServer` as a param; both call sites (sync handler, `executeToolAndUpdateTask`) call the import directly. The body moved verbatim — the only edits are the `this` → `apifyMcpServer` rewrites. `server.ts`: 2121 → 1863 lines.

Task machinery (`executeToolAndUpdateTask`, `emitTaskStatusNotification`) deliberately stays in `server.ts` — the follow-up is tracked in #1108 and #1109 (sub-issues of #1084), delivered by the stacked PRs #1105 and #1106. No `ActorsMcpServer` public-API change; nothing consumed by `apify-mcp-server-internal` moves.

## Notes for reviewers (human-written)

Mechanical move

## Proof it works

- mcpc live probe over stdio after `pnpm run build`: handshake + `tools/list` (11 default tools) OK; INTERNAL tools (`search-actors`, `fetch-actor-details`, `get-dataset-items`) end-to-end; ACTOR branch via real `apify/hello-world` run on the sync path and again via `--detach` task mode — identical result shapes; unknown tool / invalid params still return well-formed `-32602`. ACTOR_MCP branch verified by byte-level diff of the moved body (sandbox blocked `*.apify.actor` outbound, failure occurs pre-dispatch at tool load).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01JLCP4WtbFnaZYT4bDWg5Fs

---
_Generated by [Claude Code](https://claude.ai/code/session_01JLCP4WtbFnaZYT4bDWg5Fs)_

Issue #1084, opened 2026-07-11
refactor: Move dispatchToolCall and task machinery out of server.ts


jirispilka:
Follow-up to the #658 stack (#1073 → #1080 → #1081 → #1082).

## Problem

`src/mcp/server.ts` is ~1965 lines after the stack lands. The dedup deleted the duplicated ladders and catch blocks, but `dispatchToolCall` (~240 lines) landed inside the same file per #658's design (private method on `ActorsMcpServer`), so the file did not shrink. The TODO above `prepareToolCallContext` ("we should split this huge method into smaller parts as it is slowly getting out of hand") predates the stack and still stands, and the replacement TODO above `executeToolAndUpdateTask` documents the remaining outer-orchestration duplication.

## Proposal

Mechanical decomposition, no behavior change:

- Move `dispatchToolCall` to its own module (e.g. `src/mcp/tool_dispatch.ts`) — its only `this` uses are `apifyMcpServer: this` and `this.server`, so a function taking the server works. `tool_call_error_mapper.ts` (#1081) is the pattern.
- Consider moving the task machinery (`executeToolAndUpdateTask`, `completeTask`, cancel watcher wiring) with it or into a sibling module.
- The contract suite from #1061 (exact telemetry key sets, exact stored payloads, x402 shapes) is the guard; it must pass with zero assertion edits.

Deliberately deferred out of #1082 to avoid churning a reviewed stacked PR; the private-method placement was #658's recorded design decision, revisited here as a standalone move.

PR #1105, opened 2026-07-18
refactor: Extract tool-call telemetry helpers from server.ts

### LINKED ISSUE #1108: refactor: Extract tool-call telemetry helpers from server.ts
Part 2 of #1084 — prerequisite for the task-machinery move. `prepareTelemetryData` and `logToolCallAndTelemetry` move to `src/mcp/tool_call_telemetry.ts` as standalone functions; `withReportProblemNudge` moves next to the `appendReportProblemNudge` it wraps in `src/tools/dev/report_problem.ts`. Each takes a single per-function params object (CONTRIBUTING.md >3-params rule). `telemetryEnabled`/`telemetryEnv` become `public readonly` (same pattern as `server`/`tools`/`taskStore`/`options`) so extracted code reads them off the `apifyMcpServer` it receives; `setupTelemetry()` computes-and-returns, constructor assigns. Bodies otherwise verbatim; #1061 contract suite unchanged.

Delivered by #1105.

jirispilka:
## Why

Closes #1108 (part of #1084). The task-machinery move (#1106, stacked on this PR) can't leave `server.ts` while the task path depends on three class-private helpers shared with the sync handler. This PR extracts them first, per the repo's refactor-lands-first rule. Stacked on #1103.

## What changed

Before: `prepareTelemetryData`, `logToolCallAndTelemetry`, and `withReportProblemNudge` were private methods on `ActorsMcpServer`; `telemetryEnabled`/`telemetryEnv` were private fields mutated inside `setupTelemetry()`.

Now: the two telemetry functions are standalone in `src/mcp/tool_call_telemetry.ts`

@MQ37 I made a call to make  telemetryEnabled and telemetryEnv public so that I don't have to pass it all over the place. I'm doubting that decision now but I would keep it that way for now and reconsider later.

https://claude.ai/code/session_018MfTtEvQ9MYhydr8qSUopw

### DIFF
--- a/src/mcp/AGENTS.md
+++ b/src/mcp/AGENTS.md
@@ -29,6 +29,10 @@ implementations, not by importing from here.
 - `tool_dispatch.ts` — `dispatchToolCall()`: the single exhaustive `switch (tool.type)`
   (INTERNAL / ACTOR_MCP / ACTOR) both the sync handler and the task path run. Plain
   function taking the `ActorsMcpServer` instance; touches no class state beyond `.server`.
+- `tool_call_telemetry.ts` — `prepareTelemetryData()` / `logToolCallAndTelemetry()`: shared by
+  the sync `CallToolRequestSchema` handler and the task path. Plain functions taking the
+  `ActorsMcpServer` instance (as `apifyMcpServer`), reading `telemetryEnabled`/`telemetryEnv`
+  and `options.*` off it — both are `public readonly` on `ActorsMcpServer`.
 - `const.ts` — the invariant constants below (the single source for these values).
 
 ## Gotchas & invariants
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -60,11 +60,11 @@ import { createResourceService } from '../resources/resource_service.js';
 import type { AvailableWidget } from '../resources/widgets.js';
 import { resolveAvailableWidgets } from '../resources/widgets.js';
 import { getServerInfo } from '../server_card.js';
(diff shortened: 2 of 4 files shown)


MQ37 (approved):
Approved. Mechanical move verified — all 3 functions byte-identical vs their old private-method versions, tests/tsc/lint/build all clean.

On the public `telemetryEnabled`/`telemetryEnv` fields: I don't love it, but let's keep it as-is and move on rather than block on it.

One thing I would like addressed: `prepareTelemetryData`/`logToolCallAndTelemetry`'s param objects are inlined directly in the function signatures — extract them into named types (e.g. `PrepareTelemetryDataParams`, `LogToolCallAndTelemetryParams`) for readability. No need to `export` them unless something outside this file actually needs to reference the type.

jirispilka:
Done in afba5ae — `PrepareTelemetryDataParams` and `LogToolCallAndTelemetryParams`, non-exported, signature-only (type-check/lint/test/format/check:agents green, #1061 suite untouched). Force-pushed; the stacked #1106 was rebased onto the new head.

Worth noting the sibling extractions inline the same way — `dispatchToolCall` (`tool_dispatch.ts`) and `executeToolAndUpdateTask` (`task_execution.ts`, #1106). Kept those as-is here to hold this PR to its scope; folded the same named-`Params` treatment into #1111 so the whole `src/mcp` family lands consistent with the CONTRIBUTING `Params` convention in one sweep.

On the public `telemetryEnabled`/`telemetryEnv` — noted, same mild reservation on our side; leaving as-is per your call, revisitable if the #1110 orchestration dedup changes how that state is reached.

---

RobertCrupa (approved):

PR #1122, opened 2026-07-21
feat(evals): Code Mode A/B workflow benchmark

MQ37:
## What
Adds Code Mode A/B benchmark support to the workflow evals harness: per-turn traces, richer timeout/repeat controls, and an eval-only hardcode that forces `apify/code-runtime` (`Ewe0VxlktEE7SChSI`) to always return its full README instead of the auto-generated summary.

## Why
Comparing "Code Mode" vs direct-tool-call arms needs multi-run evidence, per-turn traces, and bounded per-test timeouts to be trustworthy. The README hardcode replaces an earlier self-declared `fullReadmeOnly` actor.json flag — any Actor could set that flag to dodge summarization, so the bypass is now pinned to the one Actor ID whose docs are an exact API contract.

## Testing
Existing + new unit tests updated (`utils.actor_details`, `evals.*`) — `pnpm test`.


MQ37:
Splitting into two stacked PRs: #1123 (hardcoded code-runtime full-README fix, needed as-is) and #1124 (evals harness, optional/chunky — can stay closed if not wanted).

PR #1123, opened 2026-07-21
fix: hardcode apify/code-runtime Actor ID for full-README bypass

MQ37:
## What
`resolveReadmeContent()` now always returns the full README for `apify/code-runtime` (matched by Actor name, `CODE_RUNTIME_ACTOR_NAME` in `const.ts`), bypassing the auto-generated summary.

## Why
`apify/code-runtime`'s README documents an exact API contract (method names/shapes) that the auto-generated summary omits — confirmed via a live eval trace where the agent fetched the README, got the summary, and still guessed the wrong dataset API. Matched by name rather than a self-declared `actor.json` flag, so no other Actor can opt itself out of summarization.

Matched by Actor name (`apify/code-runtime`) rather than Actor ID: IDs are deployment-specific (differ between local/staging/production), so a hardcoded ID would silently stop working outside production.

## Testing
Added unit tests for `resolveReadmeContent()` (regular Actor vs `apify/code-runtime`, with/without summary, blank summary). `pnpm run type-check`, `lint`, `test:unit` all pass.

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -7,6 +7,11 @@ export const ACTOR_MAX_DESCRIPTION_LENGTH = 500;
 // Actor run const
 export const ACTOR_MAX_MEMORY_MBYTES = 4_096; // If the Actor requires 8GB of memory, free users can't run actors-mcp-server and requested Actor
 
+// apify/code-runtime's README documents an exact API contract (method names/shapes) that the
+// auto-generated summary can omit — always return its full README. Hardcoded by Actor ID
+// (not an actor.json flag) so no other Actor can opt itself out of the summary.
+export const CODE_RUNTIME_ACTOR_ID = 'Ewe0VxlktEE7SChSI';
+
 // Tool output
 /**
  * Content larger than this is linked out instead of inlined, since inlining it would blow up the context
--- a/src/const.ts
+++ b/src/const.ts
@@ -8,9 +8,10 @@ export const ACTOR_MAX_DESCRIPTION_LENGTH = 500;
 export const ACTOR_MAX_MEMORY_MBYTES = 4_096; // If the Actor requires 8GB of memory, free users can't run actors-mcp-server and requested Actor
 
 // apify/code-runtime's README documents an exact API contract (method names/shapes) that the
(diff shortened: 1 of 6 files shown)


RobertCrupa (approved):
Feels a bit hacky, but if this is the only exception to the rule then it's probably the best way to go about it.

Lets 🚢😎

src/const.ts
```diff
@@ -7,6 +7,11 @@ export const ACTOR_MAX_DESCRIPTION_LENGTH = 500;
 // Actor run const
 export const ACTOR_MAX_MEMORY_MBYTES = 4_096; // If the Actor requires 8GB of memory, free users can't run actors-mcp-server and requested Actor
 
+// apify/code-runtime's README documents an exact API contract (method names/shapes) that the
+// auto-generated summary can omit — always return its full README. Hardcoded by Actor ID
+// (not an actor.json flag) so no other Actor can opt itself out of the summary.
+export const CODE_RUNTIME_ACTOR_ID = 'Ewe0VxlktEE7SChSI';
```
jirispilka: Can we have Actor name, this will break at staging

jirispilka (approved):
We should use actor name instead of hard-coding id, this will break at different envs, like local, staging

For the future, if we consider adding code-runtime as a default tool, we'll need to add integration tests

Pre-approving, thanks!

src/const.ts
MQ37: we can hardcode the ID also for staging, I am strongly against using Actor name - if we (JC) rename this Actor then it will break, using just the name is fragile.

jirispilka: The name is the contract. Everywhere else, rag-web-browser, testing Actors we are using names

If we are serious about code-mode, we will need to add it into integration tests, maybe also into the internal repo's integration tests (not sure) and then we'll need to have multiple IDs.

The concern is environment portability, not preference for names. IDs identify a deployment-specific Actor record; apify/code-runtime identifies the product we intend to ship. A rename should update this contract and test.

MQ37: let's go with the name

MQ37:
Updated: match `apify/code-runtime` by Actor name instead of Actor ID, so this works across environments (staging etc.) too.

PR #1125, opened 2026-07-21
refactor: Harden tool-call error-mapper

### LINKED ISSUE #1104: Harden the tool-call error-mapper seam (follow-up to #1081)
Follow-up to the review on #1081 (which extracted `buildToolCallErrorResult`). Three review findings there are real but were out of scope for that mechanical refactor — grouping them here.

## 1. ACTOR_MCP inner catch duplicates the classification arithmetic that `buildToolCallErrorResult` now owns

The ACTOR_MCP branch's inner catch in `server.ts` (~line 1341) still hand-rolls `getToolStatusFromError` + `classifyFailureCategory` + `.slice(0, 200)` + `buildActorFields` — the same arithmetic as the mapper's EXECUTION arm — a third copy that drifted from the two the extraction unified. The visible symptom: it omits `failure_http_status` (which the mapper includes), so the same execution failure produces different telemetry depending on the path, and dashboards keyed on `failure_http_status` under-report for ACTOR_MCP failures.

Unify it. `buildToolCallErrorResult` can't be called wholesale here — a remote `McpError` whose `.code` is 402 would be reclassified as payment — so the fix is to extract the shared execution-diagnostics construction into a small builder both the mapper's EXECUTION arm and this catch call, rather than the mapper's full classification. (Bare minimum if the extraction slips: add the missing `failure_http_status` field so the telemetry matches.)

## 2. Kind-dispatch has no exhaustiveness guard

Both tool-call catches dispatch on `errorResult.kind` with sequential `if … return`, reaching EXECUTION by fall-through, in two places. A future 4th `ToolCallErrorResult` kind that also carries `userText` would silently fall into the execution sink rather than trip a compile error. Add a `switch` + `assertNever` (precedent already in `utils/tools.ts`). The task PAYMENT/APPROVAL branches are also near-identical (only two strings differ) and collapsible. This is a structural refactor of the catches, separate from the mapper extraction.

## 3. No handler-level regression test for the McpError-hoist containment invariant

The sync-catch hoist (re-throwing `McpError` before classification) is safe only because no HTTP-range-coded `McpError` reaches the sync catch — every remote-McpError route is sealed by its own inner catch. Today that rests on code-reading plus the mapper's `McpError(402) → payment` documentation test; nothing drives a real `McpError` through the sealed routes at the handler level. If a future route stops sealing, unit tests stay green while the wire contract flips from an x402 result to a thrown JSON-RPC error. Add a handler-level test that pushes a remote `McpError` through the ACTOR_MCP / call-actor passthrough / `listTools` paths and asserts it never reaches the outer catch.

Part of #658; related to #1063.

jirispilka:
## Why

Closes #1104. Part of #658. Three follow-up findings from #1081: ACTOR_MCP failures omitted `failure_http_status` (telemetry drifted by path), the kind-dispatch reached EXECUTION by fall-through with no compile-time guard, and the McpError-containment invariant had no test.

## What changed

- Extracted the execution-diagnostics construction into one builder (`buildExecutionDiagnostics`, `utils/tool_status.ts`); the mapper's EXECUTION arm and the ACTOR_MCP catch both delegate, so ACTOR_MCP now reports `failure_http_status`. Not `buildToolCallErrorResult` wholesale — a 402 `McpError` must stay soft-fail, not reclassify as payment.
- Both catches dispatch via `switch (errorResult.kind)` with a `satisfies never` default; a 4th kind now fails type-check.
- Added a handler-level test driving a real rejecting `McpError` (incl. 402) through the sealed ACTOR_MCP route.

Intended side effect: an ACTOR_MCP **task-mode** 402 now suppresses the report-problem nudge (per the global 402-is-billing rule), where it previously appended it. Pinned by a test.

**internal impact:** `failure_http_status` now appears on ACTOR_MCP soft-fail telemetry — contract `apify-mcp-server-internal` consumes; may need a matching test.

## Out of scope (conscious, not gaps)

- Catch **sinks** not unified — only the diagnostics builder is shared; owned by #658.
- ACTOR_MCP `isError` TODO, the `server.ts` 1k-line ceiling, and the task-vs-sync nudge asymmetry all predate this change and are left alone.

## Proof it works

type-check / lint / `test:unit` (1120 passed, 1 skipped) / check:agents green. The containment tests (incl. 402) and the task-mode nudge test were each shown red→green.

### DIFF
--- a/tests/unit/mcp.server.tool_call_contracts.test.ts
+++ b/tests/unit/mcp.server.tool_call_contracts.test.ts
@@ -1,5 +1,5 @@
 import type { Client } from '@modelcontextprotocol/sdk/client/index.js';
-import { ErrorCode } from '@modelcontextprotocol/sdk/types.js';
+import { ErrorCode, McpError } from '@modelcontextprotocol/sdk/types.js';
 import { afterEach, describe, expect, it, vi } from 'vitest';
 
 import log from '@apify/log';
@@ -10,6 +10,11 @@ import type { ActorsMcpServer } from '../../src/mcp/server.js';
 import type { PaymentProvider } from '../../src/payments/types.js';
 import * as telemetry from '../../src/telemetry.js';
 import * as callActor from '../../src/tools/actors/call_actor.js';
+import {
+    REPORT_PROBLEM_INVALID_INPUT_NUDGE,
+    REPORT_PROBLEM_NUDGE,
+    reportProblem,
+} from '../../src/tools/dev/report_problem.js';
 import type { ToolEntry, ToolInputSchema } from '../../src/types.js';
 import { TOOL_TYPE } from '../../src/types.js';
 import { compileSchema } from '../../src/utils/ajv.js';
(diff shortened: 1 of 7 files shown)


tests/unit/mcp.server.tool_call_contracts.test.ts
```diff
@@ -413,6 +418,88 @@ describe('executeToolAndUpdateTask()', () => {
         expect(trackSpy.mock.calls).toHaveLength(1);
         expect(trackSpy.mock.calls[0][2].tool_status).toBe(TOOL_STATUS.SUCCEEDED);
     });
+
+    it('suppresses the report-problem nudge for a 402-carrying ACTOR_MCP task failure', async () => {
```
RobertCrupa, line 422: I thought this case was already handled, wasn't it?

Did you run into it while developing, or were you adding it to be sure?

RobertCrupa (approved):

tests/unit/mcp.server.tool_call_contracts.test.ts
jirispilka: It was there, but it never fired on the ACTOR_MCP path so an ACTOR_MCP task 402 classified as INVALID_INPUT and did get the nudge. I know that ACTOR_MCP are not allowed for agentic payments at all but this unifies the behavior. So I kept it

Good question! Thanks!

MQ37 (approved):
Approving — refactor is sound, verified type-check/lint/tests green on the PR head. Two follow-up requests before merge:

1. **`buildExecutionDiagnostics` (`utils/tool_status.ts`)**: 4 positional params (`error, isAborted, actorName, actorId`) — please switch to an object param per our TypeScript conventions (object params past 3, especially with `actorName`/`actorId` adjacent and same-typed — easy to transpose by mistake at a call site).

2. **Log-field mapping from `callDiagnostics`**: the `failureCategory/failureHttpStatus/actorName/validationKeyword/...` spread passed into `logHttpError`/`log.softFail`/`log.error` is hand-rolled at 5+ call sites now (`server.ts:897,1286`, `task_execution.ts:259,339,350,383`, `tool_dispatch.ts:207`). Since this PR is already extracting shared diagnostics construction, worth extracting a small `callDiagnostics → log fields` mapper (or context object) too, so the two don't drift back out of sync the way `buildExecutionDiagnostics` fixed for `failure_http_status`.

Both are mechanical, not blocking if you'd rather do them as fast-follow — your call.

jirispilka:
Thanks @MQ37 🙏

1. **Object param** — done in `070c847`. `buildExecutionDiagnostics` now takes a single `ExecutionDiagnosticsParams` object (matches the `ToolCallErrorParams` precedent), which drops the transpose risk on the adjacent `actorName`/`actorId`.

2. **`callDiagnostics → log fields` mapper** — agreed it's worth doing, but that spread is pre-existing across `server.ts`/`task_execution.ts`/`tool_dispatch.ts` and predates this PR, so I'd keep it out of this one per our "one concern per PR / refactor separately" rule. Handling it as a fast-follow rather than widening this diff.

---

PR #1137, opened 2026-07-23
fix: Parallelize integration test suite

MQ37:
## What
Fixes `notifications/progress` delivery (a real MCP-spec-compliance bug: `_meta` was placed at the notification's top level instead of inside `params`, so any strict client silently dropped the message) and uses it to give integration tests a deterministic way to learn a run's `runId`. Uses that to remove the last blocker to running `tests/integration/suite.ts` concurrently instead of sequentially.

## Why
The integration suite forced `concurrent: false` for two reasons: every test shared one `client` variable/`afterEach(close)` (races under concurrency), and 3 abort/cancel tests discovered the Apify runId via a run-list-polling heuristic that isn't safe if another test touches the same Actor at the same time. Fixing `notifications/progress` (see first commit) replaces that heuristic with a direct signal from the server, removing the need for isolation. Also deleted 1 test whose "top of a `desc: true, limit: N` listing" assertion only held under the old sequential ordering — real concurrent Actor runs from other tests can push their target off page 1; unit coverage for the tool already exists separately (its sibling test had the same issue but was independently fixed on `master` with a page-position-agnostic assertion instead, so it's kept).

Every test now creates and closes its own client inside a shared `itc(name, clientOptions, testFn)` helper — `createClientFn` runs, `testFn` runs, and `client.close()` runs in a `finally`, so a failed assertion can't leak a client (child process on stdio, server session on streamable-http). One place owns the cleanup contract instead of it being repeated per test.

Also merged `master` in: it had drifted significantly since this branch was opened (stateless MCP server mode, add-actor removal, new `McpSuiteClient`/`asLegacyClient` client typing for the 2026-07-28 protocol dimension). Re-applied this PR's changes on top rather than restarting — only 2 files had real conflicts (`src/mcp/task_execution.ts`, `tests/integration/suite.ts`); everything else auto-merged cleanly.

## Testing
`pnpm run type-check`, `lint`, `format`, `check:agents`, `test:unit` (1247 passed) all clean. CI's `Integration tests / Integration tests` job (now covering 3 transport dimensions — stdio, streamable-http, and the 2026-07-28 stateless protocol — against the real Apify API):
- before (sequential, master): ~5m55s
- after (concurrent, this PR): 1m55s

Independently re-verified via a fresh subagent (re-ran the full suite against the live API itself, not just the reported numbers), and again after the master merge + `itc()` refactor via a second independent verification pass (diffed test titles across master/this-branch/merged-result to confirm no test was dropped or added by accident — the only removed titles trace to the intentional deletion above or to master's own unrelated add-actor removal).


vojtechj-apify:
**Clients leak on test failure now that cleanup moved into the test body.**

The old `afterEach(() => client?.close())` ran even when a test failed. The new pattern puts `await client.close()` as the last statement of each test body, so any earlier `expect` failure or thrown error skips it:

```ts
client = await createClientFn();
const prompts = await client.listPrompts();
expect(prompts.prompts.length).toBe(0);  // if this throws…
await client.close();                     // …never runs → leaked client
```

A leaked client is a child process (stdio) or a server session (streamable-http). Several tests carry `{ retry: 1..3 }`, and with `concurrent: true` a flaky test now leaks a client per attempt, in parallel — exactly the resource pressure concurrency adds, which can cascade into more flakiness.

Suggest guaranteeing cleanup with `try { … } finally { await client?.close() }`, or a `test.extend` fixture with automatic teardown (also removes the repeated `let client` / `await client.close()` boilerplate). I know `onTestFinished` was tried and raced on a shared module-level variable — a `try/finally` or fixture sidesteps that since neither relies on vitest's current-test tracking.

---

tests/integration/utils/task_waits.ts
```diff
@@ -58,38 +55,30 @@ export async function assertStatusMessagePropagated(taskClient: Client, stream:
...
-            const runs = await apiClient.actor(actorId).runs().list({ limit: 3, desc: true });
-            return runs.items.find(
-                (r) =>
-                    r.startedAt instanceof Date && r.startedAt >= startedAfter && !TERMINAL_RUN_STATUSES.has(r.status),
-            )?.id;
-        },
-        { timeout: RUN_DISCOVERY_TIMEOUT_MS, interval: RUN_DISCOVERY_INTERVAL_MS },
-    );
-    return runId as string;
+export function captureRunIdFromProgress(): {
```
RobertCrupa, line 68: (Nit) The previous function had a timeout (watiUntil) do we not want the same here?

RobertCrupa (approved):
does this change the contract suite on the internal server?

tests/integration/utils/task_waits.ts
MQ37: Previously we polled the Apify API for the new run which could take time for the run to appear and it was flaky. Currently we are waiting for notification that is fired almost immediately after the run, so I don't think the timeout is needed here. But we can add that to be extra safe so the test case does not hang.

MQ37: pushed a new commit with this change

MQ37:
@RobertCrupa the internal server test suite is completely isolated and this does not affect it.

jirispilka (approved):
Nice, thanks!

I added one more test that was missing

Issue #1140, opened 2026-07-23
feat: Add MCP 2026-07-28 stateless server


jirispilka:
Part of #1128. Depends on #1139, #1145, #1146, and #1147. The shared tool-call engine from #1139 is already merged.

## Context and motivation

Add a 2026-07-28 stateless registration surface without changing the existing sessionful 2025-era server. PR #1143 is a spike and reference only; this issue defines its replacement after the behavior-preserving refactors land.

### In scope

- Add the v2 SDK dependencies alongside v1: `@modelcontextprotocol/server`, pinned exact while in beta (`2.0.0-beta.5` as of 2026-07-24). The v2 package name differs from v1's `@modelcontextprotocol/sdk`, so both install side by side.
- Export `createStatelessServer(actorsMcpServer)` from the package root.
- Register `tools/list`, `tools/call`, `resources/*`, and `prompts/*` as a thin protocol adapter over `ActorsMcpServer` shared logic.
- Normalize v2 per-request client context at the adapter boundary.
- Read the Apify token from `ctx.http?.authInfo?.token`, with the existing configured-token fallback where applicable.
- `createMcpHandler` performs no token verification and never derives `authInfo` from request headers — it is caller-supplied. The dev server therefore needs a bearer-passthrough shim that lifts the `Authorization` header into `ctx.http.authInfo`; hosted auth stays server-derived (see internal repo impact).
- Build a fresh request-scoped tool snapshot and resolve mode and `report-problem` visibility per request.
- Map neutral service and tool-call outcomes to v2 `ProtocolError` and `projectCallToolResult`.
- Wire `src/dev_server.ts` with `isLegacyRequest` and `createMcpHandler`; legacy requests continue through the current sessionful route.
- Add development-server routing tests for both protocol eras.

### Out of scope

- Tasks on 2026-07-28.
- `subscriptions/listen`.
- Rewriting the legacy HTTP stack.
- Adding stateless-only behavior to the private legacy adapter.
- Deleting `add-actor`; that remains #1131.

## Technical design

Construct one v2 `Server` per request, as required by `createMcpHandler`. The adapter converts the v2 `ServerContext` to the neutral client and call contexts, asks the facade for a request-scoped tool snapshot, and delegates to shared tool, prompt, resource, and telemetry logic. It owns only v2 registration and error/result projection.

Use `stateless` or the dated revision in our names and docs. The SDK literal `modern` may appear only where its external API requires it.

[...]

Issue #1132, opened 2026-07-22
test: Add MCP 2026-07-28 integration and conformance coverage


MQ37:
Part of #1128. Depends on #1140. Must merge and pass before the hosted stateless route reaches production.

## Scope

- Extend `tests/integration/suite.ts` with the dated transport value `2026-07-28`.
- Add a v2 SDK `Client` using automatic version negotiation.
- Run protocol-neutral tool, prompt, resource, structured-output, payment, and error scenarios against the new transport.
- Keep session lifecycle, Tasks, restore, standalone SSE, and termination scenarios restricted to legacy transports using the existing `it.runIf` and `it.skipIf` patterns.
- Pin the official `@modelcontextprotocol/conformance` package to an exact compatible version as a development dependency.
- Add a repeatable package command that runs the conformance server suite against the local development server.

## Implementation decisions

- **Packages, pinned exact:** `@modelcontextprotocol/client@2.0.0-beta.5` (must match the pinned server beta) and `@modelcontextprotocol/conformance@0.2.0-alpha.9` (the `latest` 0.1.16 does not know `2026-07-28`; newer alphas are blocked by pnpm `minimumReleaseAge` until they age past 3 days).
- **Client typing: union, no adapter.** `createClientFn` returns `Client | StatelessClient` (v1 `@modelcontextprotocol/sdk` client union v2 `@modelcontextprotocol/client` client). The suite only calls `callTool`, `listTools`, `listPrompts`, `readResource`, and `close`, all callable through the union. The `client.experimental.tasks` call sites (legacy-only Tasks cases) need an explicit narrow to the v1 client — add one guard helper and use it in those describe blocks. If result-type unions cause heavy assertion friction, the fallback is a v1-shaped facade over the v2 client in `tests/helpers.ts`; both hide behind the same `createClientFn` signature.
- **Follow-up (#1169):** migrate the whole suite to the v2 client (it negotiates both protocol eras via `mode: 'auto'`) once v2 leaves beta, removing the union and the v1 client from the suite.
- **v2 client auth:** verify the v2 `StreamableHTTPClientTransport` can send `Authorization: Bearer`; if awkward, use the `?token=` query parameter the server already accepts.

## Conformance command

The package script must execute the equivalent of:

`pnpm exec conformance server --url <dev-server-url> --suite all --spec-version 2026-07-28`

[...]

Issue #1128, opened 2026-07-22
Migrate to MCP spec 2026-07-28 (stateless)


MQ37:
Add MCP 2026-07-28 stateless support alongside the existing sessionful server. Existing clients keep their current behavior except for the separately shipped `add-actor` retirement.

Non-goals: Tasks on 2026-07-28 and `subscriptions/listen`.

## Public repo sequence

- [x] #1129 — stop selecting `add-actor` for new connections
- [x] #1131 — delete the retired `add-actor` implementation
- [x] #1145 — add protocol-neutral MCP client context
- [x] #1146 — make prompt and resource services protocol-neutral
- [x] #1147 — separate legacy MCP wiring from `ActorsMcpServer`; depends on #1131, #1145, and #1146
- [x] #1140 — add the MCP 2026-07-28 stateless server; depends on #1147 and merged #1139
- [ ] #1132 — add the dated integration transport and official conformance coverage

#1145 and #1146 may run in parallel. #1131 remains a separate cleanup PR and must not be mixed into the refactors.

PR #1143 stays open as a spike until the replacement for #1140 exists. Do not merge it.

## Hosted server sequence

Internal umbrella: apify/apify-mcp-server-internal#672.

- Internal SDK dependency work may proceed independently.
- Internal route wiring (#676) waits for #1140 to be released.
- Public #1132 and internal #677 must pass before the stateless route reaches production.

Use `stateless`, `legacy`, or the dated protocol revision in our names and documentation. Do not use `modern` except where an external SDK API requires that literal.

Ecocitizenz:
Following closely — we’re deep in 2026-07-28 dual-stack work right now.

One pattern that has been working cleanly: keep the legacy session path behind a feature flag for 30–60 days while the new self-contained requests go live, then externalise remaining state into explicit handles.

Happy to share the concrete checklist we’ve been using across several production servers if it helps move this forward.

mcp.ecocitizenz.com

Ecocitizenz:
Great to see Apify migrating to the 2026-07-28 stateless spec.

A quick heads-up for teams planning to deploy this updated Apify server
into enterprise client environments: because the new spec is stateless,
passing static API keys to authenticate these scraping tools breaks SOC 2
non-repudiation. CISOs will block agents that have unrestricted
web-scraping capabilities without per-request verifiable identity.

If your agency is getting blocked by procurement when deploying Apify
agents, we run fixed-scope £395 MCP Readiness Audits that map out the exact
[...]

PR #1143, opened 2026-07-23
spike: MCP 2026-07-28 stateless server

jirispilka:
## Status

Spike only. Do not merge.

This branch proved that the v2 SDK can register the 2026-07-28 surface, route it through `src/dev_server.ts`, and reuse the shared tool-call engine. It also showed that adding stateless-only methods to the legacy `ActorsMcpServer` and duplicating prompt/error adaptation produces the wrong dependency direction.

The replacement work is tracked by:

- #1145 — protocol-neutral client context
- #1146 — protocol-neutral prompt and resource services
- #1147 — legacy adapter separation
- #1140 — replacement stateless server
- #1132 — integration and conformance coverage

Keep this PR open as implementation reference until the replacement for #1140 exists, then close it as superseded.

Part of #1128.


jirispilka:
This is no longer needed, closing

PR #1161, opened 2026-07-26
fix(actors): Prevent standby telemetry leaks

### LINKED ISSUE #1085: fix: Strip internal telemetry from standby rejection
## Bug

`checkPaymentProviderStandbyConflict` builds its rejection with `respondUserError`.

`respondUserError` attaches the internal `toolTelemetry` field. Normal tool execution strips that field before returning a result, but this standby check runs as a pre-flight short circuit and bypasses that stripping.

The internal field therefore reaches:

- legacy synchronous tool-call results
- 2026 stateless synchronous tool-call results
- stored legacy task results

## Root cause

`respondUserError` is intended for tool implementations whose results pass through telemetry extraction.

The standby rejection is a framework-level result. Its telemetry is already represented by `buildPreflightFailureOutcome` as:

- `toolStatus: SOFT_FAIL`
- `failure_category: INVALID_INPUT`

The response itself must not carry `toolTelemetry`.

Existing contract tests mock the standby rejection with an object that does not contain `toolTelemetry`, so they do not reproduce the production response.

## Fix

Write a failing regression test using the production standby-rejection shape.

Change `checkPaymentProviderStandbyConflict` to build the rejection with `respondErrorNoTelemetry` instead of `respondUserError`.

Preserve the existing content and `isError` fields. Only the internal `toolTelemetry` field should disappear.

Cover:

- the standby helper result
- a synchronous result through the shared engine
- the stored legacy task result

The stateless path uses the same synchronous engine, so its response should also be checked if the existing test setup can cover it without duplicating the shared-engine assertion.

## Non-goals

- No changes to standby detection.
- No changes to payment-required responses.
- No generic response-stripping layer.
- No telemetry classification changes.

shixi-li:
## What changed

- build the standby payment-provider rejection with the existing no-telemetry response helper
- test the real standby result through both synchronous and task-backed tool-call handlers
- assert that neither the wire response nor the stored task result contains `toolTelemetry`

## Why

`checkPaymentProviderStandbyConflict` returns before the normal telemetry extraction step. Using `respondUserError` on that framework path exposed server-internal outcome metadata to clients and persisted the same field in terminal task results.

## Risk

The production change only swaps the response constructor for the existing standby rejection. Error text, `isError`, telemetry accounting, and standby-before-payment priority remain unchanged. The regression covers both public response paths with the real helper and an in-memory task store.

## Validation

- `pnpm exec vitest run tests/unit/mcp.server.tool_call_contracts.test.ts`
- `pnpm run test:unit` (1177 passed, 1 skipped)
- `pnpm run type-check`
- `pnpm run lint`
- `pnpm run format:check`
- `pnpm run check:agents`
- `pnpm run build:core`

## Reviewer focus

Please verify that the standby pre-flight remains a user error while `toolTelemetry` stays server-internal on both sync and task result paths.

Closes https://github.com/apify/apify-mcp-server/issues/1085

### DIFF
--- a/src/tools/actors/call_actor.ts
+++ b/src/tools/actors/call_actor.ts
@@ -32,6 +32,7 @@ import { wrapJsonText } from '../../utils/encode_text.js';
 import { logHttpError } from '../../utils/logging.js';
 import {
     respondAborted,
+    respondErrorNoTelemetry,
     respondRaw,
     respondServerError,
     respondUserError,
@@ -291,7 +292,8 @@ export async function checkPaymentProviderStandbyConflict(params: {
         failureCategory: FAILURE_CATEGORY.INVALID_INPUT,
     });
 
-    return respondUserError(ActorLoadError.standbyPaymentNotSupported(normalizedActorName).message);
+    // This framework short-circuit bypasses telemetry extraction; see apify/apify-mcp-server#1085.
+    return respondErrorNoTelemetry(ActorLoadError.standbyPaymentNotSupported(normalizedActorName).message);
 }
 
 /**
--- a/tests/unit/mcp.server.tool_call_contracts.test.ts
+++ b/tests/unit/mcp.server.tool_call_contracts.test.ts
@@ -10,6 +10,7 @@ import * as mcpClient from '../../src/mcp/client.js';
 import type { McpClientContext } from '../../src/mcp/client_context.js';
 import type { ActorsMcpServer } from '../../src/mcp/server.js';
 import type { PaymentProvider } from '../../src/payments/types.js';
+import { actorDefinitionCache } from '../../src/state.js';
 import * as telemetry from '../../src/telemetry.js';
 import * as callActor from '../../src/tools/actors/call_actor.js';
 import {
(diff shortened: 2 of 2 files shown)


shixi-li:
Rebased this branch onto the latest `master` (`ebe6f5e`) and resolved the test-fixture conflict. The PR is mergeable again and the change remains limited to the standby response constructor plus its regression test.

Validation on the rebased head (`1a9df41`):

- `pnpm exec vitest run tests/unit/mcp.server.tool_call_contracts.test.ts` — 41 passed
- `pnpm run test:unit` — 1246 passed, 1 skipped
- `pnpm run type-check`
- `pnpm run lint` — 0 warnings, 0 errors
- `pnpm run format` and `pnpm run format:check`
- `pnpm run check:agents`
- `pnpm run build:core`

The three new workflow runs are still marked `action_required`:

- Apify pull request toolkit: https://github.com/apify/apify-mcp-server/actions/runs/30422006841
- CI (PR label): https://github.com/apify/apify-mcp-server/actions/runs/30422006927
- CI (PR): https://github.com/apify/apify-mcp-server/actions/runs/30422006986

Could a maintainer approve the workflow runs and review the PR when convenient?

jirispilka:
Thanks for the patch and the clear write-up. I'm closing this.

The diff is correct. respondErrorNoTelemetry is the right constructor, it matches what the payment arm of buildPreflightFailureOutcome already does, and your test is a genuine regression test: it fails if the production line is reverted.

The problem is that verifying this change end-to-end needs access you doesn't have:

The telemetry itself. toolTelemetry exists to feed Segment. The test pinning the standby event's properties still runs against a mocked standby result, so nothing here shows the real event still lands as SOFT_FAIL / INVALID_INPUT with no failure_http_status. That needs our event stream.

This is a case where filing the problem is the contribution and implementing it isn't. #1085 stays open; we'll take it from there. We're tightening our contribution guidelines on exactly this in #1192.

PR #1155, opened 2026-07-24
refactor!: Separate legacy MCP wiring from ActorsMcpServer

jirispilka:
## Why

Closes #1147. Part of #1128.

`ActorsMcpServer` combined two concerns: shared Apify behavior (tools, `actorStore`, server-mode, prompts/resources, widgets, payments, telemetry) and sessionful v1 SDK wiring (the SDK `Server`, request handlers, initialize, Tasks, logging, notifications, errors, transport lifecycle). PR #1143 showed that adding a second (stateless) serving unit forced the dependency direction backwards — shared tool execution reached up into the class (`apifyMcpServer.server`, `.tools`, `extra`, `McpError`). This extraction fixes the direction so #1140 can expose both protocol eras through `dev_server.ts` while reusing one Apify core.

**Breaking:** the raw v1 `.server` and `.taskStore` are no longer public on `ActorsMcpServer` — they move to the private adapter. Runtime-safe for `apify-mcp-server-internal` (it reads neither; verified against its `new ActorsMcpServer(...)` call sites, which use only `.options.initializeRequestData`, `.loadToolsByName()`, `.loadToolsFromUrl()`, `.listAllToolNames()`, `.connect()`). External TypeScript consumers touching those two members break at compile time — hence `refactor!`.

## What changed

All v1 SDK wiring moved verbatim into a new package-private `LegacyMcpServer` (`src/mcp/legacy_server.ts`). `ActorsMcpServer` keeps the shared Apify behavior, constructs exactly one adapter, and delegates `connect()`/`close()` to it via a narrow `LegacyMcpServerHost` interface — the adapter never imports the concrete facade, and shared modules never import the adapter. Shared tool execution (`tool_call_engine`, `tool_dispatch`, `tool_call_telemetry`, `task_execution`) now takes plain values; `InternalToolArgs` drops `extra`/`apifyMcpServer`/`mcpServer`. No v2 SDK dependency and no stateless routing were added (that is #1140).

Behavior is preserved byte-for-byte: tool names, results, error codes/messages/data, notifications, Tasks, telemetry, and timing are unchanged. `close()` teardown is the deliberate reverse-of-connect order (close the transport, then clear the tool map).

## Notes for reviewers (human-written)

The huge diff is because of server.ts -> legacy_server.ts
I checked it (also with Claude), all the changes seems to be ok.

## Proof it works

[...]

PR #1163, opened 2026-07-26
fix: Drop validator nulling in close() that threw on frozen entries

jirispilka:
## Why

Closing the server tried to modify read-only built-in tools and crashed before removing them.

`close()` set each tool's compiled AJV validator to `null`, then cleared the tool map. Default tool entries are frozen module-level singletons, so assigning to `ajvValidate` throws in ESM strict mode:

```text
TypeError: Cannot assign to read only property 'ajvValidate'
```

The exception stopped execution before `this.tools.clear()`. As a result, `ActorsMcpServer.close()` rejected and left the tool map populated.

This bug already existed. It was found while building the 2026-07-28 stateless adapter.

## What changed

The validator-nulling loop is deleted; `close()` now just clears the tool map. Review (thanks @MQ37) showed the loop did no useful work in any branch:

- Frozen entries (all default tools) must be skipped — that write is the crash.
- Payment-decorated clones share the frozen original's validator (`cloneToolEntry` restores it by reference), so nulling the clone frees nothing.
- Per-Actor and proxied validators are already detached from the shared AJV instance at compile time (`fixedAjvCompile` does `removeSchema` + scope scrub), so `tools.clear()` alone makes them collectible.

The regression test covers the crash: `close()` resolves and the tool map is empty afterwards, with a frozen built-in tool registered.

## Stacking

This PR is stacked on #1155 rather than `master`. Two further PRs stack on this one (#1164, #1165); both are rebased on top of this change.

## Verification

The regression test was written first and failed against the unmodified source with the `TypeError` above. It passes after the fix.

The following checks pass on this commit:

- `pnpm run type-check`
- `pnpm run lint`
- `pnpm run test:unit`
- `pnpm run check:agents`
- `pnpm run format:check`

Unit test result: 84 files, 1178 passed, 1 skipped.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -4,7 +4,6 @@
 
 import type { Transport } from '@modelcontextprotocol/sdk/shared/transport.js';
 import type { InitializeRequest } from '@modelcontextprotocol/sdk/types.js';
-import type { ValidateFunction } from 'ajv';
 
 import log from '@apify/log';
 import { parseBooleanOrNull } from '@apify/utilities';
@@ -477,18 +476,6 @@ export class ActorsMcpServer implements LegacyMcpServerHost {
         // server close are the adapter's transport-lifecycle responsibility), then clear the shared
         // tool map. The order is unobservable because `close()` only runs on a quiesced serving unit.
         await this.legacyServer.close();
-        this.clearTools();
-    }
-
-    /** Clear all tools and null the compiled schemas of the entries this instance may write. */
-    private clearTools(): void {
-        for (const tool of this.tools.values()) {
-            // A frozen entry is not writable — the write throws in strict mode, whether it's a
-            // shared module singleton or a per-facade payment-decorated clone.
-            if (!Object.isFrozen(tool) && tool.ajvValidate && typeof tool.ajvValidate === 'function') {
-                (tool as { ajvValidate: ValidateFunction<unknown> | null }).ajvValidate = null;
-            }
-        }
         this.tools.clear();
     }
 }
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -480,10 +480,12 @@ export class ActorsMcpServer implements LegacyMcpServerHost {
         this.clearTools();
     }
 
-    /** Clear all tools and null their compiled schemas. */
+    /** Clear all tools and null the compiled schemas of the entries this instance may write. */
     private clearTools(): void {
         for (const tool of this.tools.values()) {
-            if (tool.ajvValidate && typeof tool.ajvValidate === 'function') {
+            // A frozen entry is not writable — the write throws in strict mode, whether it's a
+            // shared module singleton or a per-facade payment-decorated clone.
+            if (!Object.isFrozen(tool) && tool.ajvValidate && typeof tool.ajvValidate === 'function') {
                 (tool as { ajvValidate: ValidateFunction<unknown> | null }).ajvValidate = null;
             }
         }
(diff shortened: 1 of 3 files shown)


MQ37:
Checked the logic here and I think we do not need the guard at all — plain `this.tools.clear()` is sufficient:

```ts
private clearTools(): void {
    this.tools.clear();
}
```

The loop it guards does no useful work in any case:

- **Frozen entries** (all 5 default tools) are skipped — that is the crash you fixed.
- **Payment-decorated clones** pass the guard and get nulled, but `cloneToolEntry` reassigns the *original's* validator onto the clone, so the frozen module singleton still holds it for the whole process. Frees nothing.
- **Per-Actor / proxied entries** are the only sole owners of a validator, and `fixedAjvCompile` already detaches those from the shared `ajv` at compile time (`removeSchema` + scope scrub). `tools.clear()` on the next line makes them unreachable — nulling a field on an object that is about to be dropped whole does not change what GC collects. I verified this with a WeakRef + forced-GC script: 300 entries, isolated processes, 299/300 validators collected either way (the survivor is the last loop iteration pinned in a stack slot, identical in both), heap within 0.5%.

So every branch is either mandatory-skip or a no-op. Deleting the loop removes the special case instead of routing around it, and it cannot throw again when someone freezes a new tool source later. The `ValidateFunction` type import at the top of `server.ts` becomes unused too.

The test is still worth keeping — `await expect(server.close()).resolves.toBeUndefined()` and `expect(server.listToolNames()).toEqual([])` prove the crash is gone regardless of which fix we pick. Only the two `ajvValidate` assertions would go away.

Minor: the new comment says frozen entries include "a per-facade payment-decorated clone" — those clones are `JSON.parse` output and are not frozen, they are exactly what passes the guard.

MQ37 (commented):
⏫

jirispilka:
> Checked the logic here and I think we do not need the guard at all — plain `this.tools.clear()` is sufficient:
> 
> ```ts
> private clearTools(): void {
>     this.tools.clear();
> }
> ```
> 
> The loop it guards does no useful work in any case:
> 
> * **Frozen entries** (all 5 default tools) are skipped — that is the crash you fixed.
> * **Payment-decorated clones** pass the guard and get nulled, but `cloneToolEntry` reassigns the _original's_ validator onto the clone, so the frozen module singleton still holds it for the whole process. Frees nothing.
> * **Per-Actor / proxied entries** are the only sole owners of a validator, and `fixedAjvCompile` already detaches those from the shared `ajv` at compile time (`removeSchema` + scope scrub). `tools.clear()` on the next line makes them unreachable — nulling a field on an object that is about to be dropped whole does not change what GC collects. I verified this with a WeakRef + forced-GC script: 300 entries, isolated processes, 299/300 validators collected either way (the survivor is the last loop iteration pinned in a stack slot, identical in both), heap within 0.5%.
> 
> So every branch is either mandatory-skip or a no-op. Deleting the loop removes the special case instead of routing around it, and it cannot throw again when someone freezes a new tool source later. The `ValidateFunction` type import at the top of `server.ts` becomes unused too.
> 
> The test is still worth keeping — `await expect(server.close()).resolves.toBeUndefined()` and `expect(server.listToolNames()).toEqual([])` prove the crash is gone regardless of which fix we pick. Only the two `ajvValidate` assertions would go away.
> 
> Minor: the new comment says frozen entries include "a per-facade payment-decorated clone" — those clones are `JSON.parse` output and are not frozen, they are exactly what passes the guard.

@MQ37 fair point, remove the clearTools() function to simplify it, 

Now it lives in 
```ts
    async close(): Promise<void> {
        // Reverse-of-connect (LIFO) teardown: take the transport/server down first (SIGINT removal +
        // server close are the adapter's transport-lifecycle responsibility), then clear the shared
        // tool map. The order is unobservable because `close()` only runs on a quiesced serving unit.
        await this.legacyServer.close();
        this.tools.clear();
    }
 ```

MQ37 (approved):

PR #1164, opened 2026-07-26
refactor: Parameterize tool composition and widget resolution

jirispilka:
## Why

Preparation for #1140. A 2026-07-28 stateless request has no handshake, so it must resolve its own mode, client identity and tool set per request. Today `composeToolsForClient`, `isReportProblemServable` and `getServerInstructions` read `this._serverMode`/`this._clientContext` directly, the fetched tool sources are dropped at the initialize flush, and widget resolution is neither mode-parameterized nor reusable. None of that can serve a per-request view.

This lands the restructuring on its own so the feature PR contains only new behavior.

## What changed

Those three methods now take an explicit `{ serverMode, clientContext }` view. Every existing caller passes the view it already read, so the sessionful path is unchanged. Fetched `{ input, actorTools }` sources are retained alongside the pending-flush queue instead of replacing it — keeping them separate matters, because composing all retained sources at the flush would recompose a source already composed against constructor recovery data against the newer initialize client. Widget resolution splits into a mode-parameterized read and the instance-level writer.

**One deliberate behavior change, and it is not behavior-preserving:** widget resolution is now memoized per facade. Previously `connect()` and `applyInitialize()` each re-scanned disk, so an explicitly-apps facade scanned twice and logged twice; now it scans once. A widget file appearing after a successful scan is no longer picked up. A *failed* scan is not memoized, so it still retries — that is tested.

`toolSources` and the extracted `resolveToolEntry` have **no reader until the stateless adapter lands**. They are written and exported here so the feature PR stays additive.

_Left for the author._

Stacked on #1163, which is stacked on #1155. The bulk of the `server.ts` diff is signature and call-site churn.

## Proof it works

`tests/unit/mcp.server.capability_gating.test.ts` and `tests/unit/mcp.server.report_problem_gating.test.ts` are **byte-identical to the base commit** — no assertion was touched. That is the behavior-preservation proof for tool composition and gating. It does not cover widget resolution, which is why the memoization change is called out above rather than claimed as a no-op.

`isReportProblemServable`'s new `view.clientContext != null` check is exactly what the previous `clientKnown` getter computed (`return this.clientContext != null`).

All five checks pass on this commit alone: 84 test files, 1178 passed + 1 skipped — unchanged from the fix commit beneath it, as behavior preservation implies.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DeTjFmDV3RK6ALz9MKZdPq)_

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -2,6 +2,9 @@
  * Model Context Protocol (MCP) server for Apify Actors
  */
 
+import { dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
 import type { Transport } from '@modelcontextprotocol/sdk/shared/transport.js';
 import type { InitializeRequest } from '@modelcontextprotocol/sdk/types.js';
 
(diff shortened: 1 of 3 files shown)


src/mcp/server.ts
```diff
@@ -303,14 +377,16 @@ export class ActorsMcpServer implements LegacyMcpServerHost {
      * known, queue the source so the initialize flush re-composes and adds the client-gated tools.
      */
     private registerFetchedActorTools(input: Input, actorTools: ToolEntry[]): void {
+        const source: ToolSource = { input, actorTools };
+        this.toolSources.push(source);
```
jirispilka: I was a bit fighting with this and I don't think it is needed.
We will see in the next PR if we can clean it up

MQ37 (approved):
LGTM — but before continuing (i.e. before #1140 builds on this) we should bound `toolSources` so it does not grow endlessly.

It is currently an unbounded array: pushed on every `registerFetchedActorTools` call, never popped, cleared, or read anywhere. Traced the real growth path in `apify-mcp-server-internal`: `handleExistingSession` runs on every request for a session already in memory, diffs Redis vs local tool state, and calls `loadToolsByName` on the same long-lived `mcpServer` instance whenever they drift — each call pushes another entry. Sessions use a rolling 30-minute TTL (extended on every request), so a continuously active session keeps that instance, and its growing `toolSources`, alive indefinitely.

jirispilka:
> LGTM — but before continuing (i.e. before #1140 builds on this) we should bound `toolSources` so it does not grow endlessly.
> 
> It is currently an unbounded array: pushed on every `registerFetchedActorTools` call, never popped, cleared, or read anywhere. Traced the real growth path in `apify-mcp-server-internal`: `handleExistingSession` runs on every request for a session already in memory, diffs Redis vs local tool state, and calls `loadToolsByName` on the same long-lived `mcpServer` instance whenever they drift — each call pushes another entry. Sessions use a rolling 30-minute TTL (extended on every request), so a continuously active session keeps that instance, and its growing `toolSources`, alive indefinitely.

Fixed, thanks!

PR #1172, opened 2026-07-28
test: Add MCP 2026-07-28 integration and conformance coverage

### LINKED ISSUE #1132: test: Add MCP 2026-07-28 integration and conformance coverage

### LINKED ISSUE #1171: chore: Bump MCP v2 SDK packages to 2.0.0 before releasing the stateless server
Part of #1128. Affects #1140 (PR #1165), which currently pins `2.0.0-beta.5`.

The SDK shipped stable `2.0.0` across the v2 packages ([release](https://github.com/modelcontextprotocol/typescript-sdk/releases/tag/%40modelcontextprotocol%2Fserver%402.0.0)). This is not a routine stability bump — `2.0.0` aligns the 2026-07-28 wire with the **final** spec revision (modelcontextprotocol/typescript-sdk#2513, spec PR modelcontextprotocol/modelcontextprotocol#3002), and `beta.5` speaks the pre-final shape:

- Envelope `clientInfo` demoted from required to optional: a beta.5 **server rejects conforming clients** that omit it.
- `serverInfo` moved from the `DiscoverResult` body to response `_meta` (new `SERVER_INFO_META_KEY`): a beta.5 **client misclassifies a conforming server as legacy** (probe fails → attempted `initialize` handshake → hard connect failure).

Shipping beta.5 would break exactly the 2026-07-28 clients the migration targets, so this must land before `createStatelessServer` is released.

## Scope

- Bump `@modelcontextprotocol/{server,node}` `2.0.0-beta.5` → `2.0.0` in the #1165 stack (or immediately after merge). The v1 `@modelcontextprotocol/sdk` line is unaffected (`1.30.0` exists but is not required here).
- Re-run the stateful/stateless parity unit suites.
- Re-run the conformance suite with a release matching the final wire — beta.5 validation used `@modelcontextprotocol/conformance@0.2.0-alpha.9`; check for a newer tag.
- Verify `serverInfo` stamping: 2.0.0 stamps `_meta['io.modelcontextprotocol/serverInfo']` automatically on every 2026-era response (a handler-authored value wins) — confirm the adapter's responses carry it; nothing should need authoring.
- No API changes expected: the release notes list no breaking changes to `createMcpHandler` / `isLegacyRequest` / `toNodeHandler` / `AuthInfo`. New behavior covered by the SDK itself: the stateless entry now answers `415` for POSTs whose `Content-Type` is not `application/json` (beta.5 didn't inspect it), and `keepAliveMs` applies to every SSE stream.

Downstream: apify/apify-mcp-server-internal#676 targets `2.0.0` in lockstep (internal bumps its own pins in its wiring PR).

jirispilka:
## Why

Closes #1132.
Closes #1171.
Part of #1128.

The `2026-07-28` stateless route had no integration or conformance coverage.

## What changed

- Added the v2 client as a third integration transport.
- Extended HTTP payment and cancellation cases to the stateless route.
- Added `pnpm run test:conformance` for both `2026-07-28` and `2025-11-25`.
- Added a separate reusable conformance workflow, called by the integration workflow.
- Added expected-failure ledgers for each protocol era.

## Dependency bump

**Release-age exception:** these versions are less than 72 hours old. This PR adds exact `minimumReleaseAgeExclude` entries, bypassing the repository's normal delay. Later releases remain subject to it.

⚠️ new version

- MCP v2 SDK: `2.0.0-beta.5` → stable `2.0.0`
- Conformance: `0.2.0-alpha.9` → `0.2.0-alpha.10`

The stateless route should not ship on beta SDK packages. Alpha.10 is the matching final-spec conformance runner. Beta.5 already contained the final identity shape, so no adapter workaround was needed.

## Testing

Type-check, lint, unit tests, format, and AGENTS checks pass locally. Integration and conformance run in CI; conformance can also run locally with `pnpm run test:conformance`.


MQ37 (approved):
Let's merge this — one thing first: please swap the test client to the stable `@modelcontextprotocol/client` release instead of pinning `2.0.0-beta.5`.

The runtime already runs `@modelcontextprotocol/server@2.0.0` and `@modelcontextprotocol/node@2.0.0`, so the beta pin pulls a second `@modelcontextprotocol/core@2.0.0-beta.5` into the tree and the new dimension ends up certifying the stable server against a beta client codec. No beta as an additional dependency — just the stable version.

jirispilka:
> Let's merge this — one thing first: please swap the test client to the stable `@modelcontextprotocol/client` release instead of pinning `2.0.0-beta.5`.
> 
> The runtime already runs `@modelcontextprotocol/server@2.0.0` and `@modelcontextprotocol/node@2.0.0`, so the beta pin pulls a second `@modelcontextprotocol/core@2.0.0-beta.5` into the tree and the new dimension ends up certifying the stable server against a beta client codec. No beta as an additional dependency — just the stable version.

Good point!

PR #1174, opened 2026-07-28
fix(server-card): declare conditional authentication as optional

### LINKED ISSUE #1170: fix: server card and server.json claim authentication is always required
**ℹ️  Maintainers note:** there was already discussion about changing the server.json authorization metadata to always required in https://github.com/apify/apify-mcp-server/pull/1174. The main issues is that the server.json spec does not support setting that authentication is required under some condition and we advertise in the server.json only the main `mcp.apify.com` URL that currently requires token. The MCP server might not require auth when the `?tools=` is specified with tools that do not require auth like `docs` for example, but that is out of server.json scope currently.

## Problem

`getServerCard()` (`src/server_card.ts`) emits `authentication: { required: true, schemes: ['bearer', 'oauth2'] }`. `server.json` emits `remotes[0].headers[0].isRequired: true`. Both unconditional — but the server does serve an anonymous surface.

Verified against production, no credentials:

| Endpoint | `initialize` | `tools/list` |
|---|---|---|
| `?tools=search-actors,fetch-actor-details,search-apify-docs,fetch-apify-docs` | **200** | **200** — returns those 4 tools |
| `/` (default) | **401** `invalid_token` | — |

Documented too: https://docs.apify.com/integrations/mcp names the 4 tools usable without a token.

## Why it matters

Discovery metadata is read *before* connecting. An agent that trusts `required: true` prompts for a token before anonymous exploration, or skips the server — friction on exactly the path we want frictionless.

## Fix

Auth is **conditional** on the `?tools` param, so neither `true` nor `false` is correct on its own. Options:

- `remotes[].headers[].isRequired: false` plus a description naming the anonymous toolset (what the old hand-maintained apify.com card did).
- Per-tool auth annotation.

`test/unit/server-card.test.ts` here and in the internal repo pins `required === true` — update alongside.

## Related

- apify/apify-mcp-server-internal#582 (lazy auth) — makes this worse: under lazy auth every connection starts anonymous, so `required: true` is always wrong.
- #790 (card shape rework) — touches `remotes[]` but not auth accuracy.
- apify/apify-web#6366 — apify.com now proxies this card verbatim, so this value ships to `apify.com/.well-known/mcp/server-card.json` as well.

huige66631:
## Problem

The server card declares `authentication.required: true`, and `server.json` marks the `Authorization` header as required. This is inaccurate for the hosted `?tools=` surface, where `search-actors`, `fetch-actor-details`, `search-apify-docs`, and `fetch-apify-docs` work without an Apify token.

## Solution

Declare authentication as optional in both discovery metadata formats and document that a token remains required for the default tool set and non-public tools. This keeps the metadata aligned with the existing runtime authorization policy without changing access control.

## Changes

- Set server-card authentication to optional while retaining bearer and OAuth2 schemes.
- Mark the `Authorization` header in `server.json` as optional and document the anonymous `?tools=` surface.
- Add unit coverage for both metadata declarations.

## Testing

- `pnpm run type-check`
- `pnpm run lint`
- `pnpm run test:unit`
- `pnpm run format:check`
- `pnpm run check:agents`
- Manual metadata and authorization-policy check: public tools remain anonymous; default and private tools still require a token.

## Notes for Reviewer

This PR intentionally does not change runtime authentication. `isApiTokenRequired()` continues to require a token when no tool selection is supplied or when any non-public tool is selected. The private hosted-server repository may need its matching server-card assertion updated when this change is released.

Fixes #1170

### DIFF
--- a/server.json
+++ b/server.json
@@ -14,8 +14,8 @@
       "headers": [
         {
           "name": "Authorization",
-          "description": "Apify API token for authentication with Apify platform services. For example 'Bearer <apify-api-token>'",
-          "isRequired": true,
+          "description": "Optional Apify API token. Required for the default tool set and non-public tools. Anonymous access is available with ?tools=search-actors,fetch-actor-details,search-apify-docs,fetch-apify-docs.",
+          "isRequired": false,
           "isSecret": true
         }
       ]
--- a/src/server_card.ts
+++ b/src/server_card.ts
@@ -55,7 +55,7 @@ export function getServerCard(): ServerCard {
             tools: { listChanged: true },
         },
         authentication: {
-            required: true,
+            required: false,
             schemes: ['bearer', 'oauth2'],
         },
         tools: 'dynamic',
--- a/tests/unit/server_card.test.ts
+++ b/tests/unit/server_card.test.ts
@@ -6,7 +6,17 @@ import { getServerCard, getServerInfo } from '../../src/server_card.js';
 import { readJsonFile } from '../../src/utils/generic.js';
 import { getPackageVersion } from '../../src/utils/version.js';
 
-const serverJson = readJsonFile<{ description: string }>(import.meta.url, '../../server.json');
+const serverJson = readJsonFile<{
+    description: string;
+    remotes: {
+        headers: {
+            name: string;
+            description: string;
+            isRequired: boolean;
+            isSecret: boolean;
+        }[];
+    }[];
+}>(import.meta.url, '../../server.json');
 
 describe('getServerCard', () => {
     it('should return a valid MCP server card object', () => {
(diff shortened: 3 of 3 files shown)


MQ37:
Thank you for the PR! I would honestly not merge this as this PR is saying that this server on URL `mcp.apify.com` does not require auth when in reality for this exact URL without any `?tools=` it does. Also the `server.json` SEP does not have any way to communicate under what conditions the bearer auth header is require or not so I would defensively keep it as required. What do you think? @jirispilka

huige66631:
Thanks, that makes sense. I agree that the default `mcp.apify.com` endpoint requires a token, and marking authentication as optional would make the server card and `server.json` misleading.

I will not pursue this metadata change as-is. Would you prefer me to close this PR, or is there a specific way you would like the public `?tools=` surface documented without weakening the required-auth contract?

jirispilka:
> Thank you for the PR! I would honestly not merge this as this PR is saying that this server on URL `mcp.apify.com` does not require auth when in reality for this exact URL without any `?tools=` it does. Also the `server.json` SEP does not have any way to communicate under what conditions the bearer auth header is require or not so I would defensively keep it as required. What do you think? @jirispilka

@huige66631 thank you but I agree with @MQ37 let's close it

PR #1175, opened 2026-07-28
feat: Add outputSchema to report-problem and get-actor-log

### LINKED ISSUE #1159: feat: Add outputSchema to report-problem
Follow-up from #1140. Part of #1128.

`report-problem` (`src/tools/dev/report_problem.ts`) declares no `outputSchema`. Every other helper tool does, except `get-actor-run-log` (#1160), which has the same deferred `TODO` pattern.

#1140 resolves each tool's `outputSchema` up front in the stateless `tools/call` shell and threads it into `projectCallToolResult` on every return path, so tools without one get no result projection and no `structuredContent` validation. Note the 2026-07-28 codec currently discards the schema argument, so today this is only observable on the 2025-era path — but the asymmetry is real and will surface when the codec starts honouring it.

## Blast radius (verified)

Smaller than originally assumed:

- The legacy v1 path advertises `outputSchema` in `tools/list` (`src/utils/tools.ts`) but never validates or projects results server-side — projection exists only in `src/mcp/stateless_server.ts`. The only v1 contract change is: schema advertised ⇒ success results must carry matching `structuredContent`. Shipping both in the same commit keeps spec-compliant clients happy.
- `apify-mcp-server-internal`: the contract suite asserts `structuredContent` only for `fetch-actor-details`; the multinode report-problem test checks tool presence only. Nothing breaks. Flag the PR so internal can add a matching contract case.

## Scope

- Add `reportProblemToolOutputSchema` to `src/tools/structured_output_schemas.ts`: `{ reported: boolean }`, required.
- `outputSchema` on the ToolEntry; return `respondOk(ACKNOWLEDGEMENT, { structuredContent: { reported: true } })`; drop the `TODO(#1159)` comment.
- Extend `tests/unit/tools.report_problem.test.ts`: assert `structuredContent` equals `{ reported: true }` and ajv-validates against the tool's `outputSchema`.

Deliberately left out of #1140 to keep that change to the stateless surface.

### LINKED ISSUE #1160: feat: Add outputSchema to get-actor-run-log
Follow-up from #1140. Part of #1128.

`get-actor-run-log` (`src/tools/runs/get_actor_run_log.ts`) declares no `outputSchema`.

Same root cause as the sibling issue #1159: #1140 threads each tool's `outputSchema` into `projectCallToolResult` on every `tools/call` return path, and tools without one get no result projection.

## Why this was not done in #1140

The inline code comment gives the original rationale: the log API returns plain text, so there is no natural structured shape — not a coordination deferral. Decision now: add a minimal `{ log: string }` wrapper anyway, for uniformity — every tool advertises an `outputSchema`, so the stateless projection applies everywhere. Accepted trade-off: the log text is duplicated on the wire (`content` + `structuredContent`), capped at 50 lines by the input schema.

## Blast radius (verified)

- The legacy v1 path advertises `outputSchema` in `tools/list` but never validates or projects results server-side — projection exists only in `src/mcp/stateless_server.ts`. Shipping schema + `structuredContent` in the same commit is the whole v1 contract story.
- `apify-mcp-server-internal`: no contract-suite case touches this tool's result shape. Nothing breaks; flag the PR so internal can add a matching case.

## Scope

- Add `getActorRunLogToolOutputSchema` to `src/tools/structured_output_schemas.ts`: `{ log: string }`, required.
- `outputSchema` on the ToolEntry; switch the return from `respondRaw` to `respondOk(text, { structuredContent: { log: text } })`; replace the "does not make sense" + `TODO(#1160)` comment.
- New `tests/unit/tools.get_actor_run_log.test.ts` (none exists today): stub `client.run().log().get()`, assert last-N-lines slicing, `structuredContent.log` matching the text content, and ajv-validation against `outputSchema`.

jirispilka:
## Why
Closes #1159, Closes #1160. `report-problem` and `get-actor-log` were the last helper tools with no `outputSchema`, so `structuredContent`-consuming clients got nothing machine-readable back, and the stateless server's result projection has no schema to project against.

## What changed
- `report-problem` now returns `structuredContent: { reported: true }`.
- `get-actor-log` now returns `structuredContent: { log: <text> }`, and fixes a pre-existing off-by-one in its line-tail slice: it used to return N+1 lines when N were requested (and, against real Apify run logs which end with a trailing newline, N-1 — a live `mcpc` probe caught this after the first fix attempt). It now returns exactly N.
- `lines:=0` documented as returning the entire log (existing behavior, not new — just previously unstated).
- Any impact on `apify-mcp-server-internal`: `structuredContent` is new on both tools' `tools/list` entries, and `get-actor-log`'s output text is one line shorter than before. Additive/low-risk, but the internal contract suite likely needs a matching test case for both.

## Proof it works

image

---
_Generated by [Claude Code](https://claude.ai/code/session_01GaQ7VgYvm3m865aY4HEWbE)_

### DIFF
--- a/src/tools/dev/report_problem.ts
+++ b/src/tools/dev/report_problem.ts
@@ -5,6 +5,7 @@ import type { ToolEntry, ToolInputSchema } from '../../types.js';
 import { TOOL_TYPE } from '../../types.js';
 import { compileSchema } from '../../utils/ajv.js';
 import { respondOk } from '../../utils/mcp.js';
+import { reportProblemToolOutputSchema } from '../structured_output_schemas.js';
 
 const REPORT_PROBLEM_DESCRIPTION = `Report a problem with Apify's MCP tools or Actors to the Apify team.
 
@@ -127,8 +128,7 @@ export const reportProblem: ToolEntry = Object.freeze({
     title: 'Report a problem',
     description: REPORT_PROBLEM_DESCRIPTION,
     inputSchema: reportProblemInputSchema,
-    // TODO(#1159): no `outputSchema` — the tool returns a fixed acknowledgement, so nothing here is
-    // affected by the `tools/call` result projection against an advertised schema.
+    outputSchema: reportProblemToolOutputSchema,
     ajvValidate: compileSchema(reportProblemInputSchema),
     paymentRequired: false,
     annotations: {
@@ -139,6 +139,6 @@ export const reportProblem: ToolEntry = Object.freeze({
         openWorldHint: false,
     },
     call: async () => {
-        return respondOk(ACKNOWLEDGEMENT);
+        return respondOk(ACKNOWLEDGEMENT, { structuredContent: { reported: true } });
     },
 } as const);
--- a/src/tools/runs/get_actor_run_log.ts
+++ b/src/tools/runs/get_actor_run_log.ts
@@ -4,11 +4,16 @@ import { HELPER_TOOLS } from '../../const.js';
 import type { InternalToolArgs, ToolEntry, ToolInputSchema } from '../../types.js';
 import { TOOL_TYPE } from '../../types.js';
 import { compileSchema } from '../../utils/ajv.js';
-import { respondRaw } from '../../utils/mcp.js';
+import { respondOk } from '../../utils/mcp.js';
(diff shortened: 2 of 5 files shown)


RobertCrupa (approved):
Thank you for resolving this!

PR #1178, opened 2026-07-29
docs: Update README to clarify agentic payment options and recommend AGI for token management

birosrichard:
## Why
Follow-up to [apify-web#6391](https://github.com/apify/apify-web/pull/6391). Agentic-payment info is scattered and out of sync across repos. Tracked in [apify-agi#89](https://github.com/apify/apify-agi/issues/89). This README described x402 and Skyfire as the only payment options with no mention of AGI, so a reader had no way to know AGI is now the recommended path for new integrations.

## What changed
**Before**: README led with x402/Skyfire as co-equal options; AGI wasn't mentioned anywhere.
**Now**: AGI added as the recommended path (top summary line, TOC, new "AGI (recommended)" section linking to `agi.apify.com/AGENTS.md` as the source of truth for its terms). x402 gets a one-line "narrower scope, Pay Per Event Actors only" callout. Skyfire is untouched, still actively supported.

Consulted/validated most of the stuff with @MQ37.

Btw.. Not really `adhoc` issue, but I cannot link https://github.com/apify/apify-agi/issues/89

## Proof it works
Docs-only change, no runtime behavior touched. What was fact-checked instead: live-confirmed `agi.apify.com/protocols` returns both `x402` and `mpp`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The Apify Model Context Protocol (MCP) server at [**mcp.apify.com**](https://mcp
 
 > ⚠️ **Legacy SSE transport removed.** The `https://mcp.apify.com/sse` endpoint has been removed in favor of Streamable HTTP. Migrate your client to **[`https://mcp.apify.com`](https://mcp.apify.com)** — drop the `/sse` suffix from your configuration.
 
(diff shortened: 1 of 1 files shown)


jirispilka:
Nice, I'll leave this for Kuba and Dan to review. 
If the goals is also to promote agi, this information should also be at: https://mcp.apify.com/

MQ37 (approved):
Thank you for the PR! I made a follow-up PR (#1179) with a few wording/structure refinements — please review that and if it looks good, merge them together and then into master.

PR #1179, opened 2026-07-29
docs: address review feedback on AGI/x402/Skyfire wording

MQ37:
## What
Follow-up to #1178: fixes the reviewer's pending suggestions (drop contradictory "no API token required" clause, token works against both `mcp.apify.com` and `api.apify.com`, single source-of-truth link), consolidates the "Pay Per Event only / no Standby Actor support" caveat into one shared note instead of duplicating it across the x402 and Skyfire sections, and renames the x402 section to "Direct x402" to disambiguate it from AGI's own x402 funding rail.

## Why
Same limitation applies to both x402 and Skyfire, so a per-section callout would've duplicated text; AGI mentions x402 too (as one of its funding methods), so the standalone x402 section needed a distinct name — "Direct x402" is what we call it internally.

## Testing
`pnpm run check:agents` (README link-tree validator) passes. Docs-only change.


birosrichard (approved):
Thanks 🙏🏼

PR #1185, opened 2026-07-30
fix: Honor AbortSignal on Actor-MCP calls and hash-dedupe proxy names

Ayush7614:
## Summary
- Forward the request/cancel AbortSignal into remote Actor-MCP client.callTool (dispatch + call-actor passthrough) so cancel/disconnect can stop remote work before the 120s timeout.
- Yield a macrotask in the abort catch before client.close() so the SDK's fire-and-forget notifications/cancelled can flush (otherwise close() aborts the same transport AbortController and the remote never sees cancel).
- Hash-suffix over-length proxied MCP tool names (same pattern as Actor tool names) so bare truncation cannot collide two origin tools into one exposed name.

## Notes
- handleMcpToolCall now passes EXTERNAL_TOOL_CALL_TIMEOUT_MSEC (120s), matching the dispatch path. That raises this route from the SDK default 60s to 120s for parity.

## Why
Actor and internal tools already honor abort. Actor-MCP only checked signal.aborted in the catch after failure — the SDK call never received signal. Proxy naming documented the collision risk and violated the capped + hash-deduped tool-name invariant in src/mcp/AGENTS.md.

## Test plan
- [x] pnpm run type-check
- [x] pnpm run lint
- [x] pnpm run test:unit (new proxy naming tests + AbortSignal forwarding assertions)
- [x] pnpm run format / pnpm run check:agents
- [ ] Optional: mcpc cancel against a long Actor-MCP tool call and confirm remote work stops early (and that notifications/cancelled reaches the remote)

### DIFF
--- a/src/mcp/tool_dispatch.ts
+++ b/src/mcp/tool_dispatch.ts
@@ -181,6 +181,10 @@ export async function dispatchToolCall(params: {
                     CallToolResultSchema,
                     {
                         timeout: EXTERNAL_TOOL_CALL_TIMEOUT_MSEC,
+                        // Same abort source as ACTOR/INTERNAL branches: request signal for sync,
+                        // cancel-watcher signal for tasks. Without this, cancel/disconnect cannot
+                        // stop the remote call before EXTERNAL_TOOL_CALL_TIMEOUT_MSEC.
+                        signal,
                     },
                 );
 
--- a/src/mcp/tool_dispatch.ts
+++ b/src/mcp/tool_dispatch.ts
@@ -181,9 +181,6 @@ export async function dispatchToolCall(params: {
                     CallToolResultSchema,
                     {
                         timeout: EXTERNAL_TOOL_CALL_TIMEOUT_MSEC,
-                        // Same abort source as ACTOR/INTERNAL branches: request signal for sync,
-                        // cancel-watcher signal for tasks. Without this, cancel/disconnect cannot
-                        // stop the remote call before EXTERNAL_TOOL_CALL_TIMEOUT_MSEC.
                         signal,
                     },
                 );
@@ -202,6 +199,11 @@ export async function dispatchToolCall(params: {
 
                 result = { ...res };
             } catch (error) {
+                if (signal.aborted) {
+                    // Yield a macrotask first: the SDK sends notifications/cancelled fire-and-forget on
+                    // the transport's AbortController, which the finally's close() would abort.
+                    await new Promise((resolve) => setImmediate(resolve));
+                }
                 ({ toolStatus, callDiagnostics } = buildExecutionDiagnostics({
                     error,
                     isAborted: Boolean(signal.aborted),
--- a/src/tools/actors/call_actor.ts
+++ b/src/tools/actors/call_actor.ts
@@ -1,5 +1,6 @@
 import type { Client } from '@modelcontextprotocol/sdk/client/index.js';
 import type { ContentBlock } from '@modelcontextprotocol/sdk/types.js';
+import { CallToolResultSchema } from '@modelcontextprotocol/sdk/types.js';
 import dedent from 'dedent';
 import { z } from 'zod';
 
@@ -15,6 +16,7 @@ import {
 } from '../../const.js';
 import { ACTOR_LOAD_ERROR_KIND, ActorLoadError } from '../../errors.js';
 import { connectMCPClient } from '../../mcp/client.js';
+import { EXTERNAL_TOOL_CALL_TIMEOUT_MSEC } from '../../mcp/const.js';
 import type { PaymentProvider } from '../../payments/types.js';
 import type { ActorInfo, ApifyToken, InternalToolArgs, ToolEntry, ToolInputSchema } from '../../types.js';
 import { TOOL_TYPE } from '../../types.js';
@@ -324,8 +326,10 @@ export async function handleMcpToolCall(params: {
     mcpServerUrl: string | false;
     apifyToken: string;
     mcpSessionId?: string;
+    signal: AbortSignal;
 }): Promise<ToolResponse | null> {
-    const { baseActorName, mcpToolName, input, isActorMcpServer, mcpServerUrl, apifyToken, mcpSessionId } = params;
+    const { baseActorName, mcpToolName, input, isActorMcpServer, mcpServerUrl, apifyToken, mcpSessionId, signal } =
+        params;
 
     if (!isActorMcpServer) {
         return respondServerError(`Actor '${baseActorName}' is not an MCP server.`);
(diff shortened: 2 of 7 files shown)


vojtechj-apify (commented):
Nice work @Ayush7614 - this found two real problems, and the tool-name fix is straightforwardly correct.

The abort forwarding is right too, and it's a real improvement: cancel now returns immediately instead of holding the call for up to 120s. One piece is missing, and it isn't visible from the client side - the explicit `notifications/cancelled` never reaches the remote, because `client.close()` in the `finally` aborts the same `AbortController` the notification rides on. I caught it by standing up a live remote and logging what arrived. The SDK only documents `signal` as raising an AbortError locally, so you got exactly what's specified.

Two comments below with the fix - one macrotask yield in each catch. Verified against a real remote: cancellation delivered, remote handler aborts, no added latency, unit suite green. This same race silently breaks the pre-existing 120s timeout path too, so your PR plus this patch fixes something older than your change.

One thing to note in the description: passing `EXTERNAL_TOOL_CALL_TIMEOUT_MSEC` in `handleMcpToolCall` moves that call from the SDK default 60s to 120s. Parity with the dispatch path makes sense, but it doubles worst-case latency on that route, so it's worth stating.

Two optional nits as suggestions, plus one on a test. Take or leave them.

Process note, and the timing is on us: we published new contribution guidelines a few days after you opened this - issue-first, AI disclosure, evidence. Not applying them retroactively. Worth a read before the next one: [CONTRIBUTING.md](https://github.com/apify/apify-mcp-server/blob/master/CONTRIBUTING.md#before-you-write-code)

src/tools/actors/call_actor.ts
```diff
@@ -369,6 +388,9 @@ export async function handleMcpToolCall(params: {
             },
         });
     } catch (error) {
+        if (signal.aborted) {
+            return respondAborted();
+        }
```
vojtechj-apify, line 392: The cancellation never reaches the remote. `Protocol.cancel()` sends `notifications/cancelled` fire-and-forget on the transport's shared `AbortController`, and the `finally`'s `client.close()` aborts that controller — killing the POST before it hits the wire. One macrotask yield fixes it, since the `finally` runs after the catch body.

```suggestion
        if (signal.aborted) {
            // Yield a macrotask first: the SDK sends notifications/cancelled fire-and-forget on the
            // transport's AbortController, which the finally's close() would abort before it flushes.
            await new Promise((resolve) => setImmediate(resolve));
            return respondAborted();
        }
```

src/mcp/tool_dispatch.ts
```diff
@@ -181,6 +181,10 @@ export async function dispatchToolCall(params: {
                     CallToolResultSchema,
                     {
                         timeout: EXTERNAL_TOOL_CALL_TIMEOUT_MSEC,
+                        // Same abort source as ACTOR/INTERNAL branches: request signal for sync,
+                        // cancel-watcher signal for tasks. Without this, cancel/disconnect cannot
+                        // stop the remote call before EXTERNAL_TOOL_CALL_TIMEOUT_MSEC.
+                        signal,
```
vojtechj-apify, line 184: Same fix needed in this branch's catch (line 204) — can't be a suggestion, it's below the hunk:

```diff
             } catch (error) {
+                if (signal.aborted) {
+                    // Yield a macrotask first: the SDK sends notifications/cancelled fire-and-forget on
+                    // the transport's AbortController, which the finally's close() would abort.
+                    await new Promise((resolve) => setImmediate(resolve));
+                }
                 ({ toolStatus, callDiagnostics } = buildExecutionDiagnostics({
```

src/tools/actors/call_actor.ts
```diff
@@ -337,17 +341,32 @@ export async function handleMcpToolCall(params: {
...
         }
 
-        const result = await client.callTool({
-            name: mcpToolName,
-            arguments: input,
-        });
+        if (signal.aborted) {
+            return respondAborted();
+        }
+
```
vojtechj-apify: Nit, not a defect: this second check is redundant. `client.callTool` calls `signal.throwIfAborted()` before anything reaches the wire, so an already-aborted signal lands in your catch below and returns the same `respondAborted()` — I confirmed no `tools/call` is sent. The pre-connect guard at :344 does earn its keep, since it skips the connection entirely. Fine to keep this for symmetry if you prefer.

```suggestion
```

src/mcp/tool_dispatch.ts
```diff
@@ -181,6 +181,10 @@ export async function dispatchToolCall(params: {
                     CallToolResultSchema,
                     {
                         timeout: EXTERNAL_TOOL_CALL_TIMEOUT_MSEC,
+                        // Same abort source as ACTOR/INTERNAL branches: request signal for sync,
+                        // cancel-watcher signal for tasks. Without this, cancel/disconnect cannot
+                        // stop the remote call before EXTERNAL_TOOL_CALL_TIMEOUT_MSEC.
```
vojtechj-apify: Nit: lines 1–2 restate what `dispatchToolCall`'s own doc comment says further up, and line 3 is PR context that'll go stale in the code. `signal,` reads fine on its own. See AGENTS.md § Communication style.

```suggestion
```

tests/unit/tools.call_actor_common.test.ts
```diff
@@ -331,4 +335,80 @@ describe('call_actor_common', () => {
...
+            expect(options.signal).toBe(controller.signal);
+            expect(options.timeout).toBe(EXTERNAL_TOOL_CALL_TIMEOUT_MSEC);
+        });
+
+        it('returns aborted when the signal is already aborted before the remote call', async () => {
+            const connectSpy = vi.spyOn(mcpClient, 'connectMCPClient');
+            const controller = new AbortController();
+            controller.abort();
+
+            const result = await handleMcpToolCall({
```
vojtechj-apify, line 377: Nit: this spies on `connectMCPClient` without a stub, so it calls through. Inert today because the guard short-circuits — but if that guard ever regresses, the test attempts a real connection to `example.invalid` instead of failing cleanly. `.mockResolvedValue(null)` makes the assertion the only thing that can fail.

Ayush7614:
Thanks @vojtechj-apify — addressed in a0a108c.

**Required fix:** macrotask yield (`setImmediate`) in both abort catches (`handleMcpToolCall` + `dispatchToolCall` ACTOR_MCP branch) so `notifications/cancelled` can flush before `finally` `client.close()` aborts the transport controller.

**Nits taken:**
- Dropped the redundant post-connect abort check
- Removed the stale `signal` comment in `tool_dispatch`
- Stubbed `connectMCPClient` with `.mockResolvedValue(null)` in the already-aborted test

**PR description:** noted that `handleMcpToolCall` now uses `EXTERNAL_TOOL_CALL_TIMEOUT_MSEC` (120s) for parity with dispatch (was SDK default 60s).

Unit suite still green.

vojtechj-apify (approved):
Thanks for the fix, looks all good now!

Issue #675, opened 2026-04-15
Align Actor input-schema transformation with the canonical Apify pattern


jirispilka:
## Context

While fixing #637, we found that `src/tools/utils.ts` and `src/utils/ajv.ts` reimplement
input-schema normalisation logic that is already solved — and battle-tested — in an
internal Apify library on the platform side. #637 patched the symptom (MCP tool listings
showing required fields as optional) by switching `fixZodSchemaRequired` to a value-check.
The root corruption — `filterSchemaProperties` writing `key: undefined` for every
whitelisted key — was left in place with a TODO. This issue tracks the structural follow-up.

## What's wrong today

`src/tools/utils.ts::filterSchemaProperties` (line 120) does:

```ts
filteredProperties[key] = {
    title: property.title,
    description: property.description,
    enum: property.enum,
    type: property.type,
    default: property.default,    // ← assigned even if upstream has no default
    prefill: property.prefill,
    ...
};
```

Consequences:
1. Phantom `default: undefined` on every property → any `'default' in field` check
   returns `true` incorrectly (which is how #637 manifested).
2. The in-memory shape doesn't match the on-disk Apify Actor schema → confusing when
   dumping a schema during debugging.
3. `tests/unit/tools.utils.test.ts:725-727` asserts this bug
   (`arrayContaining([..., 'default', 'prefill', ...])`) so the codebase actively resists
   fixing it.

## The canonical pattern

Apify's platform-side validator (internal library) does input-schema normalisation
correctly:
- Operates on the original Actor schema directly; no intermediate filter/rewrite step.
- Reconciles `required` vs `default` with a value-check (`field.default !== undefined`).
- Bumps `minItems` to ≥ 1 for required array fields (we don't do this today).
- Provides helpers for merging defaults, stripping UI-only fields, and extracting
  prefill/example values — all of which `buildActorInputSchema`'s 7-step transformation
  chain reinvents piecemeal.

We can't import the internal library into the public MCP server. The right move is to
mirror the algorithms and invariants, and delete the duplicated scaffolding here.

### Phase 1 — stop the corruption (minimal)

- [ ] `filterSchemaProperties`: preserve only keys whose upstream value is not `undefined`.
  One-line change (conditional spread per key).
- [ ] Update `tests/unit/tools.utils.test.ts:725-727` assertion to expect only the keys
  actually defined upstream.
- [ ] Revert the defensive value-check in `fixZodSchemaRequired` back to the original
[...]

PR #1186, opened 2026-07-30
fix: Preserve Actor input-schema validation keywords

Ayush7614:
## Summary
- Keep Actor-declared JSON Schema validation keywords (minimum, maximum, minLength, maxLength, minItems, maxItems, exclusive bounds, uniqueItems, pattern) through filterSchemaProperties.
- Only emit keys whose upstream value is defined — no more phantom default: undefined on every property.
- Extend SchemaProperties so the type matches what AJV and tools/list actually need.

## Why
filterSchemaProperties whitelisted only title/description/enum/type/default/prefill/properties/items/required. Real Actor schemas declare bounds (e.g. maxResults maximum: 100), but MCP-side AJV never saw them, so out-of-range inputs reached Apify runs and failed late. Agents also never saw those constraints in tools/list.

## Test plan
- [x] pnpm run type-check
- [x] pnpm run lint
- [x] pnpm run test:unit (new filter + AJV enforcement cases; updated transform assertions)
- [x] pnpm run format / pnpm run check:agents
- [x] Spot-check tools/list for an Actor with maximum/minLength and confirm bounds appear

### DIFF
--- a/src/tools/actor_input_schema.ts
+++ b/src/tools/actor_input_schema.ts
@@ -80,14 +80,15 @@ export function buildApifySpecificProperties(
 }
 
 /**
- * Filters schema properties to include only the necessary fields.
- * This is done to reduce the size of the input schema and to make it more readable.
+ * Filters schema properties to the fields MCP tools need: core shape, examples/defaults,
+ * and JSON Schema validation keywords. UI-only Actor hints (`editor`, `enumTitles`, …) are
+ * dropped to keep advertised schemas smaller.
  *
- * TODO(#675): This object literal unconditionally assigns every whitelisted key,
- * including `default: undefined`, on properties that didn't declare them upstream.
- * This creates phantom keys that broke `fixZodSchemaRequired()` via key-presence
- * checks (#637). The symptom is patched in `src/utils/ajv.ts` (value-check), but
- * this function should only emit keys whose upstream value is not `undefined`.
+ * Only emits keys whose upstream value is not `undefined` — assigning `default: undefined`
+ * (and similar) created phantom keys that broke `'default' in field` checks (#637 / #675).
+ *
+ * Validation keywords (`minimum`, `maximum`, `minLength`, …) are preserved so AJV compile
+ * and tools/list honor Actor-declared bounds instead of silently accepting out-of-range input.
  *
  * @param properties
  */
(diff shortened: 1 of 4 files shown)


jirispilka:
@RobertCrupa please check this PR when you have time, thanks!

RobertCrupa:
Thanks for the PR @Ayush7614, and for the detailed test plan.

Could you reduce this to just the #675 Phase 1 scope? Two things need to come out, and one needs to go in.

### Please remove: the validation keywords

`minimum`/`maximum`/`minLength`/`exclusiveMinimum`/etc. were never emitted before, so this is a new feature rather than a preservation fix, and it isn't covered by #675.
In phase 2 of the PR, we deliberately park the related question of whether we need `filterSchemaProperties` at all, so I'd rather not widen the list ahead of that.

There's also a concrete hazard as written. These keywords now reach `fixedAjvCompile`, and AJV 8 type-checks them at compile time:

```
{ minimum: 0, exclusiveMinimum: true }  → COMPILE ERROR: exclusiveMinimum must be number
{ uniqueItems: 'true' }                 → COMPILE ERROR: uniqueItems must be boolean
{ minLength: '2' }                      → COMPILE ERROR: minLength must be integer
```

`actor_tools_factory.ts:141-148` catches compile errors, so the whole Actor drops out of `tools/list`. The old whitelist stripped these keys, so this class of schema always compiled. 

If you think the bounds are worth surfacing to clients, please open an issue, and we'll scope it properly.

### Please add: #675 Phase 1 item 3

The issue lists three items. This PR does 1 and 2; item 3 is reverting `fixZodSchemaRequired` to the key-presence check (`'default' in fieldSchema`), which is safe once upstream stops emitting phantom keys.

Right now the PR keeps the value-check and rewrites the comment to justify it, and removes the `TODO(#675)` marker while the issue is still open.

### Implementation

#675 specifies a conditional spread per key, which avoids the helper, the keys array, and the `as Record<string, unknown>` cast:

```ts
filteredProperties[key] = {
    type: property.type,
    title: property.title,
    ...(property.enum !== undefined && { enum: property.enum }),
    ...(property.default !== undefined && { default: property.default }),
    // …
};
```

Two smaller notes:

- The docstring's invariant doesn't hold end-to-end. `shortenProperties` runs after `filterSchemaProperties` and does `property.enum = filterAndShortenEnum(property.enum)`, which returns `undefined` for an empty list — so an `enum: undefined` own-key comes back. Harmless (AJV's `enum` handler checks `Array.isArray`), but worth not claiming otherwise.
- `.not.toHaveProperty('editor')` at lines 758-759 and 96 passes unchanged on `master`, so it doesn't add coverage — `editor` was never whitelisted, and the `tags` fixture has no top-level `editor` key.

MQ37:
We are filtering the fields because back in the day this caused issues with quite a lot of MCP clients - before merging this change we would need to verify that it works with at least major MCP clients.

cc @jirispilka @RobertCrupa

jirispilka:
Closing this PR as inactive, it has merge conflicts, better to start from scratch

Issue #776, opened 2026-04-30
refactor: Split the shared integration suite by capability


jirispilka:
## Problem

`tests/integration/suite.ts` contains 3,292 lines of shared integration cases.

The file is difficult to navigate and creates frequent conflicts, but it has an important role: one `createIntegrationTestsSuite` factory runs the same cases against:

- stdio
- legacy streamable HTTP
- 2026-07-28 stateless HTTP

Splitting cases into standalone `*.test.ts` files would duplicate runner setup or lose the shared transport matrix.

## Goal

Keep one parametric suite entry point while moving capability-specific case registration into smaller modules.

## Scope

Keep `createIntegrationTestsSuite(options)` in `tests/integration/suite.ts`.

Move groups of existing tests into non-test modules under a directory such as:

```text
tests/integration/cases/
    registration.cases.ts
    tools.cases.ts
    actors.cases.ts
    storage.cases.ts
    tasks.cases.ts
    apps.cases.ts
    payments.cases.ts
```

The exact groups should follow the current `describe` structure. Do not create empty or artificial capability files only to match this example.

Each module exports a registration function called by `createIntegrationTestsSuite`. The modules must not register tests at import time or run as independent Vitest entry points.

Preserve:

- existing test names and bodies
- transport gates
- client creation and cleanup
- suite hooks
- sequential/concurrent behavior
- the shared stdio, legacy HTTP, and stateless HTTP matrix

## Dependency scoping

Before extracting each group, identify which suite values it uses.

Pass only those values to its registration function. Do not introduce one broad `IntegrationSuiteContext` containing every helper and option.

Keep helpers local to a capability module when they have one consumer. Move a helper to a shared integration utility only when multiple capability modules use it.

## Delivery

Move one coherent capability group per PR. Each PR is mechanical: file movement and imports only.

Update `AGENTS.md` and `DEVELOPMENT.md` when the first capability module lands so new tests are added to the owning capability module rather than directly to `suite.ts`.

## Non-goals

- No new test cases or assertions.
- No test deletions.
- No production-code changes.
- No transport-matrix changes.
- No unrelated helper rewrites.
- No standalone per-transport copies of capability tests.

jirispilka:
@MQ37 can you please check these tests too: https://github.com/apify/apify-mcp-server/issues/777

MQ37:
## Addendum: sharing critical cases with `apify-mcp-server-internal`

[...]

PR #1196, opened 2026-08-03
docs: Add integration test suite rework plan

MQ37:
## What
Proposal doc for splitting `tests/integration/suite.ts` into per-concern modules (`tools/` + `protocol/`) and sharing `critical: true` scenarios with `apify-mcp-server-internal` instead of hand-duplicating them there.

## Why
Discussion/planning doc only — no code changes. Opening as draft to gather feedback before starting the actual split.

## Testing
N/A — docs only.

### DIFF
--- a/docs/tests-rework-plan.md
+++ b/docs/tests-rework-plan.md
@@ -1,0 +1,112 @@
+# Integration test suite rework
+
+## What
+
+Split `tests/integration/suite.ts` (3292 lines, 115 cases, one file) into per-concern
+modules. Share the majority of cases with `apify-mcp-server-internal` instead of
+hand-duplicating ~40 of them there.
+
+## Why
+
+One file, 115 cases, unmaintainable by more than one person at a time. Internal
+re-implements a third of the same assertions against a live deploy — drifts, doubles
+maintenance cost, no shared source of truth.
+
+## Things to decide
+
+Two top-level directories, not one flat pile:
+
+```
+tests/integration/
+  tools/                          # mirrors src/tools/ — tool-owned behavior
+    actors/{call_actor,fetch_actor_details,search_actors,mcp_server_actor}.ts
+    docs/docs_tools.ts
+    runs/actor_run.ts
+    storage/storage_tools.ts
+    widgets/widget_contracts.ts
+  protocol/                       # mirrors src/mcp/ + src/utils/server_mode.ts
+    tool_loading.ts                # selectors/categories/env-vars — registration, not a tool
+    tasks.ts
+    notifications_cancellation.ts
+    server_mode.ts
+    payments/{skyfire,x402}.ts
+  actor.server_stateless.test.ts   # unchanged: real vitest entry points, compose
+  actor.server_streamable.test.ts  # every register*Tests() into one sequential
+  stdio.test.ts                    # describe block, same as today
+```
+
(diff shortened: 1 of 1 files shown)


jirispilka (commented):
I would keep it even simpler, as outlined in this issue: https://github.com/apify/apify-mcp-server/issues/776

MQ37:
@jirispilka could you specify what would you simplify? the directory structure?

the core idea behind this is that this will build foundation for moving everything into a single public repo and having only specific internal logic in the internal repo.

docs/tests-rework-plan.md
```diff
@@ -0,0 +1,112 @@
...
+One file, 115 cases, unmaintainable by more than one person at a time. Internal
+re-implements a third of the same assertions against a live deploy — drifts, doubles
+maintenance cost, no shared source of truth.
+
+## Things to decide
+
+Two top-level directories, not one flat pile:
+
+```
+tests/integration/
```
jirispilka, line 20: I would keep the tests flat, like this (just an idea). 
Let's not run to the opposite corner :) (from one file to hundreds of separate files)

```
tests/integration/cases/
    registration.cases.ts
    tools.cases.ts
    actors.cases.ts
    storage.cases.ts
    tasks.cases.ts
    apps.cases.ts
    payments.cases.ts
```

jirispilka:
> @jirispilka could you specify what would you simplify? the directory structure?

This: https://github.com/apify/apify-mcp-server/pull/1196/changes#r3710623293

docs/tests-rework-plan.md
MQ37: On one hand I like how simple the flat structure is but on the other hand I like the idea that you can easily find a test case for a specific tool just by opening a file since you know the `search-actor` tools tests would be in `tools/actors/search-actor.ts`.

RobertCrupa: Personally, I am more inclined to the proposed idea with the folders; although I'm not so keen on mirroring the `src/` directory, I think we will end up with more folders than needed.

Having one folder per domain (tools, transport, payments, etc.) and then grouping many files by scenario/flow is how I would go about it.

e.g.

```
tests/
  integration/
    tools/
      tools-call-apify-actor.test.ts
    prompts/
      prompts-list.test.ts
    resources/
      resource-read.test.ts
    tasks/
      task-cancel.test.ts
    transport/
      stdio-bootstrap.test.ts
      sse-bootstrap.test.ts
      streamable-http-bootstrap.test.ts
    shared/
      fixtures.ts
      client.ts
      assertions.ts
```

The issue I see with this is that when a test overlaps two domains, it can become unclear where to place it.

RobertCrupa (approved):
good stuff!

docs/tests-rework-plan.md
jirispilka: I would not mirror it too. I would keep it flat in the beginning (without folders), unless more test should be grouped together

vojtechj-apify (approved):

MQ37:
Thank you for the comments gang :sunglasses: 

So let's keep it simple and go with the flat approach that @jirispilka proposed first and we will see - we can always restructure things if needed.

MQ37:
Converted to issue #1199 — flat cases/ structure (per @jirispilka's proposal) instead of the nested tools/+protocol/ layout in this doc.

PR #725, opened 2026-04-20
feat: Shorten task ID to <tool-name>-<12-char-suffix>

jirispilka:
Closes #705.

## Summary
- Task IDs were long and noisy — `call-tool-apify--rag-web-browser-550e8400-e29b-41d4-a716-446655440000` (~70 chars).
- New form: `${tool.name}-${randomBytes(9).toString('base64url')}` → `apify--rag-web-browser-x7kQ9mZpA1bC` (~35 chars).
- Uses Node built-in `randomBytes`; no new dependency (avoids pulling `nanoid` for one call).
- Uses `tool.name` (canonical, slash-safe) rather than raw request `name` (which can be the legacy `apify-slash-…` form).

## MCP spec compliance
- `taskId` MUST be a string — ✓
- MUST be unique — ✓ (prefix + 72-bit random suffix; session-scoped `taskStore`)
- MUST be receiver-generated — ✓
- Entropy: 72 bits, plus the spec's preferred auth-context binding via `mcpSessionId` (see `src/mcp/server.ts:543`).

Spec ref: https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks

## Changes
- `src/mcp/server.ts` — swap `randomUUID` import for `randomBytes`; replace the task-ID expression.
- `res/mcp_task_reference.md` — update stale `// 32 hex chars` comment.
- `tests/integration/suite.ts` — add format assertion to the existing task-flow test.

## Test plan
- [x] `npm run type-check` — passes
- [x] `npm run lint` — passes
- [x] `npm run test:unit` (515/515) — passes
- [ ] `npm run test:integration` — requires `APIFY_TOKEN`; task-flow suite includes a new format assertion
- [ ] mcpc probe: `tools-call <actor> task:='{"ttl":60000}' <args>` → inspect `taskCreated.taskId` shape
- [ ] Non-task path: `tools-call <actor> <args>` returns unchanged `CallToolResult` (regression check)

https://claude.ai/code/session_012N7gXd5tWLp9Xic5JDVif9

Issue #709, opened 2026-04-17
[Bug]: `call-actor` tool result doesn't conform MCP standard


jancurn:
### Server type

Local (stdio via npx @apify/actors-mcp-server)

### MCP Client

Claude Desktop

### Operating System

None

### What happened?

According to
https://modelcontextprotocol.io/specification/2025-11-25/server/tools#structured-content:

Structured content is returned as a JSON object in the `structuredContent` field of a result.
For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block.

Now it looks as follows - the content[0] vs. structuredContent differs:

<img width="1727" height="785" alt="Image" src="https://github.com/user-attachments/assets/450e0984-5f7f-4fdd-9729-594bcb5a14f5" />

### Steps to reproduce

mcpc @apify tools-call apify--rag-web-browser query:="something" maxResults:=3

### MCP server configuration

```json
N/A
```

### Error logs

```text

```

jirispilka:
Audit across all tools shows this isn't unique to `call-actor` — every tool in the repo has some form of `content` ↔ `structuredContent` shape mismatch, and most carry LLM-facing guidance only in text where `structuredContent`-only clients can't see it.

Opened #744 to track standardization (shared builder, `summary` + `nextSteps` fields, phased migration). This issue will be fixed as the Phase 1 pilot and serve as the first concrete usage of the new shape.

---
_Generated by [Claude Code](https://claude.ai/code)_

jirispilka:
Fixed. `content[0]` is now the exact JSON serialization of `structuredContent` — verified live against a real run with `jq '(.content[0].text|fromjson) == .structuredContent'` (deep equality): the `apify--rag-web-browser` repro, `get-actor-run`, and `call-actor` all match. `content[1]` carries the prose narrative separately.

Shared builders: `buildGetActorRunResponse` (`src/tools/runs/get_actor_run.ts:73`) and `buildStartRunResponse` (`src/tools/actors/actor_run_response.ts:761`). Pinned by `tests/unit/tools.get_actor_run.response.test.ts`.

Remaining tools tracked in #744 — including `src/tools/storage/storage_helpers.ts:67-71`, which still doesn't mirror (`content[0]` omits the `summary`/`nextStep` that `structuredContent` carries) and is inherited by every storage tool.

jancurn:
Great stuff, cheers

Issue #670, opened 2026-04-14
test: Cover the cancelled-task result-storage guard


jirispilka:
## Context

The original bug (402 path in `executeToolAndUpdateTask()` stored a result on an already-cancelled task) was fixed incidentally by #1125, which factored all storage paths through a shared `skipIfTaskCancelled()` helper (`src/mcp/task_execution.ts:191`). Guarded paths: pre-`working` cancel-race, success, 402, permission-approval, execution-error.

## Gap

No test exercises `skipIfTaskCancelled` — grep for it in `tests/` returns nothing. The invariant is spec-mandated ("once a task is cancelled it MUST remain cancelled") and easy to regress when that catch block is next refactored.

## Scope

Unit tests in `tests/unit/`: task cancelled while the tool throws a payment error → no `storeTaskResult` call, telemetry status `ABORTED`. Same for the success and execution-error paths.

jirispilka:
Verified against current `master`: the bug described here is fixed, so I'm rescoping the issue to the part that is still open.

What is still missing is test coverage. Nothing in `tests/` exercises `skipIfTaskCancelled` — grep returns no hits. The invariant is spec-mandated ("once a task is cancelled it MUST remain cancelled") and easy to regress the next time that catch block is refactored. Rescoping the title and body to that.

PR #1202, opened 2026-08-05
refactor: Split integration suite into a shared Case-model test-kit

### LINKED ISSUE #776: refactor: Split the shared integration suite by capability

MQ37:
**Cross-repo:** ships with apify-mcp-server-internal PR chain [#709](https://github.com/apify/apify-mcp-server-internal/pull/709) → [#724](https://github.com/apify/apify-mcp-server-internal/pull/724) → [#725](https://github.com/apify/apify-mcp-server-internal/pull/725) — merge this chain first, internal depends on it.

**Chain:** **PR1 (this)** → [PR2: CaseCtx bugfixes](https://github.com/apify/apify-mcp-server/pull/1230).

## What

Replaces the 3292-line monolithic integration suite with flat `Case[]` arrays
per capability, plus a `test-kit` package export so `apify-mcp-server-internal`
consumes them directly instead of hand-duplicating assertions (issue #776).

## What's in this PR

- Flat `tests/test_kit/cases/*.cases.ts` per capability, replacing
  `tests/integration/suite.ts`.
- `isDeploymentTest` flag per case (17 marked true — runs against internal's
  live deploy automatically); `getFixture` for shared expensive setup;
  `withClient` + MCP client factories exported via `test-kit`.

## Review follow-ups (all addressed)

- **test-kit location + Docker**: moved `test_kit` from `src/` to
  `tests/test_kit/` (test-only tooling, not production code); fixed Docker
  builder stage that never copied the relocated dir.
- **Renames + dead field**: `critical`/`criticalOnly` → `isDeploymentTest`/
  `isDeploymentTestOnly`; removed unused `CaseCtx.hasTasksSupport`.
- **stdio coverage regressions**: `tasks.cases.ts`'s 9 cases and 3
  dynamic-tools-client cases in `tools.cases.ts` were skipping on stdio via a
  copy-pasted `onlyLegacyHttp` predicate — stdio supports tasks (same client
  API as legacy HTTP) and the transport restriction on the dynamic-tools
  cases had no remaining justification. Both fixed to run on stdio; verified
  live in CI.
- **Real npm-install packaging bug**: `tests/test_kit/**` imported production
  code via relative paths into `src/` (e.g. `../../src/const.js`) — works in
  a monorepo checkout, breaks on a real tarball install (only `dist/` ships).
  Routed every such import through the package's own
  `@apify/actors-mcp-server/internals.js` self-reference; added a
  `tsconfig.json` `paths` alias so the no-build type-check job still resolves
  it; promoted `@modelcontextprotocol/client` to an optional peer dependency
  (needed by `test-kit` at runtime, never installed for a real consumer).
  Verified against a genuinely packed tarball installed fresh with no
  monorepo context — reproduced the reported error, then confirmed the fix.
- **8 code-review nits**: renamed `skipIf` predicates by skip behavior, not
  transport (`notStdio`→`skipOnStdio` etc. — the old names read backwards);
  `createMcpStdioClient` now honors the shared `token` contract instead of
  only reading `process.env.APIFY_TOKEN`; dropped a stale comment
  contradicting its own `isDeploymentTest: false`; `build:watch` no longer
  races two `tsc -b` watchers; `DEVELOPMENT.md` no longer links a deleted
  file; `withClient` now passes `(client, ctx)` instead of forcing a nested
  `withClient(...)(ctx)` invocation; replaced 9 IIFE-wrapped cases with plain
  objects + named module constants; narrowed an oxlint override to what's
  actually needed.

## Testing

`type-check`, `lint`, `format`, unit suite (91 files / 1268 tests) green.
Integration suite needs a human with `APIFY_TOKEN` — not run by an agent.

## AI disclosure

Written with AI assistance (Poisson/Claude); reviewed and directed throughout.

closes https://github.com/apify/apify-mcp-server/issues/776


jirispilka (commented):
I understand why you want to export the tests and import them into the internal repository. Having a common suite for both repositories sounds attractive, but I don’t think this is the right direction.

We have created tight coupling between these repositories before, and we are now trying to reduce it. Sharing the test suite would increase that coupling through shared test structures, helpers, dependencies, and release coordination.

It is fine to duplicate a small number of tests in both repositories. The internal tests have been stable and rarely need changes, so the duplication cost is low.

I would keep these cases under tests/integration/cases, remove the exported test kit, and let the internal repository own its smoke and hosted-specific tests. I don’t want us to repeat the coupling problem we are trying to fix.

Let's discuss in person :)

tests/test_kit/helpers.ts
jirispilka, line 1: I believe we already had this file, this is be `git mv` instead of plain copy

jirispilka:
We discussed in person.

Let's try to use it this way, even though I don't like that we create a new coupling. It is pretty non-standard and we should document the reasoning.

1 - Please document it in CLAUDE, AGENTS.md so that's clear that test-kit

The test kit should not live under `src`. It is not production server code, and keeping it there makes the source layout misleading. 

2 - Please move it to a top-level directory and build it separately:

```text
src/                 # production server code
tests/               # local test entry points
test_kit/            # exported test-kit source
    cases/
    helpers.ts
    index.ts
    mcp_client.ts
    register.ts
    types.ts
```

The separate build should continue producing `dist/test_kit`, so the existing package import remains unchanged:

```ts
import { allCases, registerCases } from '@apify/actors-mcp-server/test-kit';
```

`CaseCtx` is a reasonable dependency-injection interface. It lets each repository provide its own client and environment while running the same cases. Keeping `criticalOnly` is also fine because internal can run the smaller live-deployment smoke suite without running every expensive integration case:

```ts
registerCases('shared critical cases', allCases, {
    createClientFn,
    transport: '2025-11-25',
    criticalOnly: true,
    // Other environment dependencies.
});
```

The `getApifyToken` and `getApifyApiBaseUrl` additions solve a real problem. Public tests normally use `APIFY_TOKEN` and `APIFY_API_BASE_URL`, while internal uses a test-user token and an environment-specific API URL. Without both values, an MCP call can start an Actor on staging while the direct `ApifyClient` verification accidentally queries production.

3 - I would still prefer passing one client factory instead of two separate getters:

```ts
createApifyClient: () =>
    new ApifyClient({
        token: getTestToken(),
        baseUrl: settings.apiBaseUrl,
    });
```

Cases would then use:

```ts
const apifyClient = ctx.createApifyClient();
```

This keeps the token and API URL together, avoids exposing credentials to individual cases, and prevents `CaseCtx` from gaining another field whenever `ApifyClient` needs more configuration.

So requested changes are:

- Keep the exported test kit, `CaseCtx`, `critical`, and `criticalOnly`.
- Move the test-kit source outside `src` and give it a separate build.
- Keep the existing `@apify/actors-mcp-server/test-kit` package import.
- Prefer `createApifyClient` over separate token and API URL getters.
- Preserve all internal-only auth, rate-limit, Redis, replay, and multi-node tests.
- Verify the packed package from a temporary consumer, including Vitest peer resolution.

MQ37:
Addressed:

- Moved `test_kit` out of `src/` into `tests/test_kit/`, built via TS project references — `dist/test_kit/index.js` and the `./test-kit` export unchanged, zero impact on `apify-mcp-server-internal`.
- Replaced `getApifyToken`/`getApifyApiBaseUrl` with a single `ctx.createApifyClient()` factory — token+baseUrl stay together, never exposed separately to case code.
- Documented in `DEVELOPMENT.md` why `CaseCtx` couples the two repos: MCP contract tests live here once, internal keeps only what's genuinely deployment-specific (databases, rate limiting, auth, multi-node).
- Rate-limit burst test is now wired into the multinode suite too (existed as an opt-in helper, was never called there).
- Verified the packed tarball from a temporary consumer (internal), including Vitest peer resolution. I also verified locally and ran the integration suites in both repos on my machine.

The replay/resumability tests for the `2025-11-25` protocol version were intentionally left deleted — the new protocol version has no such mechanism, and testing it properly required a hand-rolled low-level MCP client. My take: add telemetry to see if this feature is actually used; if not, remove it completely — it needs an additional Redis store and complicates the server.

Dockerfile
```diff
@@ -14,6 +14,7 @@ COPY src/web/package.json ./src/web/
 RUN pnpm install --frozen-lockfile
 
 COPY src ./src
+COPY tests/test_kit ./tests/test_kit
```
jirispilka: We should not ship tests within the dockerfile, it is bad

Claude is saying:

 // package.json
-    "build:core": "tsc -b src && tsc -b tests/test_kit",
-    "build:watch": "tsc -b src -w & tsc -b tests/test_kit -w & wait",
+    "build:core": "tsc -b src",
+    "build:watch": "tsc -b src -w",
+    "prepack": "tsc -b tests/test_kit",
     "clean": "tsc -b src --clean && tsc -b tests/test_kit --clean && rm -rf dist/web",

 // Dockerfile — drop the added line
 COPY src ./src
-COPY tests/test_kit ./tests/test_kit
 COPY tsconfig.json ./

jirispilka:
I'm sorry, this is a real pain to review :(, long pr, long time to create it, long time to deliver it

Verified with Codex:
The packed @apify/actors-mcp-server/test-kit export does not load.

  I verified this directly in apify-mcp-server-internal. I packed the public branch, replaced internal’s local link: dependency with the resulting tarball, and ran:

  node -e "import('@apify/actors-mcp-server/test-kit')"

  It fails with:

  ERR_MODULE_NOT_FOUND:
  Cannot find module
  '.../node_modules/@apify/actors-mcp-server/src/const.js'
  imported from
  '.../node_modules/@apify/actors-mcp-server/dist/test_kit/helpers.js'

  tests/test_kit/helpers.ts imports ../../src/const.js. After compilation into dist/test_kit, that specifier still points to <package>/src/const.js. The package publishes dist, not src.

  The internal tests work with the current local link: because it points to the complete public worktree. That makes src/*.ts available, and the TypeScript-aware test resolver masks the invalid packaged path. Installing the
  tarball exposes the failure.

  A clean consumer also finds an earlier problem: dist/test_kit/mcp_client.js imports @modelcontextprotocol/client, but that package is declared only in devDependencies. After installing it explicitly, the src/const.js
  failure above occurs.

  Please fix both package-loading issues and add CI coverage that:

  1. Builds and packs the package.
  2. Installs the tarball into a clean temporary consumer.
  3. Imports @apify/actors-mcp-server/test-kit with plain Node.

jirispilka:
I'm sorry, I don't really have strength to review this tonight, so here you go from my agent, apologies for the slop:

1. Misleading skipIf predicate names

  The new `skipIf` helpers are named for the transports on which a case runs, although `skipIf` interprets `true` as skip. For example, `skipIf: notStdio` skips stdio, while `skipIf: onlyStdio` skips everything except
  stdio.

  This naming already contributed to the task cases accidentally skipping stdio earlier in this PR. Please rename them by skip behavior, for example `skipOnStdio`, `skipUnlessStdio`, and `skipUnlessLegacyHttp`.

  2. Shared token option is ignored by stdio

  This PR gives `createMcpStdioClient` the new shared `SuiteClientOptions`, including its documented `token` behavior, but the stdio implementation ignores `options.token`. It validates and forwards only
  `process.env.APIFY_TOKEN`.

  Consequently:

  - `{ token: null }` still gives the child process the environment token.
  - `{ token: 'explicit-token' }` ignores the explicit token.

  Please honor the shared token contract or use a stdio-specific options type that does not expose `token`.

  3. Deployment-test comment contradicts the new flag

  The new `isDeploymentTest` justification above the rental-Actor case contradicts `isDeploymentTest: false`.

  Please either set the flag to `true` if this case should run against internal deployments, or remove/reword the justification.

  4. New build watchers overlap

  The new `build:watch` starts both `tsc -b src -w` and `tsc -b tests/test_kit -w`. Because the test-kit project references `src`, the second build-mode watcher also watches and rebuilds `src`.

  That leaves two processes owning `dist/*.js` and `dist/tsconfig.tsbuildinfo`. Please use one build-mode watcher over the project graph, or otherwise ensure only one watcher emits the `src` project.

  6. Deleted file remains documented

  This PR deletes `tests/const.ts`, but `DEVELOPMENT.md` still references it twice:

  - The Test Actors section links to `./tests/const.ts`.
  - The Test structure section says it contains test constants.

  Please update both references to the new location. `check:agents` does not detect this because it validates only the AGENTS.md link tree.

  ### Refactor artifacts introduced here

  7. withClient omits the new context

  The new `withClient(clientOptions, testFn)` helper passes only the client to `testFn`. The new `CaseCtx.createApifyClient` use in `actors.cases.ts` therefore has to build a `withClient` closure and immediately invoke it as `withClient(...)(ctx)`.

  Please pass `(client, ctx)` to `testFn`. Existing one-argument callbacks remain valid, and context-dependent cases no longer need the nested invocation.

  8. Nine IIFEs introduced by the array conversion

  The conversion to `Case[]` introduced nine `(() => { ... })()` wrappers in `actors.cases.ts` and `registration.cases.ts`. They preserve block scope from the old `itc` statements, but most now exist only to declare one
  short constant.

  Please replace those with inline values or clearly named module constants where appropriate.

  10. New Oxlint override is broader than required

  This PR disables `vitest/expect-expect` across all of `tests/test_kit/**`, although `register.ts` already suppresses that rule at the indirect `runIt(...)` calls.

  Please remove the directory-wide `expect-expect` override unless there is a demonstrated additional false positive. Keep `no-standalone-expect` disabled for `Case.run` bodies, since those assertions execute inside tests
  created by `registerCases`.

RobertCrupa (approved):
This is a very low risk change so I've skimmed through and I'm happy with the refactor.

If it works and the agents are happy with it, LGTM!

MQ37:
Fixed in `0810e63b` — routed every `tests/test_kit/**` import through the package's own `@apify/actors-mcp-server/internals.js` self-reference instead of relative `../../src/*` paths, added a `tsconfig.json` paths alias for the no-build type-check job, and made `@modelcontextprotocol/client` an optional peer dependency. Verified against a real `pnpm pack` tarball installed fresh with no monorepo context — reproduced your exact error, then confirmed it's gone.

MQ37:
All 8 fixed in `6d16c5d4` (+ `28000f32` for a second stale doc link the first pass missed): renamed `skipIf` predicates by behavior, `createMcpStdioClient` now honors the token contract, dropped the stale comment, `build:watch` only watches `src`, fixed both `DEVELOPMENT.md` links, `withClient` passes `(client, ctx)`, all 9 IIFEs replaced with named constants, oxlint override narrowed to `no-standalone-expect`.

jirispilka (changes requested):
Please remove the exports from ./internals. We're trying to shrink that surface, not grow it. Putting test-only symbols in there was the wrong call from the start, and it makes worse.

I had an agent check every open PR in internal at its head SHA, plus all branches: 12 of the 13 new exports aren't used anywhere. HELPER_TOOLS is the exception, #670 imports it.

Remove the other 12 and we're good to go.

tests/test_kit/helpers.ts
MQ37: Checked — git already treats this as a rename via content-similarity, regardless of `git mv` vs. a manual move: `git show a69af33c --stat` shows `{src => tests}/test_kit/helpers.ts`, and `git log --follow` traces the full history through it correctly. `git mv` is just a convenience wrapper that stages delete+add in one step — it doesn't produce different rename metadata than doing it manually. Nothing to fix here.

Dockerfile
MQ37: Agreed, fixed in `d96b6b87` — Dockerfile-only change rather than restructuring `build:core`/adding a `prepack` script, so `build:core`/`build:watch`/`test:integration`'s existing meaning for local dev stays untouched. The builder stage now runs `tsc -b src` directly instead of `pnpm run build`, so `tests/test_kit` is never copied in or compiled into the shipped image. Confirmed via CI's Docker build & smoke test job (passes) and locally — no `test_kit` output anywhere in the build result.

jirispilka (approved):

Issue #1064, opened 2026-07-08
fix: Sweep defects — tool metadata, config, payment client


jirispilka:

jirispilka:
Most of the issues were stale, closing

PR #1210, opened 2026-08-06
fix: Align Node floor to 22 and drop dead smithery.yaml

jirispilka:
Part of #1064.

**Node floor**: `manifest.json` said `>=20.0.0` and the `stdio.ts` guard checked `< 20`, while `package.json` `engines.node` already required `>=22.0.0`. Node 20 hit EOL 2026-04-30. Both now match `engines.node`.

`.nvmrc` and the Dockerfile stay at 24 — dev/build pins, not consumer floors.

**smithery.yaml**: pointed at `dist/main.js`, which the build does not produce (stdio entry is `dist/stdio.js`). No workflow or script referenced it.

The README Smithery badge is left in place — removing it is a distribution call, not a config fix.


RobertCrupa (approved):
Nice clean up!

PR #1205, opened 2026-08-05
feat: Port workflow evals to Langfuse

RobertCrupa:
## Why

Closes #1211 

We needed to pick one eval backend. Six were spiked: Braintrust, Opik, Harbor, Arize AX, Bedrock AgentCore, and Langfuse. 

Decision on the issue: collect evals in managed Langfuse now, move to our own deployment later.

## What changed

Before: `pnpm run evals:workflow` ran test cases from local JSON, printed a table, and optionally wrote `results.json` for baseline diffing.

Now: it upserts `test_cases.json` into the Langfuse `workflow-evals` dataset, runs the filtered dataset items as an experiment, and records traces, scores, and dataset runs in Langfuse. Agent and judge LLM calls are traced, and each MCP tool call is a tool observation under its item's trace.

The byte/token baseline diff is gone. Those two metrics are now scores (`total_tokens`, `result_bytes`) alongside `workflow_judge`

Needs `LANGFUSE_PUBLIC_KEY`, `LANGFUSE_SECRET_KEY`, and `LANGFUSE_BASE_URL`. The runner exits before running anything if any are missing.

The Phoenix tool-selection suite at the `evals/` root is untouched. Migrating that off Phoenix is a separate issue.

## Notes for reviewers (human-written)

Let me know if you need access to the managed langfuse instance.

Env variables can be found on our 1Pass

## Proof it works

Live run against Langfuse Cloud: the experiment is recorded as a dataset run and linked to the `workflow-evals` dataset.

https://langfuse.apify.dev/project/cmshkde21000krg07shb46d8g

PR #1220, opened 2026-08-10
chore: Drop the --lines test-case filter

RobertCrupa:
Stack 2/6, splitting up #1205. Base: #1219.

## Why

`--lines` selected workflow test cases by their line span in `test_cases.json`. Nothing uses it, and it is not free: `loadTestCasesWithLineNumbers` re-read the file, string-searched for each `"id": "..."`, then walked characters counting braces to find where each case ended, attaching `_lineStart`/`_lineEnd` to every case.

## What changed

Before: `--lines 277-283` ran the cases overlapping those lines.
Now: gone. `--category` and `--id` cover the same filtering by names that survive editing the JSON.

Deletes `evals/shared/line_range_parser.ts`, `evals/shared/line_range_filter.ts`, the brace matcher, and the README section.

## Proof it works

`pnpm run type-check`, `lint`, `test:unit`, `check:agents` pass. No test covered `--lines`, and nothing in `.github/` or `package.json` passes it.

---


RobertCrupa:
Dropping this: `--lines` is staying. It is carried through the stack and wired into the Langfuse runner in #1224

PR #1221, opened 2026-08-10
refactor: Own the workflow test case schema in the loader

RobertCrupa:
Stack 2/5, splitting up #1205. Base: #1219.

## Why

Two problems in `test_cases_loader.ts`:

1. It cast the parsed JSON to `WorkflowTestCase` and then hand-checked four fields for emptiness. Unknown keys passed. A test case written with `failTool` instead of `failTools` loaded fine, injected no failure, and the eval went green without testing what it claimed to test.
2. The default path was `path.join(process.cwd(), 'evals/workflows/test_cases.json')`, so which file a run read depended on where it was started from.

## What changed

- Strict zod schema owned by the loader. An unknown key now fails the load.
- `resolveTestCasesPath()` resolves the default from the module, and `--test-cases-path` against cwd, once.
- `WorkflowTestCase` moves out of `evals/shared/types.ts` into the loader. Only the workflow evals use it.

`loadTestCasesWithLineNumbers` (which backs `--lines`) had its own copy of the same hand-checks. It now reuses the single validated load and only computes the line spans.

Also drops the loader's `filterTestCases` wrapper, which just forwarded to the shared `filterById`/`filterByCategory`. Those are now called directly. It would otherwise be the only caller-less export left in the file.

## Proof it works

The real `test_cases.json` parses under the strict schema, all 30 cases, and the line spans still land on the right objects:

```
$ pnpm exec tsx -e "...loadTestCasesWithLineNumbers(DEFAULT_TEST_CASES_PATH)..."
totalLines 220
search-google-maps: 5-9
search-add-call-python-actor: 11-16
instagram-posts: 18-22
```

Six new unit tests in `tests/unit/evals.test_cases_loader.test.ts` cover unknown-key rejection, duplicate ids, and path resolution. Full gate passes.

---


jirispilka:
I'll leave this one to @MQ37 as he his more familiar with the workflow code.

MQ37 (approved):
Nice 👍 LGTM - I would maybe trim the comments a bit

PR #1222, opened 2026-08-10
fix: Normalize judge verdict casing instead of erroring the item

RobertCrupa:
Stack 3/5, splitting up #1205. Base: master (#1219 and #1221 have landed).

## Why

The judge requests a strict JSON schema (`JUDGE_RESPONSE_SCHEMA`, `strict: true`), but OpenRouter honours that only on some provider routes. When it is dropped, the judge still reaches a verdict - it just comes back as `"pass"` instead of `"PASS"`.

## What changed

`parseJudgeResponse` now validates with Zod, mirroring `WorkflowTestCaseValidator` in the loader:

## Scope

Casing and whitespace only, because that is the failure that was actually observed. Deliberately not handled: verdicts nested under a wrapper key, alternative key names, and prose-only responses. Those stay hard errors, and the thrown message still prints the raw response so the next occurrence identifies itself.

## Proof it works

Test added first, failing on the old code with the real defect:

Passes after the fix. Three cases cover it: lowercase normalizes to `PASS`, `"maybe"` still rejects, an unknown extra key keeps the verdict. Full gate passes.

workflow eval still run fine.

---
Written with Claude Code: it interviewed me on the design, wrote the test and fix, and ran the verification gate.

### DIFF
--- a/evals/workflows/workflow_judge.ts
+++ b/evals/workflows/workflow_judge.ts
@@ -5,6 +5,7 @@
 
 // eslint-disable-next-line import/extensions
 import type { ResponseFormatJSONSchema } from 'openai/resources/shared';
+import { z } from 'zod';
 
 import { JUDGE_PROMPT_TEMPLATE, MODELS } from './config.js';
 import type { LlmClient } from './llm_client.js';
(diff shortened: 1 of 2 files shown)


MQ37 (approved):
Nice, thank you and generally LGTM. The only thing we forgot is passing the parse error to the judge so it can be also judged as parse failure and judge has the context.

**Problem:** `formatConversationForJudge` reads only `toolCalls` + `finalResponse`, never `toolResults`. On a parse failure the call now survives (good) but shows up as `args: {}` with no failure marker — judge scores it as a normal empty-args call.

**Fix:** pair `toolCalls[i]` with `toolResults[i]` (always 1:1 by construction) and append `[FAILED: <error>]` when `success === false`. Reuses the existing success flag instead of a parse-specific special case, so it also fixes the same blind spot for real tool-execution failures.

jirispilka:
@RobertCrupa do we need to fix this? Won't it be replaced by claude harness from the Anthropic SDK?

RobertCrupa:
We would still need this for the Judges output as it has the same issue.

Let me double check how Claude harness would change this.

RobertCrupa:
@jirispilka, you are right, this fix won't be needed when the Claude SDK lands, but it will still be needed for the judge's output, so this fix is staying.

RobertCrupa:
@MQ37 Thank you for the comment. I will keep this in mind for when I add the Claude code harness.

This PR will now only focus on the judge's output. It is considerably smaller now.

jirispilka (approved):
LGTM 👍🏻

PR #1223, opened 2026-08-10
feat: Trace workflow evals to Langfuse

RobertCrupa:
Stack 4/5, splitting up #1205. Base: #1222.

## Why

We have no place to share evals across teams; someone has to pull in the repo and run the evals in their terminal. Now anyone can view our dashboard on Langfuse and see for themselves.

A failed eval could only be debugged from the terminal scrollback. `--verbose` dumped the conversation, and that dump was the only record of what the agent actually did. Nothing persisted per-turn tool arguments, results, or what the run cost.

## What changed

Runs export to Langfuse over OpenTelemetry. One trace per test case, the agent's LLM calls as generations with token usage, every MCP tool call as a tool observation with its arguments and result.

New required env vars: `LANGFUSE_PUBLIC_KEY`, `LANGFUSE_SECRET_KEY`, `LANGFUSE_BASE_URL`. 

## Notes for reviewers (human-written)

Let me know if you need to be added to our langfuse instance

## Proof it works

Live run against Langfuse Cloud: the experiment is recorded as a dataset run and linked to the workflow-evals dataset.

https://langfuse.apify.dev/project/cmshkde21000krg07shb46d8g

---
Written with Claude Code: it produced the split and the commits from an existing branch (#1205), and ran the verification gate.

PR #1224, opened 2026-08-10
feat!: Run workflow evals as Langfuse experiments

RobertCrupa:
Stack 5/5, splitting up #1205. Base: #1223.

## Why

- The harness owned its own results pipeline: a committed `results.json` that doubled as the default `--baseline`.
- Comparing two runs meant diffing that file by hand.
- Test cases lived only in the repo, so nobody could inspect or edit them outside a checkout.

#1223 already sends traces to Langfuse. This makes Langfuse the results store and the test case store too.

## What changed

A run reads its test cases from a Langfuse dataset and executes them as an experiment.

**The dataset is the source of truth.** A run fetches it and never writes to it, so editing an item in the Langfuse UI changes the next run. 

**Breaking, for this harness only:**

- `--test-cases-path` and `--lines` are removed, replaced by `--dataset`. Line numbers in a file that no longer feeds the run describe nothing; to run a variant set, clone the dataset in the UI.

## Other changes

Deletes `output_formatter.ts`, `results_writer.ts`, the committed `results.json` and the now-unused test case loader and line-range filters.

Scoped to the results and test case stores only: the hand-rolled agent (`llm_client.ts`, `mcp_client.ts`, `conversation_executor.ts`) is untouched, since it is being replaced by the Claude Agent SDK next.
## Proof it works
https://langfuse.apify.dev/project/cmshkde21000krg07shb46d8g/experiments
---


MQ37 (commented):
In general LGTM. But I found some issues that we should address before approving.

**Dataset sync vs. UI edits** — the README says "Editing an item in the Langfuse UI changes the next run," but I'm not sure that holds: `syncDataset()` runs from the local `test_cases.json` before every run, and `createItem()` upserts by id, so wouldn't an existing item's UI edit get overwritten back to the repo file first? Might be missing something — can you confirm the intended behavior here?

Also I think we should reduce the duplication and make the code more maintainable — move stuff like `workflow_judge` to a const so there's no risk of typos and it's type checked. For example:

```ts
export const SCORE_NAMES = { WORKFLOW_JUDGE: 'workflow_judge', TOTAL_TOKENS: 'total_tokens', RESULT_BYTES: 'result_bytes' } as const;
// then: name: SCORE_NAMES.WORKFLOW_JUDGE, and e.name === SCORE_NAMES.WORKFLOW_JUDGE everywhere
```

A typo in any of the 5 current occurrences breaks the pass/fail gate silently since it's a runtime string match, not a checked reference.

Also there are some possible stale comments:

`run_workflow_evals.ts:40-41` — *"Before anything reads process.env..."* placed above `sanitizeProcessEnv()`. `shared/config.ts`'s `OPENROUTER_CONFIG` is a module-level const reading `process.env.OPENROUTER_BASE_URL`/`API_KEY` at import time, and ESM hoists imports before a module's own statements, so that read always happens first regardless of placement. Not a live bug (the fields self-sanitize independently), but the comment's guarantee doesn't hold as written.

evals/workflows/run_workflow_evals.ts
jirispilka, line 1: I believe that with Anthropic SDK this will be be either gone or heavily refactored.
I'm gonna skip it in the review

evals/workflows/README.md
jirispilka, line 1: You need to describe the flow in the README

Create dataset -> run experiment -> results -> judge

evals/workflows/README.md
```diff
@@ -73,7 +56,20 @@ Tests AI agents executing tasks using Apify MCP server tools through multi-turn
 
 ## Critical Design Decisions
 
-### 1. MCP Server Isolation Per Test
+### 1. The dataset item is the source of truth
+
+**Decision:** Every run upserts `test_cases.json` into the Langfuse dataset first, then executes the matching dataset items.
```
jirispilka: Do we really want to create a new dataset for each run?

evals/workflows/README.md
```diff
@@ -73,7 +56,20 @@ Tests AI agents executing tasks using Apify MCP server tools through multi-turn
...
+
+**Why:**
+- `experiment.run` only records a comparable **dataset run** (with a shareable run URL) when it is given real dataset items
+- One source of truth means no id-matching contract to maintain between a local file and a remote dataset
```
jirispilka: Can we get rid of local file and only used langfuse as a source of truth?
That would remove all the friction with syncing. And we would have one ground truth in the DB

evals/workflows/README.md
```diff
@@ -181,7 +145,7 @@ Separation allows independent optimization for speed vs evaluation quality.
 
 **Location:** `config.ts`
 
-### 7. MCP Server Instructions in System Prompt
+### 8. MCP Server Instructions in System Prompt
```
jirispilka: In all the heading it should be like this:
## 8. MCP Server Instructions in System Prompt -> ## 8. MCP server instructions in system prompt

It is pre-existing but you already touched the README

jirispilka (approved):
Honestly, since we are moving this into Anhtropic SDK, I would put focus on the current workflow setup and make it right with the SDK

Because we are reviewing the code that's gonna be changed anyway. 

I understand why you did but since this is not prod use, I would quickly move to Anthrpic SDK and we would review that with @MQ37 properly

jirispilka:
Also, we will rename the workflow to evals only. Because that will be the default way to do it.

@RobertCrupa Please trim all the comments to a reasonable length, it is super long :D

evals/workflows/README.md
RobertCrupa: Yes I have made it so langfuse is the source of truth and all experiements run from there.

evals/workflows/README.md
RobertCrupa: This never happened so i'm not sure why it says that

evals/workflows/run_workflow_evals.ts
RobertCrupa: Agreed, some changes just have to happen to support exporting to langfuse, they will be undone soon

evals/workflows/README.md
RobertCrupa: Done [b272bcb](https://github.com/apify/apify-mcp-server/pull/1224/commits/b272bcb8655dcb1db1a25d65c52308059a30d7ea)

evals/workflows/README.md
RobertCrupa: all evals start by reading from the langfuse dataset [5f88c95](https://github.com/apify/apify-mcp-server/pull/1224/commits/5f88c953f959830bf2f923410b3b5e6c57fd6955)

evals/workflows/README.md
RobertCrupa: [4e91459](https://github.com/apify/apify-mcp-server/pull/1224/commits/4e9145913fe70bc67f50ff11c602ad41b8d02643) Done

jirispilka:
@RobertCrupa I missed it earlier, the PR title should not have been `feat!` because there was nothing breaking from user perspective

PR #1230, opened 2026-08-10
fix: Wire getApifyToken/getApifyApiBaseUrl through CaseCtx

### LINKED ISSUE #776: refactor: Split the shared integration suite by capability

MQ37:
Closes #776

**Cross-repo:** ships with apify-mcp-server-internal PR chain [#709](https://github.com/apify/apify-mcp-server-internal/pull/709) → [#724](https://github.com/apify/apify-mcp-server-internal/pull/724) → [#725](https://github.com/apify/apify-mcp-server-internal/pull/725) — merge this chain first, internal depends on it.

**Chain:** [PR1: Case-model architecture](https://github.com/apify/apify-mcp-server/pull/1202) → **PR2 (this)**.

## What

Two real bugs in the direct-`ApifyClient` polling path (`waitForRunAborted`,
run-status checks), found once internal ran these cases against its own dev
platform instead of production.

## Why

- **Token**: 4 call sites read `process.env.APIFY_TOKEN` directly instead of
  the case's own token — silently unauthenticated under internal, 180s hang
  instead of a clear failure.
- **BaseUrl**: same call sites had no environment override, always resolving
  to production. First fix attempt was a no-op — `ApifyClient`'s wrapper
  discarded any caller-supplied `baseUrl`; fixed in the same commit.

Both are required on `CaseCtx`, not defaulted — a missing wire-up is a
compile error, not a silent wrong-environment call. Superseded below by
`createApifyClient`.

## Review follow-ups (rebased from #1202, all addressed)

- **createApifyClient**: collapsed the two always-together getters
  (`getApifyToken`/`getApifyApiBaseUrl`) into one `ctx.createApifyClient()`
  factory — keeps token+baseUrl together, stops exposing the raw token to
  case code. `DEVELOPMENT.md` documents why `CaseCtx` couples the two repos
  at all (MCP contract tests live here once; internal keeps only
  deployment-specific concerns).
- **Renames + dead field**: rebased onto #1202's `isDeploymentTest` rename;
  found and fixed a real regression mid-rebase (this branch's own
  `createApifyClient` commit had silently dropped `CaseCtx.hasTasksSupport`)
  before removing it again cleanly once confirmed dead.
- **stdio coverage regressions**: rebased onto #1202's two stdio-coverage
  fixes, one conflict resolved keeping both sides.
- **Real npm-install packaging bug**: rebased onto #1202's `internals.js`
  routing fix — this branch's own `createApifyClient()` call sites needed
  the same self-reference-import treatment; one extra leftover
  (`tests/test_kit/types.ts`'s relative `src/apify_client.js` import,
  introduced by this branch's own commit after #1202's fix landed) found and
  fixed in the same rebase.
- **8 code-review nits**: rebased onto #1202's fixes; also converted this
  branch's own `withClient(...)(ctx)` nested-invocation call site
  (`ctx.createApifyClient()` in `actors.cases.ts`) to the new direct
  `(client, ctx)` form.
- **execution-field assertion**: `registration.cases.ts`'s "execution field"
  case lost its transport-conditional check when `hasTasksSupport` was
  removed — the `2026-07-28` codec strips the field entirely, every other
  transport must still expose it. Restored the same branch, now keyed
  directly off `ctx.transport` instead of the removed field.

## Testing

`type-check`, `lint`, `format`, unit suite green after every rebase,
including live CI (Integration tests job).

## AI disclosure

Written with AI assistance (Poisson/Claude); reviewed and directed throughout.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -177,6 +177,15 @@ Integration tests run against two purpose-built Actors defined in [apify/mcp-ser
 
 ### Test organization across repos
 
+**Why `CaseCtx` couples the two repos.** MCP contract tests (protocol, tools, prompts,
+resources) have one correct behavior regardless of who's hosting the server, so they
+live here once and internal imports them — no second copy to keep in sync. Internal
+keeps only what's genuinely deployment-specific: databases, rate limiting, auth,
+multi-node coordination. `CaseCtx` is the seam that makes sharing possible without
+merging the two codebases — it's a non-standard cross-repo dependency injection
+(each side supplies its own client/environment for the same case logic), accepted
+deliberately for this trade-off.
+
 This package is also used by the hosted server in `apify-mcp-server-internal`. Every integration case is a `Case` object (`{ name, isDeploymentTest, run, ... }`) defined in `tests/test_kit/cases/*.cases.ts` — published behind the package's `./test-kit` export (`vitest` optional peerDependency). This repo's own `suite.ts` runs every case via `registerCases(name, allGroupCases, ctx)`; internal imports the same case arrays via `@apify/actors-mcp-server/test-kit` and calls `registerCases(name, allCases, { ...ctx, isDeploymentTestOnly: true })` against its own live staging/prod deploy — cases with `isDeploymentTest: false` register as `it.skip` there, so nothing needs re-registering by hand.
 
 Cases that share expensive setup (one seeded Actor run, say) do it via `ctx.getFixture(fixture)` (`Fixture<T> = { key, setup(ctx) }`) instead of a vitest `beforeAll` — `setup` runs at most once per `registerCases` call (once per transport dimension), memoized by `fixture.key`, no matter how many or few of the cases sharing it actually run (e.g. under `isDeploymentTestOnly`). This is what lets `storage.cases.ts`'s 13 cases built on one seeded run stay ordinary, individually eligible-for-`isDeploymentTest` `Case` objects instead of a separate non-flattenable group. See `tests/test_kit/register.ts` and `storage.cases.ts`'s `normalModeRunFixture`.
--- a/src/apify_client.ts
+++ b/src/apify_client.ts
@@ -53,7 +53,8 @@ export class ApifyClient extends _ApifyClient {
         super({
             // token null case is handled, we can assert type here
             ...(clientOptions as ApifyClientOptions),
-            baseUrl: getApifyAPIBaseUrl(),
+            // Explicit baseUrl wins over env default.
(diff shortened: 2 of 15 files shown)


jirispilka (commented):
This PR will need a rebase after the test kit moves outside `src`, including updates to its imports and separate build.

I would also prefer one dependency on `CaseCtx` instead of two related getters:

```ts
createApifyClient: () => ApifyClient;
```

Public can create it from `APIFY_TOKEN` and `APIFY_API_BASE_URL`, while internal can use `getTestToken()` and `settings.apiBaseUrl`. This keeps the token and URL together and avoids adding another `CaseCtx` field for future `ApifyClient` options.

RobertCrupa (approved):
lgtm 🚀

PR #1235, opened 2026-08-11
feat: add actor tasks

Janjiran:
Adding new Actor Tasks tool category to `get`, `create`, `update` or `publish` them via:
`get-actor-task`, `create-actor-task`, `update-actor-task`, `publish-actor-task` and `unpublish-actor-task` tools.

- `create-actor-task`: save an Actor configuration as a named task.
- `get-actor-task`: read a task, its publication state and its public display configuration.
- `update-actor-task`: change a task's input, run options, or the display configuration of its landing page.
- `publish-actor-task`: publish the task on its public landing page.
- `unpublish-actor-task`: take it back down. The display configuration is kept, so it can be published again without re-entering anything.

Also bumping apify-client to newly released 2.25.0 (had to extend `minimumReleaseAgeExclude`) which includes ability to publish and unpublish a task.


evals/test_cases.json
```diff
@@ -591,6 +591,90 @@
...
+      "query": "Set the landing page title of my task insta-daily to 'Daily Instagram scraper'",
+      "expectedTools": ["update-actor-task"],
+      "reference": "The landing page title is publicConfig.seoTitle, written through update-actor-task. It does not publish the task."
+    },
+    {
+      "id": "publish-actor-task-1",
+      "category": "publish-actor-task",
+      "query": "Publish my task insta-daily",
+      "expectedTools": ["publish-actor-task"],
+      "reference": "It should call publish-actor-task with the task ID 'insta-daily'."
```
katzino: Suggestion: it should rather be task name `'insta-daily'` instead of task ID, no?

evals/test_cases.json
```diff
@@ -591,6 +591,90 @@
...
+      "id": "unpublish-actor-task-1",
+      "category": "unpublish-actor-task",
+      "query": "Unpublish my task insta-daily",
+      "expectedTools": ["unpublish-actor-task"],
+      "reference": "It should call unpublish-actor-task with the task ID 'insta-daily'."
+    },
+    {
+      "id": "unpublish-actor-task-2",
+      "category": "unpublish-actor-task",
+      "query": "Take my task insta-daily off its public page but keep its display settings",
```
katzino, line 883: Question: unpublish tool can never really do anything else rather than unsetting `publishedAt`, right?

evals/test_cases.json
katzino, line 1: Suggestion: Would be nice to also against error cases - task can't be published due to permission issues or invalid publication state, maybe?

src/tools/tasks/update_actor_task.ts
```diff
@@ -0,0 +1,84 @@
...
+        destructiveHint: false,
+        idempotentHint: true,
+        openWorldHint: false,
+    },
+    call: async (toolArgs: InternalToolArgs) => {
+        const { args, apifyClient: client } = toolArgs;
+        const { taskId, name, title, description, input, build, timeoutSecs, memoryMbytes, publicConfig } =
+            updateActorTaskArgs.parse(args);
+
+        const options = { build, timeoutSecs, memoryMbytes };
```
katzino: Question: When `build` arg is provided and later on passed down to update method will just completely override the current value or does it merge? Maybe, we will need to get the current value and merge it manually?

src/tools/tasks/get_actor_task.ts
```diff
@@ -0,0 +1,59 @@
...
+        destructiveHint: false,
+        idempotentHint: true,
+        openWorldHint: false,
+    },
+    call: async (toolArgs: InternalToolArgs) => {
+        const { args, apifyClient: client } = toolArgs;
+        const parsed = getActorTaskArgs.parse(args);
+        const task = await client.task(parsed.taskId).get();
+        if (!task) {
+            return respondOk(`Task ${parsed.taskId} was not found.`);
```
katzino: Suggestion: `respondUserError` should probably be returned instead (as used elsewhere on not found items)

src/tools/tasks/unpublish_actor_task.ts
```diff
@@ -0,0 +1,52 @@
...
+- user_input: Unpublish my task my-task
+- user_input: Unpublish task E2jjCZBezvAZnX8Rb`,
+    inputSchema: z.toJSONSchema(unpublishActorTaskArgs) as ToolInputSchema,
+    outputSchema: actorTaskOutputSchema,
+    ajvValidate: compileSchema(z.toJSONSchema(unpublishActorTaskArgs)),
+    annotations: {
+        title: 'Unpublish Actor task',
+        readOnlyHint: false,
+        destructiveHint: false,
+        idempotentHint: false,
```
katzino: Suggestion: `true` makes more sense to me since calling this on the same task shouldn't have an effect on second try, right?
```suggestion
        idempotentHint: true,
```

src/tools/tasks/update_actor_task.ts
```diff
@@ -0,0 +1,84 @@
...
+        const hasOptions = Object.values(options).some((value) => value !== undefined);
+
+        const update: TaskUpdateData = {
+            ...(name && { name }),
+            ...(title && { title }),
+            ...(description && { description }),
```
katzino: Suggestion: `description` is still optional so we should probably allow setting an empty string as way to unset it I think

src/tools/tasks/publish_actor_task.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * https://docs.apify.com/api/v2/actor-task-put
+ */
+export const publishActorTask: ToolEntry = Object.freeze({
+    type: TOOL_TYPE.INTERNAL,
+    name: HELPER_TOOLS.ACTOR_TASK_PUBLISH,
+    title: 'Publish Actor task',
+    description: `Publish a task on its public landing page.
+The task's Actor must be public and the task must have its public display configuration set up -
+at least \`publicConfig.inputSchemaFields\` and \`publicConfig.datasetView\`. If publishing fails
```
katzino: Suggestion: I would also mention 50 published tasks per Actor cap

tests/unit/tools.actor_task_crud.test.ts
```diff
@@ -0,0 +1,178 @@
...
+        const result = (await (getActorTask as HelperTool).call(
+            stubToolCallContext({ taskId: 'nope' }, apifyClient),
+        )) as TextToolResult;
+
+        expect(result.content[0].text).toContain('not found');
+    });
+});
+
+describe('create-actor-task', () => {
+    it('maps the flat run options into the options object', async () => {
```
katzino, line 116: Note: Check one of my first comments, if you go and fix it this test should also change to something similar to

```
it('merges run options into the stored ones instead of replacing them', async () => {
      const { apifyClient, calls } = mockTaskApiClient(
          mockTask({ options: { build: 'latest', timeoutSecs: 300, memoryMbytes: 2048 } }),
      );
      await (updateActorTask as HelperTool).call(
          stubToolCallContext({ taskId: 'task-1', build: 'beta' }, apifyClient),
      );

      expect(calls).toEqual([
          { fn: 'get' },
          { fn: 'update', payload: { options: { build: 'beta', timeoutSecs: 300, memoryMbytes: 2048 } } },
      ]);
  });
```

evals/test_cases.json
Janjiran: correct

src/tools/tasks/update_actor_task.ts
Janjiran: Good point! I did the merge manually as you suggested

katzino (approved):

jirispilka (commented):
I'm sorry, I won't be able to review this today, let me share AI feedback first. Sorry fro the AI slop 😊 
I'll do a proper pass tomorrow.

The previous blocker is fixed correctly. `get_actor_task.ts` now uses `respondUserError`, restoring `SOFT_FAIL` / `INVALID_INPUT` telemetry. Unpublish is also correctly marked idempotent.

I verified the update behavior against the backend:

- `options` is replaced as a whole. The read-modify-write in `update_actor_task.ts` is necessary.
- `publicConfig` is merged by the backend. It must not use the same client-side read-modify-write.

For example, if the stored options are:

```json
{
  "build": "latest",
  "timeoutSecs": 300,
  "memoryMbytes": 1024
}
```

and the user changes only `build`, sending this directly would lose the other settings:

```json
{
  "options": {
    "build": "beta"
  }
}
```

The tool correctly reads the stored options and sends:

```json
{
  "options": {
    "build": "beta",
    "timeoutSecs": 300,
    "memoryMbytes": 1024
  }
}
```

`publicConfig` behaves differently. Given:

```json
{
  "seoTitle": "Old title",
  "inputSchemaFields": ["query"],
  "datasetView": "overview"
}
```

this update is safe:

```json
{
  "publicConfig": {
    "seoTitle": "New title"
  }
}
```

The backend preserves `inputSchemaFields` and `datasetView`.

## What blocks now

The tools promise support for bare task names, but the API interprets an unqualified value as a task ID.

For example, the descriptions and examples encourage this call:

```json
{
  "taskId": "insta-daily"
}
```

That requests `/actor-tasks/insta-daily`. The backend treats `insta-daily` as an ID, so a task named `insta-daily` is not found.

Working forms are:

```json
{
  "taskId": "~insta-daily"
}
```

for the authenticated user, or:

```json
{
  "taskId": "janjiran/insta-daily"
}
```

The JS client converts the slash form to the API's `janjiran~insta-daily` form.

This affects the parameter descriptions or examples in all four tools:

- `get-actor-task`
- `update-actor-task`
- `publish-actor-task`
- `unpublish-actor-task`

Either normalize a bare task name to `~task-name` before calling the client, or stop claiming that bare task names are supported and update the examples. Normalization would preserve the intended user experience.

## Non-blocking follow-ups

- Integration coverage belongs in `tests/integration/suite.ts`, but the suite is being refactored. Please create and link a follow-up issue covering the full task lifecycle and partial `options` / `publicConfig` updates.
- Eval formatting, categorization, and the new recovery cases are being handled separately. They do not affect this verdict.
- An options update for a missing task currently performs GET → `undefined` → PUT, then returns the PUT error instead of a friendly not-found response.
- Empty `description` clears the field, while empty `title` is silently dropped.
- `create_actor_task.ts` still includes undefined keys inside `options`, while update now builds a sparse object.
- Removing `categorization` from writable input is defensible. Omitting it from the returned task configuration is unnecessary.

## What is good

The options preservation fix matches the real backend contract. Its unit test asserts both the initial GET and the merged PUT payload. The telemetry and idempotency fixes also reuse the existing project patterns cleanly.

src/tools/tasks/create_actor_task.ts
```diff
@@ -0,0 +1,90 @@
...
+    description: `Create a saved Actor task — a named, reusable Actor configuration (input plus run options).
+The public display configuration (\`publicConfig\`) can be set here or later with ${HELPER_TOOLS.ACTOR_TASK_UPDATE};
+either way the task is not published until you call ${HELPER_TOOLS.ACTOR_TASK_PUBLISH}.
+
+USAGE:
+- Use when the user wants to save an Actor configuration for repeated use.
+
+USAGE EXAMPLES:
+- user_input: Save this instagram-scraper config as a task called daily-posts
+- user_input: Create a task for apify/rag-web-browser with my query`,
```
jirispilka: nit: the first example is good! It says to save config for daily posts (=repetitive use), the second is not that good. I would remove it

src/tools/tasks/create_actor_task.ts
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * https://docs.apify.com/api/v2/actor-tasks-post
+ */
+export const createActorTask: ToolEntry = Object.freeze({
+    type: TOOL_TYPE.INTERNAL,
+    name: HELPER_TOOLS.ACTOR_TASK_CREATE,
+    title: 'Create Actor task',
+    description: `Create a saved Actor task — a named, reusable Actor configuration (input plus run options).
+The public display configuration (\`publicConfig\`) can be set here or later with ${HELPER_TOOLS.ACTOR_TASK_UPDATE};
```
jirispilka: Unfortunatelly, as of now, you can't assume that a tool will be avaiable in the description. Because user can create a configuration when only `create-task` will be included using `mcp.apify.com?tools=create-task`. I know, bad config, but still we should handle that. 

I had to change it for the other [tools as well](https://github.com/apify/apify-mcp-server/pull/1261/changes#diff-4187c2a2b9450b8bd772a51849d2ce624173b5218a2b2af8c68b03cae5eea0ffR66) 

What was added in that PR: `Use tool xyx if the tool is available in the session`. A better way would be to a) include all the tools by default or b) include it conditioanlly. But that's out of the scope for these PRs.

src/tools/structured_output_schemas.ts
```diff
@@ -352,6 +352,43 @@ export const reportProblemToolOutputSchema = {
     required: ['reported'],
 };
 
+/**
+ * Schema shared by every Actor task tool. Input values are omitted on purpose (they may hold
+ * secrets)
+ * `inputFields` lists the field names that `publicConfig.inputSchemaFields` can use.
+ */
```
jirispilka, line 357:
```suggestion
/**
 * Schema shared by every Actor task tool. Input values are omitted on purpose (they may hold secrets) `inputFields` lists the field names that `publicConfig.inputSchemaFields` can use.
 */
```

jirispilka (changes requested):
Appologizes for a delayed review.

It is mostly good! Thanks! 
You should explain the task to LLM the same way you would explain it to users. It is mostly there but it can be improved at several places. Like what is a taks, why to create it, why to publish it, what are the benefits. These kind of things should be explained better. You are almost there, you just need tighten the loose ends.

1. What you need to do is to explain LLM what is a tasks and what is a task purpose. Something like: `Actor tasks are saved, reusable configurations of Actors. A task stores the Actor input and run options, such as the build, memory, and timeout. Tasks are useful for repeated or scheduled jobs because users do not need to configure the Actor again for every run. Creating or updating a task does not make it public. Publishing creates a public landing page for a specific use case.` You should put the explanation into  `public-tasks-api/src/utils/server-instructions/index.ts` and also disolve it into specific tool descriptions.

1. FYI, we mostly use `actor` instead of `actorId` as a tool argument. Because it was either id or name. I've created an issue, we'll change it to `actorId` at other places too.

1. For the publish tasks. I would change the tool description to explain why the task should be public: "Publish a saved task as a public landing page for a specific Actor use case. The page shows what the task does, selected input values, and the expected output. Published tasks appear in the Actor’s Examples tab and can be discovered by users, search engines, and AI agents. This can help users understand and try the Actor and can increase its runs. Publish only tasks that represent a useful, reliable, and specific use case. Not every saved task needs to be public."

1. Integration tests for tasks are missing. I would prefer to have it, if possible. Perhaps not the publishing one but at least create/delete? You can create a follow up issue and handle it later

1. If you have a tool to create a task, we should have a tool to delete IMO?

1. To have it in at mcp.apify.com, we need to update the internal repository as well: https://github.com/apify/apify-mcp-server-internal

1.  We'll handle evals with @RobertCrupa on the Claude harness is merged. Don't worry about that.

src/tools/tasks/task_helpers.ts
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * The API reads an unqualified id as an ID, so a bare resource name has to be prefixed with `~` to be
+ * resolved against the authenticated user's own resources. Ids that already carry a username, in either
+ * the `username/name` or `username~name` format, and ids that are already IDs, are returned unchanged.
+ *
+ * Applies to both tasks and the Actor a task is created for, which the API resolves the same way.
+ */
+export function toSafeResourceId(idOrName: string): string {
+    const trimmed = idOrName.trim();
```
jirispilka:
```suggestion
    const trimmed = idOrName.trim().replace('/', '~');
```

We should fix this because otherwise it returns the input unchanged when it contains /. So apify/instagram-scraper still goes out with a slash.

src/tools/tasks/get_actor_task.ts
```diff
@@ -0,0 +1,64 @@
...
+        title: 'Get Actor task',
+        readOnlyHint: true,
+        destructiveHint: false,
+        idempotentHint: true,
+        openWorldHint: false,
+    },
+    call: async (toolArgs: InternalToolArgs) => {
+        const { args, apifyClient: client } = toolArgs;
+        const parsed = getActorTaskArgs.parse(args);
+        const task = await client.task(toSafeResourceId(parsed.taskId)).get();
```
jirispilka: I run the analysis using mcpc, Claude complains that you can recognize ID from name when name is 17 characters

Image

Claude:

In the failing row the task exists — the probe created it seconds earlier — but the tool told the API "find the task whose ID is zzmcpcprobeseven1", no such ID exists, 404, and the user is told their task doesn't exist.

The fix. The ambiguity is genuinely unresolvable by inspection, so stop trying to resolve it — fall back instead. Only the 17-alnum window is ambiguous at all, so the extra request happens almost never:

```ts
/** A bare 17-alnum string is shape-identical to an ID and to a legal task name. */
function isAmbiguous(idOrName: string): boolean {
    const t = idOrName.trim();
    return !t.includes('/') && !t.includes('~') && APIFY_ID_REGEX.test(t);
}

// get_actor_task.ts
const task =
    (await client.task(toSafeResourceId(id)).get()) ??
    (isAmbiguous(id) ? await client.task(`~${id.trim()}`).get() : undefined);
```
Correct in every case, one extra call only on an ambiguous miss. Then add the row the current test table is missing — ['a 17-character name', 'zzmcpcprobeseven1', '~zzmcpcprobeseven1'] — which fails today and would have caught this.

jirispilka (approved):
Thanks for the changes.

We are almost there. 
While testing, Claude has found one issue with the ID vs task name. I believe we should fix that

There was a couple of other tool description issues while testing using mcpc and I asked Claude to fix it right away in this PR: https://github.com/apify/apify-mcp-server/pull/1273. Please review it and merge at will. 

Pre-approving

PR #1238, opened 2026-08-12
feat(evals): Run workflow evals on the Claude Agent SDK

RobertCrupa:
Stacked on #1224. First of two: this one makes the agent under test the real Claude Code harness. #1239 traces its conversation to Langfuse.

## Why
We want to simulate a user but having the same environment as them during evals. This means using the same harness, system prompt and tools.

## What

The hand-rolled OpenRouter loop and MCP client are gone. Each case now runs Claude Code headlessly through the Claude Agent SDK, with the `claude_code` system-prompt and tool presets, driving its own freshly spawned Apify MCP server. 

The judge, dataset, scores, and run gate are unchanged

## Note for reviewers

The evaluator is run with `allowDangerouslySkipPermissions` (headless, never prompts; the SDK requires both) so run it with the MCP tools only to stop it from executing bash

## Verification

- `--id search-google-maps`: 1/1 passed, and a direct probe confirmed the agent calls the MCP `search-actors` tool (thinking, tool call, answer).
- `report-problem-on-tool-error`: injection confirmed. The agent receives the refusal with the real `report-problem` nudge, retries `call-actor`, then explains to the user instead of calling `report-problem`. The case fails, and that is a genuine eval signal rather than a harness bug.
- `type-check`, `lint`, `test:unit` (1280 pass / 1 skip), `format`, `check:agents` all green.


MQ37 (approved):
Approving. One issue worth addressing (before or after merge):

`sdk_conversation_adapter.ts` only handles the `error_max_turns` result subtype from the SDK. If the run instead fails mid-conversation for a different reason (e.g. `error_during_execution` — an Anthropic API error, context overflow, etc.), that failure is silently dropped. The partial conversation gets sent to the LLM judge exactly like a normal run, so the judge scores it PASS/FAIL as if the agent just answered badly — it has no way to know the run was actually cut short by an error. That failure then shows up in the run's pass/fail results looking like a real eval failure, with no error logged anywhere.

Verified this in a sandbox: an API-error run and a max-turns run produce identical `conversation.turns`, and `workflow_judge.ts` never even reads `completed`/`hitMaxTurns`.

This contradicts the harness's own stated design: harness errors should fail the run, not be scored as a failing eval. Recommend throwing on any non-success, non-max-turns result subtype instead.

RobertCrupa:
@MQ37 Thank you! good catch it is now fixed [a442572](https://github.com/apify/apify-mcp-server/pull/1238/commits/a442572eb404641072dbcf791eaf7a41c92ddd18)

jirispilka (approved):
It is mostly good but, I believe, you've chosen a wrong transformation, which makes the code more complicated than it should be.

I'm thinking aloud here, we plan to move these evals to Actor so I don't think it is worth to change it :). I hope that we will remove this sooner than later. I also hope that Arto is not going to build on this. So please perhaps just check what it would take to re do, if u agree.

My notes below:

I would rather change judge than introduce conversation adapter. You need to rebuild `ConversationHistory` .
 
`formatConversationForJudge` reads four things: `userPrompt`, `turns[].toolCalls[].name`/`.arguments`, and `turns[].finalResponse`. `langfuse_experiment.ts` reads `conversation.totalTokens`. Nothing else in the repo reads a conversation. The turn grouping the judge needs is one blank line, so a formatter over `SDKMessage[]` produces the same prompt with no intermediate object.

**Pros**

- No conversion on the judge path. The judge reads the SDK's own types. And these types are pretty much a standard for other provides too!
- Five types deleted: `ConversationHistory`, `ConversationTurn`, `McpToolCall`, `McpTool`, `McpToolResult`.
- The 4-clause terminal-turn guard disappears. "Is the final answer already in the stream" becomes one condition.
- What is left of the adapter drops from 223 lines to 101, and it stops being an adapter: usage totals, the transcript, and the error-subtype gate, which the stream genuinely does not give you for free. Across `evals/workflows`, 732 lines becomes 609.
- The judge prompt is unchanged, so this is a refactor and not a behavior change. 
- 
**Cons**

- It touches the judge and its test, which this PR set out not to touch.
- #1239 needs rework, more than a rebase. It picks up `toolInvocations`, `metrics`, and `claudeCodeVersion` along with the `receivedAt` threading, and `langfuse_observations.ts` reads the final answer from the `result` message instead of `turns.at(-1)?.finalResponse`.

RobertCrupa:
@jirispilka Thank you for the comments. I have simplified the data structure.

When we do the Actor implementation, we will change the judge also.

PR #1239, opened 2026-08-12
feat(evals): Trace the agent conversation to Langfuse

RobertCrupa:
Stacked on #1238. Second of two: #1238 makes the agent under test the real Claude Code harness, this one puts its conversation on the Langfuse trace.

## What

- `langfuse_observations.ts` builds the tree from the adapted SDK stream and emits it.
- Tool spans are timed from when the SDK delivered the call and its result.
- `llm_client.ts` wraps each call in a generation, which is what puts the judge call, its verdict, and its tokens on the trace.
- A failed tool call is raised to `ERROR` level with the payload as the status, and a run that never reached a final answer is flagged `WARNING`, so both are findable in the UI.
- The tree is emitted before the judge call, so a failing judge still leaves the conversation on the trace to debug. A crashed agent run leaves no spans.
- Adds `@opentelemetry/api`, the peer dependency of `@langfuse/tracing`, for the `SpanContext` type.

## Verification

- `--id search-google-maps`: 1/1 passed, the five observations land with the right nesting, and `GET /api/public/v2/metrics` reports 71.5k tokens and $0.039 on the agent generation plus 798 tokens on the judge one.
- `type-check`, `lint`, `test:unit` (1289 pass / 1 skip), `format`, `check:agents` all green.

### DIFF
--- a/evals/workflows/README.md
+++ b/evals/workflows/README.md
@@ -170,7 +170,7 @@ Separation allows independent optimization for speed vs evaluation quality.
 - `ConversationHistory` carries only what the judge and the scores read; tool results and metrics live on `ToolInvocation` and `ConversationMetrics`
 - MCP tool names are stripped of their `mcp__apify__` prefix, so the judge sees `search-actors` as before
 - Subagent messages (via the `Task` tool) are excluded, so the transcript reflects the main agent
-- Cached prompt tokens are counted into `total_tokens`; the API reports them separately and a cached run would otherwise look nearly free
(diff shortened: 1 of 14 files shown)


MQ37 (approved):
LGTM.

🔴 `emitObservations` errors are caught by `makeTask`'s generic try/catch and treated as harness failures, dropping an otherwise-passing item. Suggest wrapping the emit call in its own try/catch (log + continue), same pattern already used for flush/shutdown.

🟡 `usageNode` uses three non-null assertions (`metrics.promptTokens!` etc). Destructuring first would let TS narrow without `!`.

🟡 `emitObservations`'s `asType` dispatch is a nested ternary. An `if/else` reads the same and matches the "no nested ternaries" convention.

RobertCrupa:
This PR has once again grown too big; I will split it further

jirispilka (approved):
Nice! 

Let's not changes this PR, I created a follow up issue [here](https://github.com/apify/apify-mcp-server/issues/1286)

Issue #1253, opened 2026-08-13
[Bug]: Enum truncation makes the validator reject valid Actor input, and the agent cannot recover (follow-up to #287)


ondraulehla:
### Server type

Hosted (mcp.apify.com)

### MCP Client

Other: raw JSON-RPC 2.0 probes over streamable HTTP (no MCP client app), so the reproduction below is exact requests.

### Operating System

macOS

### What happened?

The server shortens large enums for display, but it also compiles its AJV validator from the same shortened schema. The display cap silently becomes a validation cap: a value the Actor accepts is rejected at the MCP layer before the run starts. It happens every time.

Concrete case. `compass/crawler-google-places` declares `categoryFilterWords` with 4036 allowed values (71,122 characters in total). `filterAndShortenEnum` caps the enum at `ACTOR_ENUM_MAX_LENGTH` (2000 characters), which keeps the first 237 values. The last kept value is `observatory`. Common values such as `restaurant` and `zoo` are dropped, and calls that use them fail validation.

The agent also has no way out. The error message tells the agent to check the schema with `fetch-actor-details`, but that response does not contain the enum for `categoryFilterWords` at all. The description there only links to an external URL with the category list. So the agent bounces between a validator that rejects a valid value and a details tool that cannot show the full list. If it retries with one of the 237 listed values, it silently narrows the user's search in a way the user did not ask for.

Where this comes from, as far as I can tell:

- `src/tools/actor_input_schema.ts` (`filterAndShortenEnum`, called from `shortenProperties`) shortens the enum for display.
- `src/tools/actors/actor_tools_factory.ts` compiles the AJV gate from that same shortened schema object.
- `src/mcp/tool_call_engine.ts` rejects the call before dispatch.

Why I think this is a bug and not a display trade-off:

- #287 reported this failure class ("Users can't run Actors if enum is truncated"). PR #292 raised the cap from 200 to 2000 characters. That fixed geocodes, but any enum over 2000 characters still hits the original problem. This Actor's enum is 35 times the cap.
- `src/utils/ajv.ts` already removes `pattern` and `format` from validation, with the rationale that this layer only sanitizes LLM args and the Actor re-validates its real input on the run. The same rationale applies to shortened enums: not enforcing them loses no protection, because the platform validates the full schema at run start (step 5 below shows it does).
- The act-validate-input endpoint from #736 would also validate against the full schema, at the cost of one round-trip.

[...]

PR #1256, opened 2026-08-14
feat: Validate Actor input against the platform before every run

### LINKED ISSUE #736: Validate input schema before calling an Actor (especially for task)
Use this endpoint: https://docs.apify.com/api/v2/act-validate-input-post

## Where we stand

**Validation already runs before anything starts.** `call-actor` fetches the Actor definition and validates input at `src/tools/actors/call_actor.ts:463`; the run starts at :596. Every `tools/call` is gated by per-tool AJV at `src/mcp/tool_call_engine.ts:215`, before MCP task creation. No saved-task calling exists in the server.

**But AJV validates a derived schema, not the Actor's real one.** `buildActorInputSchema` (`src/tools/actor_input_schema.ts:164`) deletes `$ref`/`$schema` (AJV can't resolve them), transforms properties, and runs with `removeAdditional` so unknown props are silently dropped. The endpoint would validate the real schema.

Cost of switching: one round-trip per call, no client method for it, and MCP-only fields (`waitSecs`) must be stripped before sending or we'd reject valid input.

**Decision point is telemetry:** how many MCP-started runs fail on input errors *after* passing the AJV gate? Each one is a run the user paid for that was doomed at start. Common → build it. Rare → close this.

MQ37:
## What
Adds a remote check via Apify's `validate-input` endpoint right before an Actor run starts: `call-actor`/`call-actor-widget` (shared `resolveAndValidateActor`) and direct per-Actor tools (`actor_executor.ts`). Local AJV validation is unchanged and still runs first — this is an added second gate, not a replacement.

## Why
The server's AJV gate validates a derived/shortened schema copy, not the Actor's real one — enum truncation in particular lets it pass input the Actor would reject at runtime, after a billable run already started (#736). The remote check asks the platform directly. Fails open on anything but a clean 400 (network error, timeout, 5xx, unexpected shape): the Actor's own run-time validation is the real backstop, so an unavailable best-effort check must never block a run that would otherwise have started.

Closes #736.

## Testing
`pnpm run type-check/lint/format/check:agents` clean. `pnpm run test:unit`: 1290 passed, 1 skipped, zero regressions — covers the validator (200/400/fail-open), `resolveAndValidateActor`, and `actorExecutor`.

Also tested live end-to-end (built `dist/stdio.js`, real token, via mcpc): `apify/instagram-scraper`'s `directUrls` has a `pattern` the local AJV strips — `call-actor` with a non-matching URL was rejected before any run started with the platform's own message; a valid URL still started a run normally.

### DIFF
--- a/src/errors.ts
+++ b/src/errors.ts
@@ -15,6 +15,15 @@ export class SchemaTooLargeError extends Error {
     }
 }
 
+/**
+ * Thrown when Apify's `validate-input` endpoint confirms Actor input is invalid against its real
+ * schema. `statusCode` lets the existing HTTP-based classifiers treat it as a 4xx user error.
+ */
+export class ActorInputValidationError extends Error {
+    override readonly name = 'ActorInputValidationError';
+    readonly statusCode = 400;
+}
+
 export const ACTOR_LOAD_ERROR_KIND = {
     NOT_FOUND: 'not-found',
     LOAD_FAILED: 'load-failed',
--- a/src/mcp/tool_dispatch.ts
+++ b/src/mcp/tool_dispatch.ts
@@ -241,6 +241,7 @@ export async function dispatchToolCall(params: {
                 });
                 const executorResult = await actorExecutor.executeActorTool({
                     actorFullName: tool.actorFullName,
+                    actorId: tool.actorId,
                     input: toolArgs,
                     apifyClient,
                     callOptions: { memory: tool.memoryMbytes },
--- a/src/tools/actors/actor_executor.ts
+++ b/src/tools/actors/actor_executor.ts
@@ -3,6 +3,7 @@ import log from '@apify/log';
 import type { ActorExecutionParams, ActorExecutionResult, ActorExecutor } from '../../types.js';
 import { getConsoleLinkContext } from '../../utils/console_link.js';
 import { redactSkyfirePayId } from '../../utils/logging.js';
+import { validateActorInputRemotely } from '../../utils/validate_actor_input.js';
 import { buildGetActorRunResponse } from '../runs/get_actor_run.js';
 import { abortRunOnSignal, CALL_ACTOR_WAIT_SECS_DEFAULT, fetchActorRunData } from './actor_run_response.js';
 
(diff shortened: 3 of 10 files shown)


jirispilka (commented):
@MQ37 this one is my fault, not yours. In the referenced issue I left note: "Decision point is telemetry: how many MCP-started runs fail on input errors after passing the AJV gate? Each one is a run the user paid for that was doomed at start. Common → build it. Rare → close this."

I should have run it before creating the issue

## The failures are real

65,832 platform input rejections in three months, 1.32% of `call-actor` calls, 36,341 unique users — 17.9% of everyone who calls an Actor. So the instinct that something here needed fixing was right.

## But this check can't reduce them

Both paths hit the same platform validator, so the only thing a preflight can change is *when* you hear the verdict. `cheap_scraper/linkedin-job-scraper` declares `maxItems` with `minimum: 150`, and the agent sends `100`

Invalid input never costs the user anything. There's no harm for the preflight to prevent. 

## The two places timing could have mattered

- **#1253** — can't help. Local AJV rejects before the remote check is reached, as the code stands, so the truncated-enum false rejection never gets that far. #1258 is the fix for that one.
- **Tasks** — can't help either, and this is the part I had most wrong in #736. The AJV gate runs in `prepareToolCall()`, *before* `taskStore.createTask()`. This check runs inside `dispatchToolCall()`, *after* the task exists — so the client gets a task handle and then an async failure either way. Tasks were the weakest case for the idea, not the strongest.

## Where it led instead

This PR is what pushed me to go get the numbers, and they turned up the actual cause: `filterSchemaProperties()` drops every value constraint in the input schema spec, so the agent never sees `minimum: 150` and our AJV can't enforce it either — it's compiled from the same stripped schema. 96% of the rejections above name a keyword we strip. `fetch-actor-details` uses the same function, so the agent's recovery path is blind too.

That's #1266. Much smaller diff, and it removes the failures rather than relabelling them — including giving tasks the synchronous rejection this PR was reaching for, since AJV already runs before `createTask()`.

One piece of this PR worth keeping regardless: your input-error response shape. Today a `start()` invalid-input 400 falls through to *"Please verify the Actor name, input parameters, and ensure the Actor exists"*, which points the agent at the wrong problem entirely. Branching that path to return the input schema plus the platform's validation message — no network call — would be a genuinely good small PR on its own.

We can close this one.

MQ37:
Closing per review — the remote pre-flight check can't reduce failures since both paths hit the same platform validator. Split out the worthwhile piece (input-error response shape) as #1275.

Issue #637, opened 2026-04-02
RAG Web Browser: query input reported as optional; tool field ordering incorrect


jirispilka:
Two issues with the RAG Web Browser tool schema:

### 1. `query` field is not marked as required

The `query` input for `apify/rag-web-browser` is reported as optional in the MCP tool schema, even though it is mandatory for meaningful results.

**Root cause:** `pruneSchemaPropertiesByWhitelist()` in `src/tools/utils.ts` only preserves required fields that already exist in the upstream Actor's `required` array. If the upstream schema omits `query` from `required`, it gets dropped.

**Fix:** After pruning, explicitly ensure `query` is in the `required` array for the RAG Web Browser tool.

### 2. `annotations`/`title`/`description` should appear before `icons`

In `src/tools/core/actor_tools_factory.ts`, the tool object has `_meta` and `icons` before `annotations` (which contains `title`). The `title` and `description` fields should be placed before `icons` for correct ordering.

## Files involved

- `src/tools/utils.ts` — `buildActorInputSchema()`, `pruneSchemaPropertiesByWhitelist()`
- `src/tools/core/actor_tools_factory.ts` — tool object construction
- `src/utils/tools.ts` — `getToolPublicFieldOnly()`

PR #1257, opened 2026-08-14
fix: Stop filterSchemaProperties from writing phantom default/prefill keys

### LINKED ISSUE #1266: [Bug]: Input-schema value constraints are stripped from Actor tool schemas
## Problem

`filterSchemaProperties()` in `src/tools/actor_input_schema.ts` builds the schema used by Actor tools and `fetch-actor-details`. For each top-level input property, it keeps only: `title`, `description`, `enum`, `type`, `default`, `prefill`, `properties`, `items`, and `required`.

It drops top-level constraints such as `minimum`, `maximum`, `minLength`, `maxLength`, `pattern`, `minItems`, `maxItems`, and `uniqueItems`.

The same derived schema is shown to the agent and compiled into local AJV. The agent cannot see these constraints, and the server does not reject values that violate them. The platform validates the original Actor schema before creating a run, so the call fails without a billable run.

`fetch-actor-details` applies the same transformation, so it cannot show the missing constraint after a failure.

## Evidence

Segment/Mixpanel recorded 65,832 platform input-validation failures from Actor tool calls between 2026-05-17 and 2026-08-17 (~730/day). In the last 30 days, 96.1% of messages matched a keyword this transformation removes.

[Reports](https://mixpanel.com/project/2860297/view/3393129/app/insights#rQfmAH2FKU8C)

`cheap_scraper/linkedin-job-scraper` is the largest source: 9,468 failures in one month. Its `maxItems` field declares `minimum: 150`; the platform rejects lower values, but that constraint is absent from the MCP schema.

## Affected

`call-actor`, `call-actor-widget`, direct Actor tools, and `fetch-actor-details`.

## Expected

Expose the applicable top-level JSON Schema constraints and enforce the same schema locally, so agents can form valid input and receive local recovery errors when they cannot.

Telemetry context: #736.

MQ37:
## What
`filterSchemaProperties()` unconditionally wrote every whitelisted key onto each output property, including `default: undefined`/`prefill: undefined` on properties that never declared them upstream. Only copy a key when the upstream value is actually defined. Reverts `fixZodSchemaRequired()`'s value-check workaround back to the cleaner key-presence check now that the corruption it worked around is gone.

## Why
The phantom keys previously broke `fixZodSchemaRequired()`'s original key-presence check (#637), patched there instead of at the source. This is the minimal phase-1 fix from #675 — stop the corruption itself. Phase 2/3 (restructuring the wider transform pipeline) are separate, out of scope here.

Closes https://github.com/apify/apify-mcp-server/issues/1266.

## Testing
`pnpm run type-check/lint/test:unit/format/check:agents` all clean. `pnpm run test:unit`: 1282 passed, 1 skipped. Updated the assertion that previously expected the phantom keys to instead expect exactly the real upstream keys (`description`, `title`, `type`) for the test fixture. Two regression tests for #637 (one in `tools.mode_contract.test.ts`, one found in `utils.ajv.test.ts` not originally flagged) hand-constructed the exact phantom pattern (`default: undefined`) as their setup — updated both to reflect that this shape no longer occurs from real upstream data (JSON can't carry a literal `undefined`, and `filterSchemaProperties` no longer emits it), since a key-presence check correctly treats an explicit `default: undefined` as present. Independently verified by a fresh-context subagent (re-read the diff, re-ran the full verification suite, traced the test fixture through the whole transform pipeline by hand to confirm the new assertion is exactly right): PASS.

### DIFF
--- a/src/tools/actor_input_schema.ts
+++ b/src/tools/actor_input_schema.ts
@@ -93,18 +93,17 @@ export function filterSchemaProperties(properties: { [key: string]: SchemaProper
 } {
     const filteredProperties: { [key: string]: SchemaProperties } = {};
     for (const [key, property] of Object.entries(properties)) {
-        const filtered: SchemaProperties = {
+        filteredProperties[key] = {
             title: property.title,
             description: property.description,
             type: property.type,
+            ...(property.enum !== undefined && { enum: property.enum }),
+            ...(property.default !== undefined && { default: property.default }),
+            ...(property.prefill !== undefined && { prefill: property.prefill }),
+            ...(property.properties !== undefined && { properties: property.properties }),
+            ...(property.items !== undefined && { items: property.items }),
+            ...(property.required !== undefined && { required: property.required }),
         };
-        if (property.enum !== undefined) filtered.enum = property.enum;
-        if (property.default !== undefined) filtered.default = property.default;
-        if (property.prefill !== undefined) filtered.prefill = property.prefill;
-        if (property.properties !== undefined) filtered.properties = property.properties;
-        if (property.items !== undefined) filtered.items = property.items;
-        if (property.required !== undefined) filtered.required = property.required;
-        filteredProperties[key] = filtered;
     }
     return filteredProperties;
 }
--- a/src/tools/actor_input_schema.ts
+++ b/src/tools/actor_input_schema.ts
@@ -83,11 +83,8 @@ export function buildApifySpecificProperties(
  * Filters schema properties to include only the necessary fields.
  * This is done to reduce the size of the input schema and to make it more readable.
  *
- * TODO(#675): This object literal unconditionally assigns every whitelisted key,
- * including `default: undefined`, on properties that didn't declare them upstream.
- * This creates phantom keys that broke `fixZodSchemaRequired()` via key-presence
- * checks (#637). The symptom is patched in `src/utils/ajv.ts` (value-check), but
- * this function should only emit keys whose upstream value is not `undefined`.
+ * Only copies a whitelisted key when the upstream property actually declares it, so
+ * unset fields (e.g. no `default`) aren't turned into phantom `default: undefined` keys.
  *
  * @param properties
  */
@@ -96,17 +93,18 @@ export function filterSchemaProperties(properties: { [key: string]: SchemaProper
 } {
     const filteredProperties: { [key: string]: SchemaProperties } = {};
     for (const [key, property] of Object.entries(properties)) {
-        filteredProperties[key] = {
+        const filtered: SchemaProperties = {
             title: property.title,
             description: property.description,
-            enum: property.enum,
             type: property.type,
-            default: property.default,
-            prefill: property.prefill,
-            properties: property.properties,
-            items: property.items,
-            required: property.required,
         };
+        if (property.enum !== undefined) filtered.enum = property.enum;
+        if (property.default !== undefined) filtered.default = property.default;
+        if (property.prefill !== undefined) filtered.prefill = property.prefill;
+        if (property.properties !== undefined) filtered.properties = property.properties;
+        if (property.items !== undefined) filtered.items = property.items;
+        if (property.required !== undefined) filtered.required = property.required;
+        filteredProperties[key] = filtered;
     }
     return filteredProperties;
 }
--- a/src/utils/ajv.ts
+++ b/src/utils/ajv.ts
@@ -20,9 +20,6 @@ ajv.removeKeyword('format');
  * has the same issue: it lists `.default()` fields as required and emits `$schema` that
  * breaks AJV compilation.
  *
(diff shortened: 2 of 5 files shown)


src/tools/actor_input_schema.ts
```diff
@@ -96,17 +93,18 @@ export function filterSchemaProperties(properties: { [key: string]: SchemaProper
 } {
     const filteredProperties: { [key: string]: SchemaProperties } = {};
     for (const [key, property] of Object.entries(properties)) {
-        filteredProperties[key] = {
+        const filtered: SchemaProperties = {
```
jirispilka: IMHO, this would be nicer:

```ts
     filteredProperties[key] = {
         title: property.title,
         type: property.type,
         ...(property.enum !== undefined && { enum: property.enum }),
         ...(property.default !== undefined && { default: property.default }),
         ...(property.prefill !== undefined && { prefill: property.prefill }),
         ...(property.properties !== undefined && { properties: property.properties }),
         ...(property.items !== undefined && { items: property.items }),
         ...(property.required !== undefined && { required: property.required }),
     };
```

tests/unit/tools.utils.test.ts
jirispilka, line 1: Extend the local type and make sure prefill survive the transformation

```ts
  properties?: Record<string, {
      description?: string;
      prefill?: unknown;
  }>;
```
  Then add after the required assertion:
```ts
  expect(schema.properties?.query?.prefill).toBe(
      'web browser for RAG pipelines',
  );
````

 This reuses the existing fixture and verifies prefill survives the entire transformation pipeline.

File: src/utils/ajv.ts [hunk 2]
```diff
@@ -34,8 +31,7 @@ export function fixZodSchemaRequired(schema: Record<string, unknown>): Record<st
         cleaned.required = (cleaned.required as string[]).filter((fieldName) => {
             const fieldSchema = properties[fieldName];
             if (typeof fieldSchema !== 'object' || fieldSchema === null) return true;
-            // Value-check (NOT `'default' in fieldSchema`) — see docstring for why.
-            return (fieldSchema as { default?: unknown }).default === undefined;
+            return !('default' in fieldSchema);
```
jirispilka:
```suggestion
            return (fieldSchema as { default?: unknown }).default === undefined;
```

Could we keep the defensive value-check here?

return (fieldSchema as { default?: unknown }).default === undefined;

 The source fix still removes phantom keys, while this protects future hand-built schemas containing default: undefined. It also matches apify-core and preserves the #637 regression tests. The key-presence check adds no functional benefit for valid JSON schemas.

jirispilka (changes requested):
Looks ok, but the `default: undefined` is still not valid value, apify-core restrict it in that way.

 These two tests duplicate the same default: undefined behavior. Please keep only one tests

We need restore the behavior:

  - tests/unit/utils.ajv.test.ts:67

    it('keeps required fields whose `default` is explicitly undefined', () => {
        // existing setup
        expect(result.required).toEqual(['query']);
    });

  - tests/unit/tools.mode_contract.test.ts:383

    it('preserves required fields when `default` is explicitly undefined', () => {
        // existing setup
        expect(schema.required).toEqual(['query']);
    });

src/tools/actor_input_schema.ts
MQ37: Done — switched to conditional spread.

tests/unit/tools.utils.test.ts
MQ37: Done — added the `prefill` assertion to the full-pipeline test, reusing the existing rag-web-browser fixture as suggested.

src/utils/ajv.ts
MQ37: Agreed, restored the value-check. Confirmed apify-core's equivalent (`getAjvValidator`, `input_schema.both.ts:192`) also computes `hasDefault` via `default !== undefined` — key-presence here would've made this repo disagree with the platform on whether a field with a literal `default: undefined` is required. Not just theoretical: zod's `toJSONSchema()` can't produce that shape (it throws on `.default(undefined)`), so the only route to it is a hand-built schema or a future call site doing something like `{ ...x, default: input.default }` — the same class of mistake `filterSchemaProperties()` made. Root-cause fix stands; guard stays.

MQ37:
Restored `utils.ajv.test.ts`'s regression test to its original name/assertion (`required` keeps `query` when `default` is explicitly `undefined`). Dropped the duplicate in `tools.mode_contract.test.ts` rather than restoring both — same invariant one layer up, and the adjacent `should preserve required fields from Apify Actor-shape inputSchemas` test in that file already covers `getToolPublicFieldOnly` for the real-default case, so no coverage is lost. Shout if you'd rather keep an integration-level copy of the undefined-default case too.

Pushed in 6bf090a. `type-check`/`lint`/`test:unit` (93 files, 1281 passed, 1 skipped)/`format`/`check:agents` all clean.

jirispilka (approved):

PR #1258, opened 2026-08-14
fix: Stop enforcing enum values dropped by input-schema truncation

### LINKED ISSUE #1253: [Bug]: Enum truncation makes the validator reject valid Actor input, and the agent cannot recover (follow-up to #287)

MQ37:
## What
Reworked per review: instead of a separate AJV-only validation schema, the shared `shortenProperties()` now handles an oversized `enum`/`items.enum` itself — kept in full when it fits `ACTOR_ENUM_MAX_LENGTH` (no duplicate "Possible values" text, the `enum` field already carries it), dropped entirely (not partially truncated) when it doesn't, with a note + a few example values appended to the description instead. One schema, no divergence between what the LLM sees and what AJV validates, across all three consumers that share this path: direct Actor tools, `call-actor`, and `fetch-actor-details`. When the session's `tools/list` happens to include `fetch-actor-details`, the dropped-enum note additionally points to it for more context on the Actor — via a new per-session `buildInputSchema(ctx)` render on `ToolBase` (mirrors the existing `buildDescription(ctx)` convention), since a bare tool name can't be hardcoded into Actor input-schema text without knowing whether that tool is actually loaded in a given session.

## Why
Closes #1253. Reviewer (jirispilka) asked for the enum keyword to be removed entirely rather than partially truncated, folded into the one shared path so display and validation never diverge, and covering all three consumers instead of two. AJV still enforces `type`, just not enum membership, on a dropped field — the real Actor/platform holds the true, complete list and is left to reject an actually-invalid value on its own, confirmed live end-to-end (see Testing).

## Testing
`pnpm run type-check/lint/test:unit/format/check:agents` all clean — 101 test files, 1581 passed, 1 skipped. New/updated unit coverage: all-or-nothing `shortenProperties`/`filterAndShortenEnum` (kept-whole vs. dropped-whole, both `enum` and `items.enum`), `findDroppedEnumProperties`, the per-session `buildInputSchema` gating (`tools.actor_tools_factory.enum_gating.test.ts`), and a Skyfire-payment composition regression (`payments.skyfire.build_input_schema.test.ts`) — a payment decorator that clones a tool must wrap `buildInputSchema`, not only the static `inputSchema`, or an injected field silently disappears whenever the gated render is used; caught this live before it shipped.

Live e2e against the real, currently-deployed `compass/crawler-google-places` Actor (Google Maps Scraper) — the actual motivating case, a real 4,036-value / 71KB `categoryFilterWords` enum: confirmed via real `tools/list` that the enum is dropped with a generic note, additionally naming `fetch-actor-details` only in a session where it's actually loaded; confirmed `fetch-actor-details` itself doesn't expose the complete enum either (same drop, worded accordingly rather than overclaiming); confirmed a genuinely invalid category value is rejected by the real platform pre-run at zero cost; and confirmed `categoryFilterWords: ["restaurant"]` — a real category that was in the exact range the old code falsely rejected — is now accepted, a real run starts, and the filter actually works end to end (1 place scraped before the run was aborted).

`getToolPublicFieldOnly` (exported via `internals.js`) gained a new field-dependent branch (`buildInputSchema`) — additive, same safe-fallback shape as the existing `buildDescription`, but flagging per this repo's own norm on changes to what the hosted server consumes.

Written with AI assistance (Claude).


jirispilka (commented):
Thank you for this PR. When we discussed the fix, I understood that the `enum` keyword would be removed completely when the full list does not fit. I still think that is the correct fix because an incomplete enum is not a valid representation of the Actor's input.

What do you think? The enum truncation was good fix for large schema. But, still,  like your idea with removing it completely and do not pretend it is an enumeration of all values (this is what LLM understands imo). 

The current PR:

- Keeps a truncated enum in the public schema.
- Creates a separate private schema without that enum for AJV.
- Fixes false rejection in direct tools and `call-actor`.
- Does not fix `fetch-actor-details`.
- Still tells the model that the incomplete list is exhaustive.
- Adds a helper, clone, comparison logic, factory wiring, and two conflicting schema representations.

Please handle this in the shared `shortenProperties()` path instead:

- Keep the enum when every non-empty value fits.
- When it does not fit, remove the `enum` keyword.
- Add examples and a note that the complete list is not shown.

This gives us one honest schema for display and validation, and it covers direct tools, `call-actor`, and `fetch-actor-details`.

Removing `enum` does not remove type validation. AJV will still enforce constraints such as `type: "array"` and `items.type: "string"`. It will reject a wrong container type or non-string items, but it will not check membership in the omitted enum. The Actor receives the value and may reject unsupported values according to its own validation or runtime behavior.

That is the intended trade-off: the MCP server must not reject a valid Actor input based on a list it knows is incomplete.

MQ37:
Reworked per your feedback — folded into `shortenProperties()`, one schema for display and AJV, no separate validation-only copy. An enum that fits stays whole; one that doesn't gets dropped entirely (not truncated), with a note and a few examples in the description instead. Covers `fetch-actor-details` too now, for free, since it shares the same function.

Verified live against `compass/crawler-google-places`'s real 4,036-value `categoryFilterWords`: a bogus category is rejected by the platform pre-run at zero cost, and `restaurant` (the exact value class the original bug falsely rejected) is now accepted and actually runs.

One addition beyond your ask: when `fetch-actor-details` happens to be loaded in a session, the dropped-enum note also points to it — gated per-session so we never hardcode a tool name that might not be loaded.

src/tools/actor_input_schema.ts
```diff
@@ -291,30 +280,34 @@ export function addEnumsToDescriptionsWithExamples(
...
+/** Note for a dropped enum — the field still accepts these values, just too many to list. */
+function buildEnumDroppedNote(rawValues: string[]): string {
+    const examples = rawValues
+        .filter((value) => value !== '')
+        .slice(0, ENUM_DROPPED_NOTE_EXAMPLE_COUNT)
+        .map((value) =>
+            value.length > ENUM_DROPPED_NOTE_EXAMPLE_MAX_LENGTH
+                ? `${value.slice(0, ENUM_DROPPED_NOTE_EXAMPLE_MAX_LENGTH)}...`
+                : value,
+        );
```
jirispilka: If an actual value is longer than 60 characters, the code shortens it:

  Actual valid value:
  very-long-category-name-that-continues-for-many-more-characters

  Displayed example:
  very-long-category-name-that-continues-for-many-more-charac...

  The displayed value with ... is not a valid enum value. An AI may copy that example, pass MCP validation, and then have the Actor reject it.

Do we need to shorten it? 
Instead of three dots: 1) remove the value (preffered) or 2) say it very-long-category-name-that-continues-for-many-more-charac[truncated]?

For the 2) we show values to the agent that the agent can't use anyway. Therefore I prefer 1)

src/tools/actor_input_schema.ts
```diff
@@ -291,30 +280,34 @@ export function addEnumsToDescriptionsWithExamples(
...
-        if (enumValue === '') return false;
-        charCount += enumValue.length;
-        return charCount <= ACTOR_ENUM_MAX_LENGTH;
-    });
-
-    return resultEnumList.length > 0 ? resultEnumList : undefined;
+    const nonEmpty = enumList.filter((value) => value !== '');
+    if (nonEmpty.length === 0) return undefined;
+    const charCount = nonEmpty.reduce((sum, value) => sum + value.length, 0);
+    return charCount <= ACTOR_ENUM_MAX_LENGTH ? nonEmpty : undefined;
```
jirispilka: `filterAndShortenEnum` is now the wrong name: it never shortens an enum.

I’d make it private and rename it:

`function getEnumIfFits(enumValues: string[]): string[] | undefined`

It should remove blank values, return the complete remaining enum when it fits, and return undefined when it does not. shortenProperties() should then actually delete an oversized enum.

jirispilka (approved):
Fixed in the stacked PR: https://github.com/apify/apify-mcp-server/pull/1342

Issue #1260, opened 2026-08-15
feat: Serve all tools by default, drop per-session description hedging


jirispilka:
## Problem

Tool descriptions name sibling tools, but the served set varies per session: defaults are the `actors` and `docs` categories plus `AUTO_INJECTED_TOOLS`, and a `tools=` selector can produce almost any subset. A description naming a tool the client never received in `tools/list` invites a call to a tool that does not exist.

Today that is patched in prose — each cross-tool reference carries an "…if that tool is available in this session" hedge. About 20 of them across 14 files in four phrasings, held in place by a lint test that recomputes the served set for every mode and selector.

The hedging is the symptom. The variable default tool set is the cause.

## Evidence

- Hedges are hand-maintained. A new description, or a new selector combination, silently needs another one; the lint test only catches it after the fact.
- It degrades the common case to protect the rare one. Default sessions do serve the named tool, and now read a hedged instruction instead of a direct one.
- The session-accurate mechanism already exists for runtime text: `loadedToolNames` comes from the live registry at call time and `suggestTool` gates `nextStep` on it. Descriptions are static module constants, so they cannot use it.
- Whether the narrow default helps is an empirical question we can now answer. `evals/test_cases.json` measures tool selection against `expectedTools` across models, but tool-set size is not a dimension — nothing compares selection accuracy with all categories loaded against the current default.

## Affected

`toolCategoriesEnabledByDefault` (`src/tools/registry.ts`), `AUTO_INJECTED_TOOLS` (`src/utils/tools_loader.ts`), descriptions across `src/tools/`, `src/tools/AGENTS.md`, the README default-tools table, `tests/unit/tools.mode_contract.test.ts`, and the hosted server's default surface (needs coordination).

Blocked on: eval coverage that varies tool-set size.

Issue #1227, opened 2026-08-10
chore: Align Actor mode vocabulary in model-visible tool descriptions


jirispilka:
Part of the Actor mode vocabulary alignment.

Align the model-visible strings. Separate from the prose and identifier pass because these are read by the LLM, so changing them can shift tool-selection behavior and will break assertions that match on the wording.

## Strings

- `src/tools/actors/call_actor.ts` — `CALL_ACTOR_MCP_MISSING_TOOL_NAME_MSG`, `CALL_ACTOR_MCP_SERVER_SECTION`, and the `actor` input property description
- `src/tools/actors/fetch_actor_details.ts` — the `mcpTools` field description, and the "Common patterns" line that says "MCP Actors"
- `src/utils/server-instructions/index.ts` — the `call-actor` guidance line

## Verification

Unit tests are not enough. Needs an e2e run plus the conformance suites, since e2e cases and conformance expectations match on these descriptions.

PR #1261, opened 2026-08-15
fix: Correct tool descriptions naming absent tools or wrong output

jirispilka:
Tool descriptions named sibling tools that a session may never have served, and several stated outputs the code does not produce.

Cross-tool references now render per session instead of being hedged in prose: a tool with such a reference defines `buildDescription({ hasTool })`, and `getToolPublicFieldOnly` renders it against the exact set each `tools/list` response returns (`presentTools`, passed by both era handlers). `description` keeps the all-tools-present render for consumers without a session set. The mode-contract test renders every session shape (default, each category, every single-tool selector, mixed combos, both modes) the same way the handlers do and fails on any absent-tool name or hedge text. Implements #1265.

Output claims corrected against the code that produces them:

- Slack: https://apify.slack.com/archives/C08FDFSTH50/p1786625241747719
- apify/apify-ms-copilot-plugin#2

| Tool | Was | Now |
|---|---|---|
| `abort-actor-run` | no effect for `FINISHED, FAILED, ABORTING, TIMED-OUT` | `SUCCEEDED, FAILED, ABORTING, ABORTED, TIMED-OUT` — `FINISHED` is not an Apify run status, and `TIMING-OUT` is abortable |
| `get-actor-run-list` | `datasetId` / `keyValueStoreId` | `defaultDatasetId` / `defaultKeyValueStoreId` |
| `search-actors` | card list named a success-rate stat that is never rendered | matches `formatActorToActorCard` field for field |
| `call-actor` | always "returns … field metadata" | dataset field metadata only for `waitSecs > 0`; the `waitSecs: 0` path returns id-only storages |
| server instructions | a capped wait "returns the result" | returns current status and storage IDs, not output rows |

Overlaps #1227, which edits the call-actor MCP section in the same file (now `buildCallActorMcpServerSection`); that one will need a rebase, not a conflict resolution.

Follow-up: #1272 — the task tool descriptions from #1235 need the same gating once both land.

@RobertCrupa @MQ37 once we have good evals, we should enable all tools and get rid of this spaghetti tool definitions 

#1260 — serving all tools by default stays open as the default-surface question; a `tools=` selector still produces arbitrary subsets, which the per-session rendering now handles.

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -93,7 +93,7 @@ export const APIFY_AI_CLIENT_NAME = 'apify-console-ai-chat';
 export const RAG_WEB_BROWSER = 'apify/rag-web-browser';
 export const RAG_WEB_BROWSER_WHITELISTED_FIELDS = ['query', 'maxResults', 'outputFormats'];
 export const RAG_WEB_BROWSER_ADDITIONAL_DESC = `Use this tool when user wants to GET or RETRIEVE actual data immediately (one-time data retrieval).
-This tool directly fetches and returns data - it does NOT just find tools.
+This tool scrapes the data itself - it does NOT just find tools.
 
 Examples of when to use:
 - User wants current/immediate data (e.g., "Get flight prices for tomorrow", "What's the weather today?")
--- a/src/mcp/legacy_server.ts
+++ b/src/mcp/legacy_server.ts
@@ -362,10 +362,12 @@ export class LegacyMcpServer {
     private setupToolHandlers(): void {
         // Handles the request to list tools.
         this.server.setRequestHandler(ListToolsRequestSchema, async () => {
+            const presentTools = new Set(this.host.tools.keys());
             const tools = Array.from(this.host.tools.values()).map((tool) =>
                 getToolPublicFieldOnly(tool, {
                     mode: this.host.serverMode,
                     filterWidgetMeta: true,
+                    presentTools,
                 }),
             );
             return { tools };
--- a/src/mcp/stateless_server.ts
+++ b/src/mcp/stateless_server.ts
@@ -164,8 +164,9 @@ class StatelessMcpServer {
     private setupToolHandlers(): void {
         this.server.setRequestHandler('tools/list', async (_request, ctx) => {
             const snapshot = await this.resolveSnapshot(ctx);
+            const presentTools = new Set(snapshot.tools.keys());
             const tools = Array.from(snapshot.tools.values()).map((tool) =>
-                getToolPublicFieldOnly(tool, { mode: snapshot.serverMode, filterWidgetMeta: true }),
+                getToolPublicFieldOnly(tool, { mode: snapshot.serverMode, filterWidgetMeta: true, presentTools }),
             );
             // Tool entries carry the same public fields as the SDK's `Tool`; type-boundary cast only.
             return { tools } as unknown as ListToolsResult;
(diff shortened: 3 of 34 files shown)


RobertCrupa (approved):
LGTM

MQ37:
Cool, but I would honestly do this properly. I don't like that we're including instructions that might not be relevant for the session — e.g. the store-search mention in `WIDGET_ADDENDUM` (`call_actor.ts`), hedged with "...if that tool is available in this session." Naming a tool that may not exist and hoping the model self-censors on a hedge clause is a bad pattern: it confuses the model and wastes tokens on every session, including the common case where the tool is obviously present, just to cover the rare one where it isn't. This isn't isolated to that one spot either — same hedge shows up ~20 times across 14 files in this diff.

Proposal: gate this on the actual session/client context instead of prose. `composeToolsForClient` already resolves the final tool set and client identity per session — build a small context (`{ serverMode, clientContext, hasTool(name) }`) right after that resolution, and let descriptions that reference another tool call `hasTool(...)` to include or drop the clause, instead of hedging it. Bigger change than this PR, so I created #1265 to track it, so we don't lose it.

MQ37 (commented):
⏫

jirispilka:
> Proposal: gate this on the actual session/client context instead of prose. `composeToolsForClient` already resolves the final tool set and client identity per session — build a small context (`{ serverMode, clientContext, hasTool(name) }`) right after that resolution, and let descriptions that reference another tool call `hasTool(...)` to include or drop the clause, instead of hedging it. Bigger change than this PR, so I created #1265 to track it, so we don't lose it.

You are right, this is better than to populate tool description with "if the tool is available in this session".
But I don't like the solution to be honest.

Perhaps we can "solve" it that some of the tools will be always enabled. Bad that's bad UX imo.

Implemented, with one correction on where the gating runs.

Descriptions with cross-tool references now define `buildDescription({ hasTool })` on the tool entry; `description` keeps the all-tools-present render for consumers without a session set.

The correction: the context can't be built right after `composeToolsForClient` — that resolution is per-source, and the client-visible set is the union across `toolSources`, so a `hasTool` built there under-reports. Rendering instead happens once at the tools/list boundary: `getToolPublicFieldOnly` takes a `presentTools` set and both era handlers pass the exact collection being returned — correct for any `tools=` combination by construction. `serverMode`/`clientContext` stayed out of the context: nothing needs them yet.

The mode-contract test now renders every session shape (default, categories, every single-tool selector, mixed combos, both modes) the same way the handlers do, and fails on any absent-tool name or leftover hedge text.

#1265 can close as addressed.

MQ37:
`buildCallActorErrorResponse` (call_actor.ts) still hedged with the old prose ("is available in this session") for store-search/actor-get-details — the one instance e456327c didn't convert to `hasTool`-style gating, since it's runtime error text, not a `tools/list` description.

Fixed in e27884b3: gates the recovery hint on `toolArgs.loadedToolNames` (already threaded into every internal tool call, same session tool set `tools/list` renders against) — names only tools actually loaded, drops the sentence when neither is. Verified: tsc clean, lint/format clean, full unit suite 1352/1352 (incl. 2 new gating tests).

MQ37 (approved):
Gating implemented correctly: session-scoped `buildDescription`/`hasTool` at the tools/list boundary, both handlers pass exact `presentTools`. Verified full unit suite + tsc + lint on branch tip. One missed hedge (runtime error text in `buildCallActorErrorResponse`) found and fixed in e27884b3.

PR #1273, opened 2026-08-18
fix: Surface publicConfig rejections with a recovery path

jirispilka:
## Why

Found while reviewing #1235. The API validates a task's `publicConfig` on **every write**, not only on publish, and rejects it with a typed error:

```
type: "cannot-publish-actor-task"  statusCode: 400
message: Cannot publish Actor task: Dataset view doesn't exist
```

That reached the client through the generic `Error calling tool "..."` wrapper with no telemetry — and `extractToolTelemetry` (`src/utils/tool_status.ts:212-219`) maps an `isError` response with no telemetry to `INTERNAL_ERROR`, so a wrong view key was logged as a server fault.

"Verify the input parameters" is also not actionable here: `datasetView` is required to publish and is not discoverable through any tool in this server. 

## Proof it works

mcpc against the real API, rebuilt server, `tasks` category enabled:

```
$ mcpc @stdio-full tools-call create-actor-task actorId:='"20F7qU3yp7DaV9ugE"' \
    name:='"zz-probe"' publicConfig:='{"datasetView":"definitely-not-a-view"}'
{
  "content": [
    { "text": "Cannot publish Actor task: Dataset view doesn't exist" },
    { "text": "`publicConfig` is validated on write, not at publish. Ask the user for the
               `datasetView` key — no tool here can list them." }
  ],
  "isError": true
}
```

Same on `update-actor-task`, and for `inputSchemaFields` — where the API answers `Cannot publish Actor task: Select at least one input field`, confirming an unknown field name counts as zero valid fields.

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -220,6 +220,7 @@ export const FAILURE_CATEGORY = {
 export const APIFY_ERROR_TYPE_FULL_PERMISSION_NOT_APPROVED = 'full-permission-actor-not-approved';
 export const APIFY_ERROR_TYPE_MEMORY_LIMIT_EXCEEDED = 'memory-limit-exceeded';
 export const APIFY_ERROR_TYPE_CANNOT_START_ACTOR_RUNS = 'cannot-start-actor-runs';
+export const APIFY_ERROR_TYPE_CANNOT_PUBLISH_ACTOR_TASK = 'cannot-publish-actor-task';
 
 // HTTP status codes
 export const HTTP_UNAUTHORIZED = 401;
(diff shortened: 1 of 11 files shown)


Janjiran (commented):
Thank you for noticing and creating a PR for that! I have just 2 small related things to consider:

1. We should also add the `if (isCannotPublishTaskError(error)) return respondPublicConfigRejected(error);` to publishing since that's just `update({ isPublic: true })`
but 2. more importantly, we can just extend `buildToolCallErrorResult()` in `src/mcp/tool_call_error_mapper.ts:57` with `isCannotPublishTaskError` and all tools get it automatically and we can remove individual cases, wdyt?

jirispilka:
> Thank you for noticing and creating a PR for that! I have just 2 small related things to consider:
> 
> 1. We should also add the `if (isCannotPublishTaskError(error)) return respondPublicConfigRejected(error);` to publishing since that's just `update({ isPublic: true })`
>    but 2. more importantly, we can just extend `buildToolCallErrorResult()` in `src/mcp/tool_call_error_mapper.ts:57` with `isCannotPublishTaskError` and all tools get it automatically and we can remove individual cases, wdyt?

Good catch, fixed by using `tool_call_error_mapper`

Janjiran (approved):

PR #1275, opened 2026-08-18
feat: return input schema on start() invalid-input errors

MQ37:
## What
When `actor.start()` is rejected by the platform with a confirmed `invalid-input` 400 (real-schema validation, not our own AJV copy), `call-actor`/`call-actor-widget` **and direct per-Actor tools** now return the Actor's input schema and the platform's own message instead of falling through to the generic "verify the Actor name, input parameters, and ensure the Actor exists" text. No extra network call — the schema was already fetched during `resolveAndValidateActor` (or already known at tools/list time for direct actor tools).

## Why
Split out from #1256 per [review](https://github.com/apify/apify-mcp-server/pull/1256#pullrequestreview-4951041353): a remote pre-flight validate-input call can't reduce failures (same validator either way), but the existing fallback message on a real start()-time input rejection points the agent at the wrong problem. This fixes just that response shape.

Direct actor tools (`actor_executor.ts`) hit the same platform rejection via their own `actor.start()` call but had no interception at all — extended after the initial `call-actor` fix landed. Kept as independent inline logic in each of the two call sites rather than a shared helper: measured a shared module at +17 net lines to guard a duplication that's structurally capped at exactly these two places (the only two spots in this codebase that `start()` an Actor against user-supplied input).

## Testing
`pnpm run type-check/lint/test:unit/format/check:agents` clean (93 files, 1361 passed, 1 skipped, zero regressions). Unit tests cover: schema+message returned (not the generic fallback) when `inputSchema` is available, the platform message alone when it isn't, the direct-actor-tool executor delegating instead of throwing, any other `start()` error rethrown unchanged, and the `isActorInputValidationError` predicate matching only the confirmed error type (not the sibling `invalid-input-schema`).


src/utils/apify_errors.ts
```diff
@@ -26,6 +29,23 @@ export function isCannotPublishTaskError(error: unknown): error is ApifyApiError
...
+
+/** Texts for a confirmed platform input-validation error — same shape wherever start() can throw one. */
+export function buildInvalidInputTexts(actorName: string, errMsg: string, inputSchema?: ToolInputSchema): string[] {
+    return [
+        `Failed to call Actor '${actorName}': ${errMsg}`,
+        `Please ensure the input is correct and matches the Actor's input schema.`,
+        ...(inputSchema ? [`Input schema:\n${wrapJsonText(inputSchema)}`] : []),
+    ];
+}
+
```
jirispilka: Move buildInvalidInputTexts out of apify_errors.ts. The module's own header says "Kept in one leaf module (imports only const + apify-client) so logging, telemetry, payments, and the tool layer can share them without import cycles." The PR adds import { wrapJsonText } from './encode_text.js' (runtime) and import type { ToolInputSchema } from '../types.js', so that comment is now false — and a text formatter isn't an error-classification predicate anyway.

Move it to src/utils/mcp.ts, next to respondUserError and getToolCallErrorUserText. That file already imports both ./apify_errors.js (line 5) and wrapJsonText (line 6), so it's zero new imports and no cycle. Move its test from tests/unit/utils.apify_errors.test.ts to tests/unit/utils.mcp.test.ts; the isActorInputValidationError predicate test stays put.

src/mcp/tool_dispatch.ts
```diff
@@ -241,6 +241,8 @@ export async function dispatchToolCall(params: {
                 });
                 const executorResult = await actorExecutor.executeActorTool({
                     actorFullName: tool.actorFullName,
+                    actorId: tool.actorId,
+                    inputSchema: tool.inputSchema,
```
jirispilka: For the direct Actor calls, the schema is returned but the LLM already has schema in the context. This will only duplicate it.

Drop the schema echo on the direct-actor-tool path. getToolPublicFieldOnly (src/utils/tools.ts:113) serves tool.inputSchema verbatim in tools/list, and MCP clients re-send tool definitions every turn — so on that path the echoed schema is guaranteed to already be in the agent's context at the moment it reads the error. It's duplicate tokens, not redundancy-as-insurance.

src/mcp/tool_dispatch.ts:244-245 — drop inputSchema: tool.inputSchema, keep actorId (telemetry needs it).
src/types.ts — drop inputSchema from ActorExecutionParams, keep actorId.
src/tools/actors/actor_executor.ts — call buildInvalidInputTexts(actorFullName, error.message) with no schema.

The direct path keeps both real wins: no more uncaught throw, plus the platform's own message. This also retires the reviewer's dead-branch question, and shrinks the diff.

call-actor/call-actor-widget keep the echo — there the schema genuinely isn't in context, since the tool takes an Actor name and a free-form input object.

src/types.ts
```diff
@@ -473,6 +473,10 @@ export type ServerModeOption = SERVER_MODE | 'auto';
 export type ActorExecutionParams = {
     /** Full name of the Actor (e.g., "apify/rag-web-browser") */
     actorFullName: string;
+    /** Actor's platform ID — echoed into telemetry on a platform input-validation error. */
+    actorId?: string;
+    /** Actor's input schema — echoed back verbatim on a platform input-validation error. */
+    inputSchema?: ToolInputSchema;
```
jirispilka: If we drop schema that is returned for direct tool calls:

Fix the two doc comments that overstate what's returned. src/types.ts says the schema is "echoed back verbatim" and the PR body says "the Actor's input schema".

jirispilka (approved):
Nice, thanks! 

I made the two comments inline that I would like to have fixed.
Pre-approving.

One more thing: I re-ran the telemetry with Claude (details in #1266, reopened — #1257 closed it without the fix landing). 94.8% of platform schema rejections name a constraint that filterSchemaProperties strips, so the echoed schema mostly won't show why the input was rejected until #1266 lands. The platform's message is what carries the information here — still a clear improvement over "verify the Actor name…". FWIW: call-actor is 89% of these rejections, direct tools 5.4%.

Things I checked and won't nitpick: the as ActorExecutionResult cast (existing idiom there), the predicate scoping vs invalid-input-schema (matches API docs, tested), response size (same schema we already serve in tools/list).

MQ37:
`filterSchemaProperties` exists because early MCP clients broke on JSON Schema keywords they didn't expect — we pared the exposed schema down to a safe subset to stop that. That's why `minimum`/`maximum`/`pattern`/etc. never made it into the tool schema in the first place.

Ran this end-to-end against `cheap_scraper/linkedin-job-scraper` (real minimum 150, sent `maxItems: 1`):

```
Failed to call Actor 'cheap_scraper/linkedin-job-scraper': Input is not valid: Field input.maxItems must be >= 150
```

Agent gets the real rejection reason — it just comes from the platform's error message, not the echoed schema. Schema's `maxItems` is still just `{"type":"integer"}`; the `>= 150` never appears in it.

Widening `filterSchemaProperties` to carry those constraints is real scope, not this PR: needs manual verification against the major MCP clients (the exact thing this filtering was added to protect against) plus making the added keywords fully JSON-Schema-compatible. Tracking that as #1266, separate from this fix.

src/utils/apify_errors.ts
MQ37: Moved to `src/utils/mcp.ts`, next to `respondUserError`/`getToolCallErrorUserText`. `apify_errors.ts` back to pure predicates, leaf-module comment true again.

src/mcp/tool_dispatch.ts
MQ37: Dropped for direct actor tools — `tool_dispatch.ts` no longer passes `inputSchema`, `actor_executor.ts` calls `buildInvalidInputTexts` with no schema arg. `call-actor`/`call-actor-widget` keep it — that tool's own `tools/list` schema is generic (`actor: string`, `input: object`), so the target Actor's specific schema was never in the agent's context to begin with; dropping the echo there would leave zero schema info, not duplicate info.

src/types.ts
MQ37: Fixed — `ActorExecutionParams.inputSchema` field removed from `types.ts` along with its doc comment, since it's no longer echoed on that path.

Issue #1229, opened 2026-08-10
feat: Evaluate and enable apify/web-fetch as a default tool


jirispilka:
## Problem

`apify/web-fetch` launches Aug 31 and should work zero-config in Apify MCP. Blocked by #1208 (run/standby decision table). A default MCP-server Actor must not put a live `tools/list` probe on every session boot (stdio) or every stateless request (hosted) — defaults need no-probe loading.

## Proposed solution

1. **No-probe loading for default MCP-server Actors**: pinned tool snapshot (name, description, input schema, server URL) registered at boot; connection happens only at call time (dispatch already reconnects per call). CI drift test compares live `listTools()` against the snapshot so tool changes in the Actor become coordinated releases.
2. **Evals**: add cases to `evals/test_cases.json` — fetch-style queries must pick the web-fetch tool, search-style queries must stay on `apify/rag-web-browser`. Steer the overlap via tool descriptions (same mechanism as `RAG_WEB_BROWSER_ADDITIONAL_DESC`).
3. **Flip**: add `apify/web-fetch` to `defaults.actors` once evals pass.

## Plan

- [ ] Pinned tool snapshot for default MCP-server Actors + CI drift test (live `listTools` vs snapshot)
- [ ] web-fetch vs rag-web-browser steering in tool descriptions
- [ ] Eval cases (fetch vs search routing) in `evals/test_cases.json`; run evals, fix steering until green
- [ ] Add `apify/web-fetch` to `defaults.actors`
- [ ] Verify token-less and payment-provider (x402/Skyfire) sessions skip the default cleanly — standby Actors are excluded there, so the default surface differs by design

Blocked by #1208.

jirispilka:
This has been shipped

Issue #857, opened 2026-05-18
fix: Allow calling MCP server actors in normal (one-shot) mode


RobertCrupa:
## Problem

When an actor is also an MCP server, `call-actor` blocks any call that does not include a tool name — even when the intent is to run it in normal (one-shot) mode with plain actor input.

The guard that caused this:

```ts
if (isActorMcpServer && isMcpToolNameInvalid) {
     return {
         earlyResponse: buildMCPResponse({
             texts: [CALL_ACTOR_MCP_MISSING_TOOL_NAME_MSG],
             isError: true,
         }),
     };
 }
```

This block was added to steer LLMs toward the `actorName:toolName` syntax when calling MCP server actors. The side-effect is that it also prevents legitimate one-shot calls (e.g. `apify/example-mcp-server` with `{"firstNumber": 4, "secondNumber": 5}`), which should execute the actor normally and return a dataset result.

The block has been commented out to restore one-shot mode support.

## Follow-up

Add evals that verify whether the original guard is still necessary — i.e. whether LLMs reliably use `actor:tool` syntax for MCP server actors without the explicit error nudge.

jirispilka:
@RobertCrupa Good description! Thanks.

jirispilka:
won't-fix: run mode stays unreachable for MCP-server Actors, per https://github.com/apify/apify-mcp-server/issues/1208.

Issue #668, opened 2026-04-13
Reduce Sentry noise: normalize install-corruption / stale-Node / EPIPE events


jirispilka:
## Problem

The `apify-mcp` Sentry project has 88 open issues generating ~160K+ events/month. Analysis shows ~99% are NOT bugs in our code:

| Root cause | Issues | Events/month |
|---|---|---|
| Corrupt `npx` / `bunx` install cache (e.g. `APIFY-MCP-6` alone: 108,987 events for missing `@segment/analytics-node/app/analytics-node`) | ~60 | ~115K |
| Stale Node.js < 20 — undici `File`/`ReadableStream` not defined, `util.getSystemErrorMap` missing (`APIFY-MCP-1F`, `-2`, `-1E`, `-9`, `-16`, `-2R`) | 6 | ~47K |
| stdio `EPIPE` — MCP client closed the pipe mid-write (`APIFY-MCP-3`, `-4`, `-28`, `-34`, `-35`) | 5 | ~820 |
| Ajv / ajv-formats version mismatch in user npx cache (`APIFY-MCP-1G`, `-2D`, `-5`) | 3 | ~185 |

These drown out real bugs. Dropping them blinds us to whether install-corruption mitigations (future bundling, docs) actually help, so we want to **collapse, not filter**.

## Solution

Two-part, scope-minimal change:

### 1. Sentry `beforeSend` normalizer (`src/sentry_normalize.ts`, new)

Pure function `normalizeSentryEvent(event, hint)` that re-fingerprints the three noise categories into ONE Sentry issue each, preserving event counts for trending. Adds a `category` tag. Downgrades level (`warning` for install/node, `info` for EPIPE). Unclassified events pass through unchanged so real bugs still surface.

### 2. Startup self-diagnose (`src/instrument.ts`, modified)

`process.on('uncaughtException', …)` handler, installed right after `Sentry.init`. Classifies the error via `classifyStartupError(err)`. If it matches install-corruption or stale-Node, prints a friendly actionable message to stderr (cache paths for macOS/Linux/Windows; Node upgrade link) and emits `Sentry.captureMessage('install-corruption-diagnosed', …)` so we can track how many users actually see the diagnose message (vs. just crash). Unmatched errors fall through to Sentry's default handler unchanged.

## Design notes

- Co-locate both pure functions (`normalizeSentryEvent`, `classifyStartupError`) in `src/sentry_normalize.ts` — trivially unit-testable without mocking Sentry.
[...]

PR #1278, opened 2026-08-19
feat: Run/standby mode decision table for Actor tool loading

### LINKED ISSUE #1208: feat: Run/standby mode decision table for Actor tool loading
## Problem

`apify/web-fetch` (standby Actor exposing an MCP server) must become usable through Apify MCP, and the server needs one consistent rule for exposing Actors across run mode, standby mode, and `webServerMcpPath`. Decided in [Slack](https://apify.slack.com/archives/C0BHH3EPEBC/p1785996363244139) + meeting 2026-08-10; spec: [Notion — MCP and standby Actors](https://app.notion.com/p/3b8f39950a2280949acfc3ffa3196883).

## Proposed solution

Implement this decision table in tool loading and `call-actor`:

| | Run mode only | Standby mode only | Both run and standby |
|---|---|---|---|
| **has `webServerMcpPath`** | Ignore `webServerMcpPath`; `call-actor "actor:tool"` rejected | Use standby only with MCP: tools via live probe; `call-actor "actor:tool"` → MCP call; bare `call-actor` → error "specify tool name" | Use standby only with MCP — run-mode tool not loaded; `call-actor "actor:tool"` → MCP call; bare `call-actor` → error "specify tool name" |
| **no `webServerMcpPath`** | Use run only | Ignore standby: not loaded, explicit `ActorLoadError` (OpenAPI-based support later) | Use run only |

Column detection: **run only** = standby not enabled · **standby only** = standby enabled + empty input schema · **both** = standby enabled + non-empty input schema. With `webServerMcpPath` present the two standby columns behave identically, so the schema check only decides the bottom row.

Definitions and affected paths:

- *Empty input schema* = the default build's definition has no `input`, or `input.properties` is empty. The check applies only when standby is enabled — run-only Actors without an input schema keep loading as today.
- `fetch-actor-details` resolves the Actor's MCP URL through the same helper and must follow the table: a run-only Actor with a leftover `webServerMcpPath` reports "not an MCP server" in its mcpTools output.
- Bulk session-boot loading drops per-Actor load errors, so an unsupported standby-only Actor in URL params silently doesn't appear in the tool list; the new `ActorLoadError` surfaces only through `call-actor`.

## Plan

- [ ] `resolveActorToolMode(actorInfo): 'run' | 'mcp' | 'unsupported'` + `supportsRunMode` (empty-input-schema check) in `src/tools/actor_tool_naming.ts`
- [ ] `getActorsAsTools`: route per mode; new `ActorLoadError` kind `STANDBY_WITHOUT_MCP_NOT_SUPPORTED` for the unsupported cell
- [ ] `getActorMcpUrlCached`: require `actorStandby.isEnabled` (top-left cell — today the code attempts an MCP handshake against non-standby Actors)
- [ ] In-process tests for all 6 cells: real `ActorsMcpServer` (`withServer`/`withStatelessServer` helpers), Actor definitions mocked, remote MCP server run in-process over the SDK's `InMemoryTransport` — assert `tools/list` and `call-actor` behavior per cell
- [ ] Live integration in `tests/integration/suite.ts` for the two existing fixtures: `apify/normal-mode-test-actor` (run only), `apify/example-mcp-server` (standby MCP)
- [ ] Follow-up: standby-only-without-MCP fixture in apify/mcp-server-test-actor for live coverage of `STANDBY_WITHOUT_MCP_NOT_SUPPORTED`
- [ ] Follow-up: evals + `apify/web-fetch` as a default tool → #1229
- [ ] Follow-up: tool-list caching (Redis for hosted; API endpoint for stdio local server)

Note: per this decision, run mode stays unreachable for MCP-server Actors (`call-actor` keeps the "specify tool name" guard) — #857 should be closed as won't-fix.

Related: #1277 — MCP-proxied tool naming (same loading path, separate PR).

## Alternatives considered

- Hardcoded web-fetch tool starting a run-mode (batch) run per fetch (original scope of this issue) — dropped: container start per call vs one HTTP request to the warm standby pool.
- Loading both run-mode tool and MCP tools for dual-mode Actors — considered, reverted: one representation per Actor keeps the tool surface unambiguous; revisit if telemetry shows run-mode demand for these Actors.

jirispilka:
## What this solves

Closes #1208.

There was no single rule for how an Actor becomes a tool. Run mode, standby mode and
`webServerMcpPath` were each read in a different place

https://claude.ai/code/artifact/93083980-de27-4f74-b73b-467af97ef67a?org=e9c9b830-2329-42ca-bf01-09917f01fde5

## Before → after, per cell

| standby | `webServerMcpPath` | input schema | before | after |
|---|---|---|---|---|
| disabled | absent | any | run tool | unchanged |
| disabled | **present (leftover)** | any | **wrongly treated as an MCP server** | run tool, path ignored |
| enabled | present | any | proxied MCP tools | unchanged |
| enabled | absent | non-empty | run tool | unchanged |
| enabled | absent | **empty** | **unusable run tool** | no tool; clear error |

**Cell 2 — run-only Actor with a leftover `webServerMcpPath`** (the MCP-URL resolution checked the
path but never `actorStandby.isEnabled`):

```text
before  call-actor "acme/stale-mcp-path"        → demands a tool name instead of running it
        call-actor "acme/stale-mcp-path:tool"   → Failed to connect to MCP server for Actor 'acme/stale-mcp-path'.
after   call-actor "acme/stale-mcp-path"        → runs the Actor
```

**Cell 5 — standby-only Actor, no MCP server, empty input schema** (nothing to run, nothing to
proxy — the old run tool couldn't do anything, and the two `call-actor` shapes invented two
different reasons):

```text
before  call-actor "acme/standby-only"          → Failed to call Actor '…': User was not found or
                                                  authentication token is not valid
        call-actor "acme/standby-only:tool"     → Actor 'acme/standby-only' is not an MCP server.
after   both shapes                             → Actor "acme/standby-only" runs in standby mode without
                                                  an MCP server and has an empty input schema, which is
                                                  not supported yet.
```

## Telemetry change — flagging for the hosted server

Three caller-mistake rejections move from `FAILED` / `INTERNAL_ERROR` (+ error log) to
`SOFT_FAIL` / `INVALID_INPUT` (+ softFail): both standby kinds, and `actor:toolName` aimed at a
non-MCP Actor. None is a server fault, so none should raise an error-level alert.

## Known and accepted (for your agent)

- Replacing the old `string | false` MCP-URL return with a resolved discriminated union
  (`ActorToolResolutionResult`) rides along with the feature, against "refactoring is a separate
  PR". The feature needs the caller to distinguish "run tool" from "unsupported cell"; a separate
  landing would have meant a throwaway tri-state in between. `handleMcpToolCall` itself takes a
  plain `mcpServerUrl: string` — the caller owns the mode decision, so the handler never
  re-derives it.
- The `STANDBY_PAYMENT_NOT_SUPPORTED` arm in `buildCallActorErrorResponse` is currently unreachable
  in production (the pre-flight guard intercepts first). Kept as a tested defensive fallback.
- Standby Actors with an empty input schema drop out of `tools/list` silently (bulk loading discards
  per-Actor errors, with a server-side softFail trace); the reason reaches users only via
  `call-actor`. Same as the existing `NOT_FOUND` behavior.
- An Actor declaring a `webServerMcpPath` that escapes its standby origin still reaches
  `getActorMCPServerURL` unguarded in the resolver. Pre-existing — master's `getActorMcpUrlCached`
  was equally unguarded — and narrowed here, since only MCP-mode Actors reach that call now.
  Tracked in #1321, together with the missing `actorId` telemetry on the early-return rejections.

## Follow-ups

- #1229 (`apify/web-fetch` as a default tool) is unblocked by this PR.
- #857 should be closed as won't-fix: run mode stays unreachable for MCP-server Actors, per #1208.
- A live fixture for the standby-without-MCP cell, and tool-list caching, remain tracked in #1208.
- `CALL_ACTOR_MCP_MISSING_TOOL_NAME_MSG` still answers via `respondServerError`, so the exact
  mistake that message exists to teach records `INTERNAL_ERROR`. Same class as the telemetry change
  above, left out to keep this PR to one thing — fits #668.
- `getMCPServersAsTools` keeps an `if (!actorInfo.webServerMcpPath)` guard that
  `ACTOR_TOOL_MODE.MCP` now makes unreachable.

### DIFF
--- a/src/tools/actors/call_actor.ts
+++ b/src/tools/actors/call_actor.ts
@@ -18,14 +18,7 @@ import { ACTOR_LOAD_ERROR_KIND, ActorLoadError } from '../../errors.js';
 import { connectMCPClient } from '../../mcp/client.js';
 import { EXTERNAL_TOOL_CALL_TIMEOUT_MSEC } from '../../mcp/const.js';
 import type { PaymentProvider } from '../../payments/types.js';
-import type {
-    ActorToolResolutionResult,
-    ApifyToken,
-    InternalToolArgs,
-    ToolDescriptionContext,
-    ToolEntry,
-    ToolInputSchema,
-} from '../../types.js';
+import type { ApifyToken, InternalToolArgs, ToolDescriptionContext, ToolEntry, ToolInputSchema } from '../../types.js';
 import { ACTOR_TOOL_MODE, ALL_TOOLS_PRESENT, TOOL_TYPE } from '../../types.js';
 import { getActorDefinitionCached, getActorToolResolutionCached } from '../../utils/actor.js';
 import { compileSchema } from '../../utils/ajv.js';
@@ -175,13 +168,14 @@ export function buildCallActorAppsDescription(ctx: ToolDescriptionContext = ALL_
  */
 function respondStandbyRejection(
     error: ActorLoadError,
-    opts: { mcpSessionId?: string; actorId?: string },
+    opts: { mcpSessionId?: string; actorId?: string; paymentProviderId?: string },
 ): ToolResponse {
     log.softFail('Rejecting call-actor for unsupported standby Actor', {
         // The error's canonical Actor name, so both call shapes log the same value even when the
-        // caller addressed the Actor by ID.
+        // caller addressed the Actor by ID. `actorLoadErrorKind` distinguishes which cell rejected it.
         actorName: error.actorName,
         mcpSessionId: opts.mcpSessionId,
+        paymentProviderId: opts.paymentProviderId,
         actorLoadErrorKind: error.kind,
         failureCategory: FAILURE_CATEGORY.INVALID_INPUT,
     });
@@ -349,15 +343,7 @@ export async function checkPaymentProviderStandbyConflict(params: {
 
     // Canonical name, not the caller's identifier — it may be an Actor ID or carry a `:toolName` suffix.
     const error = ActorLoadError.standbyPaymentNotSupported(actorDefinitionWithInfo.definition.actorFullName);
-    log.softFail('Rejecting call-actor for standby Actor under third-party payment provider', {
-        actorName: error.actorName,
-        paymentProviderId: paymentProvider.id,
-        mcpSessionId,
-        actorLoadErrorKind: error.kind,
-        failureCategory: FAILURE_CATEGORY.INVALID_INPUT,
-    });
-
-    return respondUserError(error.message, { detail: error.kind });
+    return respondStandbyRejection(error, { mcpSessionId, paymentProviderId: paymentProvider.id });
 }
 
 /**
@@ -379,23 +365,19 @@ export function resolveActorContext(actorName: string): {
 }
 
 /**
- * Handles the MCP tool call flow (when actorName contains ":toolName").
- * Returns a response if handled, or null if this is not an MCP tool call.
+ * Proxies the MCP tool call flow (when actorName contains ":toolName"). The caller routes here only
+ * for {@link ACTOR_TOOL_MODE.MCP} Actors, so `mcpServerUrl` is always known good.
  */
 export async function handleMcpToolCall(params: {
     baseActorName: string;
     mcpToolName: string;
     input: Record<string, unknown>;
-    resolution: ActorToolResolutionResult | null;
+    mcpServerUrl: string;
     apifyToken: string;
     mcpSessionId?: string;
     signal: AbortSignal;
-}): Promise<ToolResponse | null> {
-    const { baseActorName, mcpToolName, input, resolution, apifyToken, mcpSessionId, signal } = params;
-
-    if (resolution?.toolMode !== ACTOR_TOOL_MODE.MCP) {
-        return respondServerError(`Actor '${resolution?.actorFullName ?? baseActorName}' is not an MCP server.`);
-    }
+}): Promise<ToolResponse> {
+    const { baseActorName, mcpToolName, input, mcpServerUrl, apifyToken, mcpSessionId, signal } = params;
 
     if (!input) {
         return respondServerError(
@@ -409,9 +391,9 @@ export async function handleMcpToolCall(params: {
 
     let client: Client | null = null;
     try {
-        client = await connectMCPClient(resolution.mcpServerUrl, apifyToken, mcpSessionId);
+        client = await connectMCPClient(mcpServerUrl, apifyToken, mcpSessionId);
         if (!client) {
-            return respondServerError(`Failed to connect to MCP server ${resolution.mcpServerUrl}`);
+            return respondServerError(`Failed to connect to MCP server ${mcpServerUrl}`);
         }
 
         const result = await client.callTool(
@@ -613,20 +595,26 @@ export async function callActorPreExecute(
         };
     }
 
-    // Handle MCP tool calls
+    // A `:toolName` suffix only means anything for an MCP-server Actor; every other mode is a dead end.
     if (mcpToolName) {
-        const mcpResult = await handleMcpToolCall({
-            baseActorName,
-            mcpToolName,
-            input: parsed.input as Record<string, unknown>,
-            resolution,
-            apifyToken,
-            mcpSessionId,
-            signal: toolArgs.signal,
-        });
-        if (mcpResult) {
-            return { earlyResponse: mcpResult };
+        if (resolution?.toolMode !== ACTOR_TOOL_MODE.MCP) {
+            return {
+                earlyResponse: respondServerError(
+                    `Actor '${resolution?.actorFullName ?? baseActorName}' is not an MCP server.`,
+                ),
+            };
         }
+        return {
+            earlyResponse: await handleMcpToolCall({
+                baseActorName,
+                mcpToolName,
+                input: parsed.input as Record<string, unknown>,
+                mcpServerUrl: resolution.mcpServerUrl,
+                apifyToken,
+                mcpSessionId,
+                signal: toolArgs.signal,
+            }),
+        };
     }
 
     return { parsed, baseActorName, mcpToolName };
--- a/src/tools/actors/call_actor.ts
+++ b/src/tools/actors/call_actor.ts
@@ -629,7 +629,9 @@ export async function callActorPreExecute(
         }
         if (resolution.toolMode !== ACTOR_TOOL_MODE.MCP) {
             return {
-                earlyResponse: respondServerError(`Actor '${resolution.actorFullName}' is not an MCP server.`),
+                earlyResponse: respondUserError(`Actor '${resolution.actorFullName}' is not an MCP server.`, {
+                    detail: 'tool name given for a non-MCP Actor',
+                }),
             };
         }
         return {
--- a/src/tools/actors/call_actor.ts
+++ b/src/tools/actors/call_actor.ts
@@ -359,9 +359,10 @@ export async function checkPaymentProviderStandbyConflict(params: {
     // Canonical name, not the caller's identifier — it may be an Actor ID or carry a `:toolName` suffix.
     const error = ActorLoadError.standbyPaymentNotSupported(actorDefinitionWithInfo.definition.actorFullName);
     log.softFail('Rejecting call-actor for standby Actor under third-party payment provider', {
-        actorName: baseActorName,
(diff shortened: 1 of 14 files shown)


vojtechj-apify:
Pushed fce2f0e with a small fix found during review:

**What it does:** `call-actor "acme/ghost:tool"` on a nonexistent Actor now answers *"Actor 'acme/ghost' was not found"* instead of the false claim *"Actor 'acme/ghost' is not an MCP server."*

**Why:**
- The old message lies — "is not an MCP server" implies the Actor exists. An agent then retries as a plain run, gets a *different* error ("was not found"), and wastes a round-trip on an Actor that never existed.
- The two call shapes now agree: bare call → "not found"; `:tool` call → also "not found". One Actor, one answer.
- This was unfixable on master (`string | false` couldn't tell "unknown" from "known non-MCP") — your new `resolution === null` state makes the distinction free, so the fix is two lines plus extracting the existing not-found response from `resolveAndValidateActor` into a shared `respondActorNotFound()` (message and telemetry byte-identical for the run path).

The "is not an MCP server" answer is untouched for its intended case — known Actors that aren't MCP servers — and all matrix tests pinning it pass; added one pinning the unknown-Actor case.

vojtechj-apify:
**One question for @jirispilka on the telemetry section of the PR body**:

The body says three caller-mistake rejections move to `SOFT_FAIL`/`INVALID_INPUT` — both standby kinds, and `actor:toolName` aimed at a non-MCP Actor. The two standby kinds do (via `respondStandbyRejection` → `respondUserError`), but the third still answers via bare `respondServerError` in `callActorPreExecute`:

```ts
if (resolution.toolMode !== ACTOR_TOOL_MODE.MCP) {
    return {
        earlyResponse: respondServerError(`Actor '${resolution.actorFullName}' is not an MCP server.`),
    };
}
```

`respondServerError` with no `error` records `FAILED` + `INTERNAL_ERROR` — same classification as master, so this rejection would still raise error-level alerts on the hosted server, which is exactly what the telemetry section says it's fixing.

Was the intent to move it (one-liner: `respondUserError(..., { detail: ... })`), or should the PR body claim only the two standby kinds — leaving this one for #668 alongside `CALL_ACTOR_MCP_MISSING_TOOL_NAME_MSG`?

vojtechj-apify (changes requested):
Please address the telemetry question above.

jirispilka:
> **One question for @jirispilka on the telemetry section of the PR body**:
> 
> The body says three caller-mistake rejections move to `SOFT_FAIL`/`INVALID_INPUT` — both standby kinds, and `actor:toolName` aimed at a non-MCP Actor. The two standby kinds do (via `respondStandbyRejection` → `respondUserError`), but the third still answers via bare `respondServerError` in `callActorPreExecute`:

Thanks, Telemetry fixed in the code

vojtechj-apify (approved):
Thanks for the clarification. LGTM!

src/tools/actors/call_actor.ts
```diff
@@ -598,21 +622,29 @@ export async function callActorPreExecute(
...
+        if (resolution === null) {
+            return { earlyResponse: respondActorNotFound(baseActorName, loadedToolNames) };
         }
+        if (resolution.toolMode !== ACTOR_TOOL_MODE.MCP) {
+            return {
+                earlyResponse: respondUserError(`Actor '${resolution.actorFullName}' is not an MCP server.`, {
+                    detail: 'tool name given for a non-MCP Actor',
+                }),
+            };
+        }
```
RobertCrupa, line 637: do we not want to include the `toolTelemetry: { actorId: resolvedActorId },` here?

src/tools/actors/call_actor.ts
```diff
@@ -586,8 +600,18 @@ export async function callActorPreExecute(
...
-    const isActorMcpServer = !!mcpServerUrlOrFalse;
+    const resolution = await getActorToolResolutionCached(baseActorName, apifyClientForDefinition);
+
+    if (resolution?.toolMode === ACTOR_TOOL_MODE.STANDBY_WITHOUT_MCP) {
+        return {
+            earlyResponse: respondStandbyRejection(
+                ActorLoadError.standbyWithoutMcpNotSupported(resolution.actorFullName),
+                { mcpSessionId },
+            ),
+        };
```
RobertCrupa, line 612: do we not want to return the `toolTelemetry: { actorId: resolvedActorId },` here?

src/utils/actor.ts
```diff
@@ -44,13 +46,34 @@ export async function getActorDefinitionCached(
...
+        definition: cached.definition,
+        actor: cached.info,
+        webServerMcpPath,
+    });
+    const { actorFullName } = cached.definition;
+    // `resolveActorToolMode` returns MCP only when `webServerMcpPath` is set.
+    if (toolMode === ACTOR_TOOL_MODE.MCP) {
+        return {
+            toolMode,
+            mcpServerUrl: await getActorMCPServerURL(cached.definition.id, webServerMcpPath!),
```
RobertCrupa, line 74: Should `await getActorMCPServerURL` be guarded? It can throw for a webServerMcpPath that escapes the Actor's standby origin

src/tools/actors/call_actor.ts
```diff
@@ -472,30 +499,17 @@ export async function resolveAndValidateActor(params: {
     });
 
     // NOT_FOUND falls through to the structured "Actor not found" response below.
-    // Any other error (LOAD_FAILED / STANDBY_PAYMENT_NOT_SUPPORTED) is rethrown so the
-    // outer call-actor handler reports it; STANDBY is also caught upstream by the
-    // call-time guard in server.ts, so it's a defensive fallback here.
+    // Any other error (LOAD_FAILED / STANDBY_PAYMENT_NOT_SUPPORTED /
+    // STANDBY_WITHOUT_MCP_NOT_SUPPORTED) is rethrown so the outer call-actor handler reports it;
+    // STANDBY_PAYMENT_NOT_SUPPORTED is also caught upstream by `checkPaymentProviderStandbyConflict`
+    // in the tool-call engine, so it's a defensive fallback here.
```
RobertCrupa: Not sure if this is accurate as `getActorsAsTools` is called two lines above without paymentProvider, so the gate `if (paymentProvider && isActorBlockedUnderPaymentProvider(actorInfo.actor)) {` is never triggered and thus not caught.

RobertCrupa (approved):

MQ37 (approved):
LGTM. Pushed the standby-without-MCP message fix for `getMcpToolsMessage` (commit 7fe65dc7) plus an oxfmt fix (56f55dd3).

lokicik:
I tested this PR with a normal agent request: run an Actor with a URL and return its run ID and status.

The Actor had standby disabled but still contained an old MCP path in its metadata, which is the routing case addressed by this PR. I ran the same request three times with each model:

| Model | Base (`df612ef`) | Current PR (`56f55dd`) |
| --- | ---: | ---: |
| DeepSeek V4 Flash | 0/3 | 3/3 |
| GPT-5.6 Luna | 0/3 | 3/3 |
| Qwen 3.5 Flash | 0/3 | 3/3 |
| MiMo v2.5 | 0/3 | 3/3 |
| **Total** | **0/12** | **12/12** |

On Base, all 12 requests were routed through the stale MCP path and failed. On the current PR, all 12 used the normal Actor route and returned the expected run ID and status.

src/utils/actor.ts
jirispilka: Good catch, it does throw. It's pre-existing though: getActorMcpUrlCached on master was equally unguarded

Not fixing it here, because the safe fix isn't local. The code was already reviewed, let's not make significant changes.

Tracked in #1321.

src/tools/actors/call_actor.ts
jirispilka: Good catch. ActorToolResolutionResult doesn't carry the Actor ID at all, so it needs the type extended rather than a one-liner, not doing that on already-reviewed code. Tracked in #1321

src/tools/actors/call_actor.ts
jirispilka: the same as above

src/tools/actors/call_actor.ts
jirispilka: Correct, the comment gave the wrong reason

PR #1279, opened 2026-08-19
feat: Prefix MCP-proxied tool names with Actor tool name

jirispilka:
## Why

Closes #1277

The exposed name did not identify which Actor provided a proxied tool.

## What changed

Proxied tool names now use:

```text
{username}--{actor-name}--{tool-name}
```
Simple names keep the complete username:

```text
apify/example-mcp-server + add → apify--example-mcp-server--add (30)
```

Dots in usernames are escaped:

```text
the.unc/my-mcp-server + add → the-dot-unc--my-mcp-server--add (31)
```

Actor names may contain `--`. The username and Actor name are read from `actorFullName`, not by splitting the assembled tool name:

```text
crawler_enthusiast/dataset--exporter-mcp-server + extract 
→ crawler_enthusiast--dataset--exporter-mcp-server--extract (57)
```

Tool names remain limited to 64 characters. When a name exceeds the limit, the username is capped to `MAX_TOOL_NAME_USERNAME_LENGTH` (8). A four-character hash of the uncapped name is appended:

```text
crawler_enthusiast/web-scraping-dataset-tools-mcp-server + search
→ crawler_--web-scraping-dataset-tools-mcp-server--search-e41a (60)

crawler_enthusiast/web-scraping-dataset-tools-mcp-server + get-company-profile
→ crawler_--web-scraping-dataset-tools-mcp-server--get-compan-b992 (64)
```

The cap decision uses the complete tool list returned by the Actor. If every tool fits, the full username remains:

```text
longusername1/my-mcp-server + add → longusername1--my-mcp-server--add (33)
```

If any offered tool exceeds the limit, every tool from that Actor uses the same capped username:

```text
longusername1/my-mcp-server + add → longuser--my-mcp-server--add-34ee (33)

longusername1/my-mcp-server + a-tool-name-long-enough-to-push-past-the-limit-xxxx
→ longuser--my-mcp-server--a-tool-name-long-enough-to-push-pa-fa1b (64)
```

This prevents one Actor from exposing tools with different username prefixes.

Capping the username is lossy, so a capped name always carries the hash — even when it would have fit uncapped. Without it, `crawler_alpha/x` and `crawler_bravo/x` produce the same string.

Dispatch is unchanged. Calls still use `originToolName` in `tool_dispatch.ts`; only the exposed tool name changes.

## How a call resolves

Nothing parses the exposed name. It is derived once at load, and only ever looked up at call time — the routing data rides on the tool entry, not on the string.

**Derived once, at load:**

```text
[...]

PR #1284, opened 2026-08-19
refactor: Remove the write-only serverId field on ActorMcpTool

jirispilka:
The `serverId` was written on every proxied tool and read nowhere — not in `src/`, not in `apify-mcp-server-internal`. Dispatch routes on `originToolName` + `serverUrl`, so it carried no behavior, and it cost a `sha256` per proxied tool at list time for no consumer.

### Conflicts with #1279 — read before merging

Whichever merges second needs to reconcile:
- if **#1279 first** → rebase this branch, and drop the `serverId` half of that sentence;
- if **this first** → #1279's `AGENTS.md` hunk and its `proxy.ts` object literal both need updating.

Merging #1279 first is the simpler order, since this PR is the smaller rebase.


MQ37 (approved):
Verified serverId dead field (no readers anywhere, both repos). Pushed AGENTS.md fix: getMCPServerID has two consumers, not one.

PR #1291, opened 2026-08-20
refactor: Extract shared remote MCP tool-call helper

MQ37:
## What

Extracts `withRemoteMcpClient()` into `src/mcp/remote_tool_call.ts`: connect, hand the live client to a caller-supplied `run(client)` callback, always close, return a closed outcome union (`connect-failed` / `aborted` / `error` / `success`). Replaces duplicated connect/callTool/close try-catch-finally (including the same abort-race macrotask-yield workaround) in `tool_dispatch.ts`'s `ACTOR_MCP` case and `call_actor.ts`'s MCP passthrough. Each caller keeps its own `callTool` shape and response handling.

## Why

Two copies of a non-obvious workaround drift the moment one is fixed and the other isn't.

One incidental behavior delta: the `ACTOR_MCP` `_meta.progressToken` gate now checks `progressToken != null` (previously only checked `shouldForwardNotifications`). Old code sent `_meta: {}` on the wire when a sync call had no progress token; new code omits `_meta` entirely. Same semantics for any spec-conformant server, confirmed via the SDK's request/serialization path.

## Testing

`pnpm run type-check` / `lint` / `format` / `check:agents` clean. `pnpm run test:unit`: 97 files, 1457 passed, 1 skipped — includes 2 new tests for `call_actor.ts`'s `connect-failed`/`error` outcomes.


jirispilka:
@MQ37 I removed myself from the review, I won't have time for it this week

src/mcp/tool_dispatch.ts
```diff
@@ -128,10 +127,51 @@ export async function dispatchToolCall(params: {
...
+                            // hand the remote a token nobody listens to.
+                            ...(shouldForward ? { _meta: { progressToken } } : {}),
+                        },
+                        CallToolResultSchema,
+                        { timeout: EXTERNAL_TOOL_CALL_TIMEOUT_MSEC, signal },
+                    );
+                },
+            );
+
+            switch (outcome.outcome) {
```
RobertCrupa, line 174: Should we add a default case to the switch statement?

RobertCrupa (approved):

src/mcp/tool_dispatch.ts
MQ37: Added — default case throws with a `satisfies never` exhaustiveness guard, mirroring the outer TOOL_TYPE switch. 5afc4e1f

PR #1293, opened 2026-08-20
feat: Return task input verbatim and align task tools with the API

jirispilka:
Task tools now return the stored task `input` verbatim, under the same key the API, CLI, and apify-client use. Everything in this PR came out of the task-tool workflow evals (#1294): agents could not answer "what input is my task configured with?", could not verify their own updates, and hallucinated values from the field-name list.

Not obvious from the code:

- **Returning the input is safe.** The platform encrypts input fields declared `isSecret: true` server-side on save. Verified empirically: a task created via the raw API with a plaintext secret returns `ENCRYPTED_VALUE:...` from both `GET actor-task` and `GET actor-task/input`; the plaintext never comes back. Non-secret values are already returned by every other Apify surface.
- **`seoDescription` is required to publish**, which no docs state — discovered via API probing after eval agents got stuck. Descriptions now list the full requirements, and the schema declares the SEO length limits (60/160) the API enforces.
- **Contract change:** the task tools' `structuredContent` replaces `inputFields` (names only) with `input`. 
- Tool errors append the machine-readable API error type, e.g. `(API error type: actor-task-name-not-unique)`.

### DIFF
--- a/src/tools/structured_output_schemas.ts
+++ b/src/tools/structured_output_schemas.ts
@@ -353,7 +353,9 @@ export const reportProblemToolOutputSchema = {
 };
 
 /**
- * Schema shared by every Actor task tool. Input values are omitted on purpose (they may hold secrets) `inputFields` lists the field names that `publicConfig.inputSchemaFields` can use.
+ * Schema shared by every Actor task tool. `input` mirrors the API task object verbatim; input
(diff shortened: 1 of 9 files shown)


Janjiran (approved):
lgtm 🚀

PR #1294, opened 2026-08-20
feat(evals): Add task-tool workflow eval suites

### LINKED ISSUE #1288: feat: Add task tools to the default eval tools
### Problem or motivation

The eval harness only loads a fixed set of tool categories, and the task tools had to be added to that list by hand for the new cases to work. They should be part of the default tools the evals run against so this doesn't have to be remembered per category.

### What would good look like?

_No response_

### Alternatives considered

_No response_

jirispilka:
Workflow eval suites for the Actor task tools. Closes #1288.

I created a new dataset specifically for this, we'll merge it later.
I did not want to mixed it up with existing evals as they are not "clean" (=there are errors)
We will improve the test cases later. I think they are overly verbose, no one gonna prompt in that way.

I'm sorry, we'll need to clean this up afterwards but I don't have time for it now 😊 

- Two Langfuse datasets, split on purpose: `tasks-evals` (7 proper cases) and `tasks-evals-errors` (3 cases that provoke errors: name collision, not-found lookup, publish-requirement discovery). Error-provoking cases inside a proper suite mask real failures, so they are quarantined and run with `--allow-tool-errors`.
- `--subscription` runs the agent on the local Claude Code login instead of `ANTHROPIC_API_KEY`.


evals/workflows/run_workflow_evals.ts
```diff
@@ -115,13 +129,19 @@ async function main() {
...
     ]);
     if (missing.length > 0) {
         console.error(`❌ Error: missing environment variable(s): ${missing.join(', ')}`);
         process.exit(1);
     }
 
+    // The Agent SDK's Claude Code subprocess falls back to the local login only when no
+    // API key is in its environment, which it inherits from this process.
+    if (argv.subscription) {
+        delete process.env.ANTHROPIC_API_KEY;
```
jirispilka, line 144: This is not a good solution but I guess we can live it for now 😊

evals/workflows/workflow_judge.ts
```diff
@@ -131,28 +134,31 @@ export async function evaluateConversation(
...
-    } catch (error) {
-        throw new Error(
-            `Failed to parse judge response: ${error instanceof Error ? error.message : String(error)}\n` +
-                `Raw response: ${rawResponse}`,
+    // Some OpenRouter providers occasionally ignore the JSON schema and answer in plain
+    // text (e.g. "PASS: ..."). One fresh judge call recovers those without rerunning the
+    // far more expensive agent conversation; a second malformed answer still throws.
+    let lastError: unknown;
+    let lastRawResponse = '';
+    for (let attempt = 1; attempt <= JUDGE_PARSE_ATTEMPTS; attempt++) {
```
jirispilka, line 142: I've added attempts, sometimes the judge failed on a slow interner

RobertCrupa (approved):
the datasets loog good to me, a bit verbose but like you said we can (and should) cut it down. For now it feels more like a regression test to ensure that tasks are working.

Will you keep the dataset snapshot in version control or is it only here for review purposes?

MQ37 (commented):
LGTM in general, no correctness bugs found.

Pushed `c6cb5b9d`: fixed a docstring claiming judge/OpenRouter errors retry here — they don't, the OpenAI SDK already retries them 2x underneath.

Snapshots: nothing reads `dataset_snapshot_*.json` at runtime, no importer, no CI freshness check — is this a review artifact or meant as a backup? Pick one and say so in the README, or add a CI drift check.

Other asks: `tasks_fixtures.ts` deletes all `eval-*` tasks with no confirm/dry-run — risky on wrong account; zero-tool-error gate now applies to the legacy `workflow-evals` dataset, which the PR body says is unclean — worth one run to see the damage; agent retry catches all errors, not just transient ones.

jirispilka:
@RobertCrupa - removed the dataset, thanks
@MQ37 - added the dry-run, we'll re-design the eval suite anyway

PR #1295, opened 2026-08-21
fix: Keep cross-tool references out of task tool summaries

jirispilka:
Follow-up to #1293, split out of the eval-suites branch.

📄 **[Reviewer walkthrough](https://claude.ai/code/artifact/599bc39a-cd89-4df6-840a-ac46346c0ce6)** — longer visual version with the full evidence.

## Problem

So a session started with `?tools=create-actor-task` got an input schema telling the model to call `fetch-actor-details`, which was never served. Task summaries had the same bug: `create-actor-task` named `publish-actor-task`, `publish-actor-task` named `get-actor-task`, both unconditionally.

## Fix

Deleted, not re-gated: each summary reference was already stated in a `hasTool`-gated description, so summaries now say only what the call did. The `.describe()` guidance moved into `buildDescription`. `src/tools/AGENTS.md` documented only the description surface — that silence let the bug be written — so it now covers all three.

Two removals turned out not to be redundant. Both were caught only by running the task evals (agent model `claude-haiku-4-5`; errors suite 3/3, proper suite 7/7 after fixing):

- The `fetch-actor-details` bullet must come *after* name resolution ("Once you have the exact Actor ID…") — placed first, the agent guessed an Actor slug instead of searching.
- The publish summary's `publishedAt` sentence is load-bearing: without it, an agent asked to "confirm it is live" claimed confirmation it never obtained. Restored; it names no tool.

## Also here, separately reviewable

- `ApifyApiError.type` read moved to `apify_errors.ts`, the module that owns error-type classification. Behaviour-identical; drops the `apify-client` value import from `utils/mcp.ts`. Adds the coverage the `(API error type: …)` suffix never had — the `task-create-collision` eval depends on `actor-task-name-not-unique` reaching the model.
- SEO caps `.max(60)` / `.max(160)` verified against the live API: 61/161 rejected with `cannot-publish-actor-task`, and it validates on **every** `publicConfig` write — so rejecting client-side costs the user nothing.

Same bug in other tool families is filed, not fixed here: #1296 (storage tools), #1297 (`AUTO_INJECTED_TOOLS`).

### DIFF
--- a/src/tools/AGENTS.md
+++ b/src/tools/AGENTS.md
@@ -51,7 +51,22 @@ does not exist. Any other cross-tool reference must be gated per session: define
 `buildDescription(ctx)` on the entry (see `ToolDescriptionContext` in `../types.ts`), wrap the
 reference in `ctx.hasTool(...)`, and set `description` to the `ALL_TOOLS_PRESENT` render. Rendering
 happens once, at the tools/list boundary (`getToolPublicFieldOnly` in `../utils/tools.ts`). Enforced
-by `tests/unit/tools.mode_contract.test.ts`.
+by `tests/unit/tools.mode_contract.test.ts`, which scans `description` only.
+
+Input-schema field text (`.describe()` on a Zod field) reaches `tools/list` verbatim — nothing
+renders it per session — so never name a tool there; put the guidance in `buildDescription` behind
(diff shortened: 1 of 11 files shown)


MQ37 (approved):
LGTM.

Janjiran (approved):
🚀

PR #1298, opened 2026-08-21
feat: Add apify/web-fetch as a default Actor tool

jirispilka:
Adds `apify/web-fetch` to the default Actors next to `apify/rag-web-browser`, with client-agnostic steering so any model picks the right web tool (web-fetch: one specific URL, verbatim content; rag-web-browser: search by query). Also two run-response text fixes for all Actor runs, found by evals: an unverified `itemCount: 0` no longer reads as a "no items" conclusion, and the KV key-count summary no longer counts the INPUT record.

Tool description wording is eval-tuned — behavior validated by Langfuse eval ladders (`web-fetch-evals`, `web-selection-evals`) on claude-opus-5 / claude-sonnet-5 / claude-haiku-4-5; suites land in a follow-up PR stacked on #1294. Known residual documented there: claude-haiku-4-5 silently rewrites ftp:// URLs to https:// despite both notes.

While on this PR: fixed case when dataset returns 0 items.

**Follow-ups**
- apify/apify-mcp-server-internal#753 — hosted web UI default-Actor list, deploy sequencing, contract-suite strings
- #1303 — pre-existing cleanups kept out of this PR
- https://github.com/apify/apify-mcp-server/issues/1305

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -105,10 +105,39 @@ Examples of when to use:
 - User needs to fetch specific content now (e.g., "Fetch news articles from CNN", "Get product info from Amazon")
 - User has time indicators like "today", "current", "latest", "recent", "now"
 
-This is for general web scraping and immediate data needs. For repeated/scheduled scraping of specific platforms (e-commerce, social media), consider suggesting a specialized Actor from the Store for better performance and reliability.`;
+This is for general web scraping and immediate data needs. For repeated/scheduled scraping of specific platforms (e-commerce, social media), consider suggesting a specialized Actor from the Store for better performance and reliability.
+When the user provides one specific URL and wants that page's full or verbatim content, prefer the dedicated apify/web-fetch tool when it is available - this tool is for searching and scraping by query.
+If a scraped page comes back blocked or empty (e.g. the crawl reports a 403 or the page text is missing), do not give up: retry that URL with the apify/web-fetch tool when it is available - its anti-bot fetching gets through blocks this tool cannot.`;
+
+export const WEB_FETCH = 'apify/web-fetch';
+/**
+ * Appended to the `apify/web-fetch` Actor tool description. Client-agnostic on purpose:
+ * no references to any specific client or its built-in tools, so the same text works
+ * for every MCP client. Tune only based on eval results (`web-fetch-evals` dataset).
+ */
+export const WEB_FETCH_ADDITIONAL_DESC = `Use this tool to fetch a specific http(s) URL and return its complete content (one URL per call; http and https only).
+It renders JavaScript and bypasses anti-bot protection, so it also retrieves pages where a plain HTTP fetch gets blocked, fails with an error such as 403 or 429, or returns incomplete content.
+The page comes back verbatim - full content, no summarization - as Markdown, plain text, HTML, the raw response body, or the list of links on the page.
+
+Examples of when to use:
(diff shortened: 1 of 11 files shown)


src/utils/server-instructions/index.ts
```diff
@@ -101,6 +101,8 @@ ${
         : ''
 }- **\`${HELPER_TOOLS.STORE_SEARCH}\` vs ${RAG_WEB_BROWSER}:**
   \`${HELPER_TOOLS.STORE_SEARCH}\` finds robust and reliable Actors for specific websites; ${RAG_WEB_BROWSER} is a general and versatile web scraping tool.
+- **${WEB_FETCH} vs ${RAG_WEB_BROWSER}:**
+  ${WEB_FETCH} fetches one specific URL and returns its full content verbatim; ${RAG_WEB_BROWSER} searches the web by query and returns content from the top results.
```
daniil-poletaev, line 105: > full content verbatim

Not sure about this, because Web fetch cleans HTML unless `raw` format is passed 🤔

daniil-poletaev: > searches the web by query

Isn't it what RAG browser does? Web fetch doesn't search by  query

src/const.ts
```diff
@@ -105,10 +105,39 @@ Examples of when to use:
...
+
+export const WEB_FETCH = 'apify/web-fetch';
+/**
+ * Appended to the `apify/web-fetch` Actor tool description. Client-agnostic on purpose:
+ * no references to any specific client or its built-in tools, so the same text works
+ * for every MCP client. Tune only based on eval results (`web-fetch-evals` dataset).
+ */
+export const WEB_FETCH_ADDITIONAL_DESC = `Use this tool to fetch a specific http(s) URL and return its complete content (one URL per call; http and https only).
+It renders JavaScript and bypasses anti-bot protection, so it also retrieves pages where a plain HTTP fetch gets blocked, fails with an error such as 403 or 429, or returns incomplete content.
+The page comes back verbatim - full content, no summarization - as Markdown, plain text, HTML, the raw response body, or the list of links on the page.
```
daniil-poletaev, line 120: > full content

Again, by default it's cleaned, but `raw` provides content verbatim

daniil-poletaev: Maybe "... plain text, clean HTML ..."

daniil-poletaev (approved):
Looks good, just few nits regarding description

src/utils/server-instructions/index.ts
MQ37: yes, but the LLM is interested in the clean HTML and I think it is ok considering cleaned HTML = full content (I think fetch != curl)

MQ37 (approved):
LGTM.

Pushed one small fix (ba5870b): `summarizeKv`'s truncated-KV-store branch counted the platform's `INPUT` record toward the "at least N keys" floor, while the exact-count branch already excluded it. A 50-key truncated window containing `INPUT` reported "at least 50 keys" when only 49 are real output. Floor now subtracts `INPUT` in the truncated case too; added a regression test covering truncation + `INPUT` together.

src/const.ts
jirispilka: Thanks @daniil-poletaev . We'll fine this description later on so that it is clear for the LLM. I can see that Kuba has already merged it.

PR #1300, opened 2026-08-21
feat(evals): Add web-fetch and web-tools-selection eval suites

jirispilka:
Stacked on #1299 (which stacks on #1294). Docs for four new Langfuse datasets — the snapshots stay out of version control per review (Langfuse is the only copy, see README Design Decisions); the tool changes they validated are in #1298.

- `web-fetch-evals` (8) / `web-fetch-evals-errors` (3): the `apify/web-fetch` default tool 
- `web-selection-evals` (7) / `web-selection-evals-errors` (2): the rag-web-browser vs web-fetch vs search-actors clash, specialized-Actor-over-generic-fetch

Final ladder on the current tool descriptions (#1298): claude-opus-5, claude-sonnet-5, and claude-haiku-4-5 all green on both families, except one documented residual — Haiku reproducibly rewrites ftp:// to https:// silently on `web-fetch-unsupported-protocol` (3x, despite the scheme note in the description and on the `url` parameter); the case keeps documenting it on purpose.

PR #1301, opened 2026-08-21
chore: Add the creating-workflow-evals skill to the repo

jirispilka:
This skill is specific for a tool family (because I started with tasks and then web-fetch, rag-web-browser).
We can update it later to be generic.

- probe-first case design, the two-suite split (proper vs error-provoking), waves, Opus-calibrate-then-ladder
- the failure diagnosis order: case / judge / world (live target down or changed) / model (policy refusals, reproducible habits) / product / description
- rules that prevent rework: judge-blindness clauses, one-right-answer selection targets, fail-false-claims-not-silent-success, snapshot after every dataset edit
- the Langfuse env vars and CLI iteration loop (upsert-on-id case editing, transcript dumps, error sweeps)

Distilled from the tasks (#1294), web-fetch, and web-selection (#1300) suite builds


RobertCrupa (approved):
LGTM!

Issue #1296, opened 2026-08-21
Storage tools name ungated sibling tools in result text


jirispilka:
Storage tools name sibling tools in their result text (`summary` / `nextStep` / `inspectHint`) without checking whether the session was served them. A client using a narrow selector — e.g. `?tools=get-dataset` — receives a `nextStep` telling the model to call a tool that is not in its `tools/list`.

Tool `description` text is gated per session by `hasTool`. Result text is not gated by anything unless the site opts in via `loadedToolNames`, and only one call site in the repo does (`suggestTool` in `storage/storage_helpers.ts`).

**Affected**, all in `src/tools/storage/`:

- `get-dataset` and `get-dataset-schema` — `nextStep` names `get-dataset-items` / `get-dataset`
- `get-dataset-list` and `get-key-value-store-list` — `inspectHint` names `get-dataset` / `get-key-value-store`
- `get-key-value-store` — `nextStep` names `get-key-value-store-keys`
- `get-key-value-store-keys` — the empty-store branch of its `nextStep` helper names `get-key-value-store`

Some of these are currently masked by `AUTO_INJECTED_TOOLS`, but that injection is itself conditional — see the companion issue.

The same class of problem in the task tools was fixed in #1295, which also documents the three surfaces a tool name can reach the client on in `src/tools/AGENTS.md`.

jirispilka:
Context: the [reviewer walkthrough](https://claude.ai/code/artifact/599bc39a-cd89-4df6-840a-ac46346c0ce6) for #1295 lays out the three surfaces a tool name can reach the client on, and which of them are gated. (Private link — ask @jirispilka if you cannot open it.)

jirispilka:
Another instance, in the actors family rather than storage: `buildActorNotFoundResponse` in `src/tools/actors/fetch_actor_details.ts` tells the model `You can search for available Actors using the tool: search-actors`, with no gate — it takes only the Actor name, not `loadedToolNames`. `search-actors` is not auto-injected, so a session started with `?tools=fetch-actor-details` is pointed at a tool it never received.

`call-actor` has the correctly-gated version of the same hint (`buildActorNotFoundHint`), so the pattern to copy is already next door.

jirispilka:
Fixed for the actors family in #1302 (the `buildActorNotFoundResponse` instance noted above). The storage instances in the description remain open.

PR #1302, opened 2026-08-21
fix: Gate the Actor-not-found recovery hint on the loaded tools

jirispilka:
Third instance of the #1295 bug (tracked in #1296), this time in the actors family. Review turned up two more, all fixed here.

Result text is gated by nothing, and `tools.mode_contract.test.ts` scans `description` only, so none of this was covered.

| Site | Names, ungated | Session that hits it |
| --- | --- | --- |
| `fetch-actor-details` not-found | `search-actors` | `?tools=fetch-actor-details` |
| `call-actor` not-found | `search-actors` | `?tools=call-actor` |
| `search-actors` footer | `fetch-actor-details` | `?tools=search-actors`, on every success |

None of those names are in `AUTO_INJECTED_TOOLS`. The first two need only a typo to reach.

All three are gated in place, so a session holding every tool gets byte-identical output. Verified against master, not assumed. `call-actor`'s not-found branch does not reuse `buildCallFailureRecoveryHint`, because its second offer is a detail lookup, which for a nonexistent Actor fails the same way.

`fetch-actor-details`'s description also gained the missing "resolve the name first" instruction. Only the `search-actors` half is gated; the rest names no tool and renders everywhere.

## This does not fix the guessing habit

`task-chain-hard-1` (#1294's `tasks-evals`): Sonnet 3/3, Haiku ~7/10 before this wording and 5/8 with it. No measurable change. The case is well-formed and the Haiku gap is a capability limit, so further rewording is not the lever.

That 5/8 predates the review, which split the sentence, so it does not describe what ships. Re-running is a one-variable comparison: the case loads `tasks` + `actors`, so every gate here is a no-op for it.

## Not fixed here

Same bug, unfiled, left out to keep this one thing:

- `actor_tools_factory.ts:124`: Actor tool descriptions name `call-actor`, and `?tools=<actor>` serves none. `mode_contract`'s stub description hides it from the sweeping guard.
- `call_actor_widget.ts:96`: names `call-actor` in an apps widget-only session. That sentence is the whole recovery path, so it needs replacement wording, not a gate.

1464 unit tests, type-check, lint, format, and `check:agents` clean. Merges cleanly with #1190 and #1294.

Written and reviewed with Claude Code.

### DIFF
--- a/src/tools/AGENTS.md
+++ b/src/tools/AGENTS.md
@@ -53,6 +53,18 @@ reference in `ctx.hasTool(...)`, and set `description` to the `ALL_TOOLS_PRESENT
 happens once, at the tools/list boundary (`getToolPublicFieldOnly` in `../utils/tools.ts`). Enforced
 by `tests/unit/tools.mode_contract.test.ts`.
 
+**Result text is a second surface, and `hasTool` does not reach it.** A tool name in a response
+body (`summary`, `nextStep`, `instructions`, an error's recovery sentence) is built while the tool
+runs, not at the tools/list boundary, so nothing gates it for you and `tools.mode_contract.test.ts`
+— which renders descriptions only — cannot see it. Gate it on `toolArgs.loadedToolNames`
+(`InternalToolArgs` in `../types.ts`), the names the session was actually served: see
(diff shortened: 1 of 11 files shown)


MQ37 (approved):
LGTM - tested e2e

Issue #1286, opened 2026-08-19
chore(evals): Rename workflow evals and replace stale MCP case


jirispilka:
## Context

Follow-up to #1239.

The current “workflow evals” name is too generic. The harness specifically evaluates an agent using the Apify MCP server, so call it “MCP agent evals.”

The `fetch-details-mcp-tools-then-call` case also depends on `apify/actors-mcp-server:fetch-apify-docs`. Replace it with the stable arithmetic tool from `apify/example-mcp-server`.

### Rename the harness

Use “MCP agent evals” consistently:

- `evals/workflows/` → `evals/mcp_agent/`
- `evals:workflow` → `evals:mcp-agent`
- `evals:workflow:export-dataset` → `evals:mcp-agent:export-dataset`
- `run_workflow_evals.ts` → `run_mcp_agent_evals.ts`
- Langfuse dataset `workflow-evals` → `mcp-agent-evals`
- Update related filenames, symbols, comments, help text, and documentation
- Do not retain aliases for the old internal commands

Use `git mv` for repository renames.

### Replace the MCP discovery-and-call case

Update the source item in Langfuse, then export `dataset_snapshot.json`.

```json
{
  "id": "discover-and-call-actor-mcp-tool",
  "category": "mcp",
  "query": "List the tools exposed by apify/example-mcp-server. Then use its add tool to add 17 and 25, and tell me the sum.",
  "reference": "The agent must first call fetch-actor-details with actor='apify/example-mcp-server' and output={ mcpTools: true }. It must then call call-actor with actor='apify/example-mcp-server:add' and input={ firstNumber: 17, secondNumber: 25 }. PASS only if both calls happened and the final answer states 42."
}
```

Archive the old `fetch-details-mcp-tools-then-call` item.

## Out of scope

- Judge, tracing, or scoring changes
- New eval cases
- MCP server runtime changes
- Changes to `apify-mcp-server-internal`

## Acceptance criteria

- The old workflow-eval naming is absent from active code and documentation.
- `pnpm run evals:mcp-agent -- --id discover-and-call-actor-mcp-tool` runs the replacement case.
- The case discovers the tools, calls `add`, and reports `42`.
- `dataset_snapshot.json` matches the active Langfuse dataset.
- `pnpm run type-check`, `pnpm run lint`, `pnpm run test:unit`, `pnpm run format`, and `pnpm run check:agents` pass.

## Internal repo impact

None. This changes only the evaluation harness and dataset.

PR #1313, opened 2026-08-25
chore(evals): Replace the stale MCP discovery eval case

RobertCrupa:
Second half of #1286. Stacked on #1312 (the rename) because the snapshot only lives at `evals/mcp_agent/` after it.
## What

`fetch-details-mcp-tools-then-call` called `apify/actors-mcp-server:fetch-apify-docs`, so the eval depended on the server under test shipping that tool. Replaced with `discover-and-call-actor-mcp-tool`, which discovers the tools of `apify/example-mcp-server` and calls its arithmetic `add` tool

## Verification

- `pnpm run evals:mcp-agent:export-dataset` wrote 30 cases from `mcp-agent-evals`; the diff is exactly the case swap.
- `fetch-actor-details` on `apify/example-mcp-server` confirms `add` takes `firstNumber` / `secondNumber`, matching the reference.
-  `pnpm run evals:mcp-agent -- --id discover-and-call-actor-mcp-tool`. runs and works as expected

## AI assistance

Written with Claude Code: the Langfuse dataset item archive/create and the snapshot export. Reviewed by me.

### DIFF
--- a/evals/mcp_agent/dataset_snapshot.json
+++ b/evals/mcp_agent/dataset_snapshot.json
@@ -18,16 +18,16 @@
     "reference": "The agent must find tik tok scraper, call it for natgeo profile, return the latest post and present it to the user."
   },
   {
+    "id": "discover-and-call-actor-mcp-tool",
+    "category": "mcp",
+    "query": "List the tools exposed by apify/example-mcp-server. Then use its add tool to add 17 and 25, and tell me the sum.",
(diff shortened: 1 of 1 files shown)


MQ37 (approved):

jirispilka:
@RobertCrupa I won't have time to look into this PR, please don't wait for my review, not to delay this PR

RobertCrupa:
Closing as this eval case is now in langfuse (the source of truth for our eval dataset)

PR #1315, opened 2026-08-25
fix: Redact tokens from local conformance output

### LINKED ISSUE #1245: fix: Redact tokens from local conformance output
## Problem

`scripts/test_conformance.sh` passes the Apify token in the conformance target URL. The conformance runner prints that URL, so a local `pnpm run test:conformance` exposes the raw token in terminal output.

CI masks the raw and percent-encoded token through GitHub Actions, but the local script has no equivalent output redaction.

The token also remains visible in process arguments. That cannot be fixed here while the conformance runner accepts authentication only through its target URL. Track header-based authentication upstream separately.

## Scope

- Redact raw and percent-encoded token values from local conformance output.
- Preserve line-buffered output and the runner's exit status.
- Add a regression check that output contains neither token representation.
- Document that process-argument exposure remains an upstream limitation.

Split from #1173.

SomSamantray:
## Why

`scripts/test_conformance.sh` passed `APIFY_TOKEN` in the conformance server URL, and the conformance runner printed that URL verbatim. CI already masks the raw and percent-encoded token via GitHub Actions `::add-mask::` (`.github/workflows/_conformance_tests.yaml`), but running `pnpm run test:conformance` locally had no equivalent — the token leaked to the terminal in plain text.

Closes #1245

## What changed

Before: `run_conformance()` piped nothing — the conformance runner's output (including the token in the URL it echoes) went straight to the terminal.

Now: both suite invocations pipe combined stdout+stderr through a new `redact_tokens()` awk filter that literal-replaces the raw and percent-encoded token with `***REDACTED***`, flushing after every line so a long-running suite doesn't look hung.

Two non-obvious details, both caught during review and fixed before this PR:
- `awk`'s `-v` assignment runs POSIX backslash-escape processing, so a token containing a literal backslash would have silently bypassed redaction. Reading the token via `ENVIRON` instead avoids that (and keeps the token out of `awk`'s own process arguments, which `-v` would otherwise expose the same way `--server-url` already does).
- Under the script's existing `set -uo pipefail`, the pipeline's exit status is the *rightmost* nonzero code — so a filter failure could in principle mask the conformance runner's real exit status. `run_conformance()` now returns `${PIPESTATUS[0]}` explicitly so the runner's own status always wins.

Local process-argument exposure (`ps`, `/proc/<pid>/cmdline`) is unchanged and intentionally out of scope, per the issue — the conformance runner only accepts auth via the URL, and fixing that needs upstream header-based auth support. Left as a script comment.

## Proof it works

- `tests/unit/scripts.test_conformance.test.ts` (new): runs the real script end-to-end against a stubbed `pnpm` on `PATH`, asserting neither the raw nor percent-encoded token appears in output. 5 cases: successful run, a token containing a literal backslash, modern-suite-only failure, legacy-suite-only failure, and both-fail (proving the existing modern-status-wins precedence still holds, and that redaction actually fires rather than just checking token absence).
- `pnpm run type-check`, `pnpm run lint`, `pnpm run format`, `pnpm run test:unit` (98 files / 1487 tests) all pass.
- Manually reproduced the pre-fix bug and confirmed the fix by sourcing `redact_tokens()` standalone: a token containing `\n` leaked through the original `-v`-based version and is correctly redacted after switching to `ENVIRON`.

## AI assistance

Investigated, implemented, and tested with Claude Code, including an automated multi-perspective code review (correctness, project-standards, testing, security personas plus an independent cross-model adversarial pass) that caught the `ENVIRON`/`PIPESTATUS` issues above before this PR was opened. I reviewed and verified the changes and this description myself.


MQ37 (approved):
Thank you for this PR - I remove the duplicate export and removed the unit tests as the unit test should be realted to the MCP server only and not testing the testing conformance script.

PR #1326, opened 2026-09-03
fix: Gate call-actor mentions when the tool is absent from the session

MQ37:
Closes apify/ai-team#232

Stacked on #1334 — base branch is that PR's, not master; diff here is just this PR's own commits. Server-instructions.ts's own gating logic moved to #1334 so it's reviewable and testable on its own.

## What
Gates every unconditional `call-actor` (and `apify/rag-web-browser`/`apify/web-fetch`) mention behind the session's actual tool set, in both protocol eras.

## Why
A session whose `?tools=`/`?actors=` selection omits `call-actor` still got told to use it — in dedicated Actor tool descriptions, and (via #1334) in the server-wide instructions block. A tool name the client never received in `tools/list` invites a hallucinated call.

- `actor_tools_factory.ts`: dedicated Actor tool descriptions now gate the call-actor line via the existing `hasTool` pattern used everywhere else in this codebase.
- Legacy/stdio protocol: passes the real, final per-session tool set (unchanged mechanism, already proven by report-problem gating) into #1334's `getServerInstructions`.
- Stateless (2026-07-28) protocol: resolves call-actor presence from the request URL with zero network calls — its presence is a pure function of the `?tools=`/`?actors=` selector list, never auto-injected, mode-independent, and untouched by the async Actor-metadata fetch. Falls back to today's behavior when no URL is supplied — additive, non-breaking for existing callers.
- `actorNameToToolName` also now normalizes `user~actor` selectors to `user/actor` before parsing and hashing, so a tilde-form Actor selector resolves to the same tool name (and cache key) as its slash form instead of a separate, divergent one.

## Testing
- New regression test pins the exact Claude-connector tool surface (19 tools, no `call-actor`) and asserts the resolved instructions never mention it — offline, no fixture, no live token.
- Fixed a test gap: the Actor-tool case in `tools.mode_contract.test.ts` used a hand-written stub that never exercised the real factory output, so it didn't catch this class of bug.
- New `isDeploymentTest: true` integration case pins the same 19-tool selection end-to-end: `call-actor` and `report-problem` both absent, even with telemetry enabled — runs against internal's live staging/prod on every release.
- `pnpm run type-check / lint / test:unit / format / check:agents` all green.

### DIFF
--- a/src/dev_server.ts
+++ b/src/dev_server.ts
@@ -197,7 +197,7 @@ async function serveStatelessRequest(req: Request, res: Response, taskStore: InM
             if (!isDiscoverProbe) {
                 await mcpServer.loadToolsFromUrl(req.url, new ApifyClient({ token: apifyToken }));
             }
-            return createStatelessServer(mcpServer);
+            return createStatelessServer(mcpServer, req.url);
         },
         {
             legacy: 'reject',
--- a/src/mcp/AGENTS.md
+++ b/src/mcp/AGENTS.md
@@ -25,7 +25,7 @@ Two MCP protocol revisions are served, each by its own adapter:
 - `legacy_server.ts` — package-private v1 SDK adapter for handlers, Tasks, errors,
   notifications, logging, and transport lifecycle. It reads shared state through
   `LegacyMcpServerHost`.
-- `stateless_server.ts` — `createStatelessServer(host)`: the 2026-07-28 (v2 SDK) adapter,
+- `stateless_server.ts` — `createStatelessServer(host, requestUrl?)`: the 2026-07-28 (v2 SDK) adapter,
   one `Server` per request, reading shared state through `StatelessMcpServerHost`. Serves
   `tools/*`, `resources/*` and `prompts/*`; registers no Tasks (the SDK answers
   method-not-found) and declares no `logging`.
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -19,6 +19,7 @@ import { createResourceService } from '../resources/resource_service.js';
 import type { AvailableWidget } from '../resources/widgets.js';
 import { resolveAvailableWidgets } from '../resources/widgets.js';
 import { getTelemetryEnv } from '../telemetry.js';
+import { actorNameToToolName } from '../tools/actor_tool_naming.js';
 import type {
     ActorsMcpServerOptions,
     ActorStore,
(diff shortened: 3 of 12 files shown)


jirispilka:
1. **The instructions did not match the available tools**

   Example:

   ```text
   ?ui=apps&tools=apify/rag-web-browser
   ```

   The server provided these tools:

   ```text
   apify--rag-web-browser
   get-actor-run
   get-dataset-items
   get-key-value-store-record
   abort-actor-run
   get-actor-run-widget
   ```

   But the instructions were created as if only `apify--rag-web-browser` existed. Because of that, they did not tell the model that `get-actor-run-widget` tracks progress by itself and should not be followed by more status checks.

   The new `resolveToolNamesFromInput()` helper now uses the same rules as the real tool loader. The available tools and the instructions now match.

2. **Actor IDs were incorrectly treated as tool names**

   Example:

   ```text
   ?actors=3ox4R101TgZz67sLr
   ```

   This ID belongs to `apify/rag-web-browser`. The server loads the correct tool, `apify--rag-web-browser`, after looking up the ID.

   The instruction code cannot perform that lookup, so it previously used the ID itself as a tool name. This was incorrect and also produced a warning on every request.

   The code now converts only values written as `username/name` or `username~name`:

   ```ts
   if (actorName.indexOf('/') > 0 || actorName.indexOf('~') > 0) {
       toolNames.add(actorNameToToolName(actorName));
   }
   ```

   The shared naming helper normalizes `~` to `/`, so both `apify/rag-web-browser` and `apify~rag-web-browser` resolve to `apify--rag-web-browser`.

   A current-user name such as `~my-actor` remains unresolved because its username is not available locally.

   An Actor selected by ID still gets the normal run and result tools. Its Actor-specific instructions remain absent because learning its name would require an API request.

3. **Actor tools kept unnecessary data in memory**

   The new description function used the full Actor definition even though it needed only the Actor name and short description. This kept the README and other unused data in memory until the session ended.

   The function now keeps only the two values it needs:

   ```ts
   const { actorFullName, description: actorDescription } = definition;
   ```

   The text shown to the model stays exactly the same, while unused Actor data can be released from memory.

I also added regression tests and updated the documented `createStatelessServer(host, requestUrl?)` signature. All checks pass.

MQ37:
Verified — all three already in the branch (checked against the current tip):

1. `resolveToolNamesFromInput()` wired into `getStatelessServerInstructions` (`tools_loader.ts`, `mcp/server.ts`).
2. Actor-ID-vs-name check (`indexOf('/') > 0 || indexOf('~') > 0`) in place.
3. Minimal `{ actorFullName, description: actorDescription }` destructure in `actor_tools_factory.ts`.

Plus the `createStatelessServer(host, requestUrl?)` doc/signature update. Nothing outstanding from this comment.

jirispilka (approved):

PR #1327, opened 2026-09-03
feat: Warn when a discovered Actor cannot be run in this session

MQ37:
Closes apify/ai-team#231

Stacked on #1326, which is stacked on #1334 — base branch is #1326's, not master; diff here is just this PR's own commits.

## What
`search-actors` and `fetch-actor-details` (+ widget variants) now say something when a found Actor can't actually be run in the current session — neither `call-actor` nor a dedicated tool for it is loaded.

## Why
Previously silent: the model was left to guess or hallucinate a call for an Actor that showed up in search results or details but has no runnable path in this configuration.

- `actor_tool_naming.ts`: new `canRunActor(actorId, loadedToolNames, loadedActorIds)` — true when `call-actor` is loaded or the Actor's ID is among the session's loaded Actor/Actor-MCP tools. Matches by stable Actor ID, not by reconstructing a tool-name string, so it's immune to username-length capping and to `--` inside Actor names. Soft check for a guidance hint, not a hard gate.
- `search_actors.ts`: per-result check — the caveat fires only when at least one returned Actor actually lacks a run path, not whenever `call-actor` is absent.
- `fetch_actor_details.ts`: per-Actor guidance on the success path, mirroring `buildActorNotFoundResponse`'s existing `loadedToolNames` gating pattern for the not-found case; matches by the API-resolved Actor ID directly, so querying by ID or `username/name` resolves identically.
- Both widget variants build their own response text and needed their own gated append.

## Testing
`canRunActor` covered directly (call-actor, dedicated tool, Actor-MCP sub-tool, ID-vs-full-name), the per-result search-actors caveat (single result, mixed results, all-already-loaded), guidance-present/absent for both tools and both widget variants. `pnpm run type-check / lint / test:unit / format / check:agents` all green.

### DIFF
--- a/src/mcp/tool_dispatch.ts
+++ b/src/mcp/tool_dispatch.ts
@@ -15,7 +15,7 @@ import { logHttpError } from '../utils/logging.js';
 import { respondErrorNoTelemetry } from '../utils/mcp.js';
 import type { createProgressTracker } from '../utils/progress.js';
 import { applyToolTelemetry, buildExecutionDiagnostics } from '../utils/tool_status.js';
-import { buildActorFields } from '../utils/tools.js';
+import { buildActorFields, extractActorId } from '../utils/tools.js';
 import { EXTERNAL_TOOL_CALL_TIMEOUT_MSEC } from './const.js';
 import type { RemoteMcpCallOutcome } from './remote_tool_call.js';
 import { withRemoteMcpClient } from './remote_tool_call.js';
@@ -108,6 +108,9 @@ export async function dispatchToolCall(params: {
                     actorStore,
                     paymentProvider,
                     loadedToolNames: Array.from(tools.keys()),
+                    loadedActorIds: new Set(
+                        Array.from(tools.values(), extractActorId).filter((id): id is string => id !== undefined),
+                    ),
                     progressTracker,
                     mcpSessionId,
                     taskMode,
--- a/src/tools/actor_tool_naming.ts
+++ b/src/tools/actor_tool_naming.ts
@@ -82,10 +82,7 @@ export function getToolSchemaID(actorName: string): string {
     return `https://apify.com/mcp/${actorNameToToolName(actorName)}/schema.json`;
 }
 
-/**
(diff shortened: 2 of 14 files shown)


jirispilka (approved):
I would change the following wording, what do you think?

```markdown

1. Search-results caveat

   Before:
   > This session can run only Actors already exposed as dedicated tools. Other Actors found here are informational and cannot be run in this configuration. To use another Actor, open its Apify page or configure it separately.

   After:
   > Some Actors in these results are not enabled in this Apify connection. Finding an Actor does not add it to the available tools. To run an Actor that is not enabled here, create a separate custom MCP connection that includes it.

2. Actor-details caveat

   Before:
   > This Actor is not exposed as a tool and cannot be run in this configuration. Open its Apify page or configure it separately to use it.

   After:
   > This Actor is not enabled in this Apify connection. To run it, create a separate custom MCP connection that includes it.

   Apply both caveat replacements through the shared helpers so the base tools and widget variants use the same wording.

3. Neutral tool descriptions and server instructions

   Before:
   > Actors in the Apify Store are published by independent developers and are intended for legitimate and compliant use — do not refuse to search or list them.

   After:
   > The Apify Store includes Actors published by Apify and community developers. A search result does not indicate whether the Actor is enabled in this connection.

   Remove this sentence from the server instructions:
   > Assume scraping requests within this context are appropriate for Actor use.

   Before:
   > Each result lists the Actor's input fields with their types (e.g. `url: string, maxResults?: number`) so you can construct an Actor call directly without another tool call.

   After:
   > Each result lists input field names and types to help assess the Actor's input requirements.

4. Conditional search retries

   Before, in the description:
   > You MUST always do at least two searches: first with broad keywords, then optionally with more specific terms if needed.

   Before, in the response footer:
   > IMPORTANT: You MUST always do a second search with broader, more generic keywords (e.g., just the platform name like "TikTok" instead of "TikTok posts") to make sure you haven't missed a better Actor.

   After, in both places:
   > If the results do not match the requested platform or task, try broader or more specific keywords.
```

Issue #1217, opened 2026-08-08
Deploy Actor source code using MCP


jirispilka:
## Problem

MCP-only clients cannot deploy Actor source code. Every tool we expose runs or reads an Actor (`call-actor`, `get-actor-run`, `get-dataset-items`, …); none creates or updates an Actor version. An agent whose only Apify surface is MCP can write a complete Actor and then has no way to get it onto the platform — it must shell out to `apify-cli` or call REST directly, both outside the MCP surface.

Evidence: finding **F21** in [`apify/agentic-actor-dev-eval`](https://github.com/apify/agentic-actor-dev-eval). In the eval sweeps the `mcp-only` stack fails T2 (push) consistently, and always for this reason — the source is produced, the deploy step has no tool to call. Stacks with the CLI available pass the same task.

Scope note: finding **F38** was originally filed as "template scaffolding via MCP" (#1037, closed `not_planned`). Scaffolding a fresh Actor from a template is served client-side by the public [`apify/actor-templates`](https://github.com/apify/actor-templates) manifest and raw file URLs. F38 was folded into F21: the missing piece is upload, not scaffolding.

## Dependency: `apify-core`

Blocked on [`apify/apify-core#29044`](https://github.com/apify/apify-core/issues/29044) — the REST deploy contract decision. Two endpoints currently compete for this job:

| Endpoint | Body | State |
|---|---|---|
| `PUT /v2/acts/{id}/versions/{ver}/source-files` | tarball | What the docs describe. 4xx's regardless of payload. |
| `PUT /v2/acts/{id}/versions/{ver}` | JSON `sourceFiles` array | What works today. Undocumented. `apify push` uses it internally. |

Whatever we build wraps whichever one the platform team designates. Open questions for #29044:

- Which endpoint is the supported contract for source upload.
- Whether the tarball path gets fixed, or the docs get corrected to the JSON path.
- Payload limits: max total source size, per-file size, file count.
- Binary file handling on the chosen path.
- Error semantics for a version that already exists, and for a token without write scope on the target account.

Related but out of scope: [`apify/apify-core#28933`](https://github.com/apify/apify-core/issues/28933) requested a REST endpoint for templates. That was the scaffolding path, dropped with #1037.

Solution design is open — this issue is the problem statement and its blocker.

## Cross-refs

- [`apify/apify-core#29044`](https://github.com/apify/apify-core/issues/29044) — blocker
- #1048 — placeholder PR
- #1037 — closed, scaffolding
[...]

PR #1331, opened 2026-09-04
feat: Add get-actor-build and get-actor-build-log tools

### LINKED ISSUE #1352: Add get-actor-build and get-actor-build-log tools
Part of #1351.

## Problem

Over MCP there is no observability of Actor builds: no way to read a build's state (running, succeeded, failed) and no way to read its log. Once a build has been started, an agent cannot tell whether it finished or why it failed, so a failed platform build is undebuggable without the CLI or Apify Console.

## Proposal

Two read-only tools in a new opt-in `builds` category, mirroring the run tools:

- `get-actor-build`: build status (ID, Actor ID, build number, status, timestamps) with `waitSecs` (0 to 45, default 30) like `get-actor-run`, plus a one-line summary and next step.
- `get-actor-build-log`: the build log with the same `lines` parameter as `get-actor-log` (default 10, max 50, 0 returns the whole log).

`get-actor-log` stays unchanged.

## Acceptance

- Both tools registered under `builds`, not enabled by default, not auto-injected.
- Not-found handling and response shapes match the run tools.
- Unit tests for wait forwarding, log slicing, and next-step gating.

Draft PR: #1331

DaveHanns:
## What
Two read-only tools in a new opt-in `builds` category (folder `src/tools/builds/`):
- `get-actor-build`: a build's status, mirroring `get-actor-run` (`waitSecs` 0–45, default 30; summary and next step). Like `get-actor-run`, an unfinished build is waited for while its status changes go out as MCP progress notifications, which also keep clients from timing out; the tool is in the engine's progress opt-in set.
- `get-actor-build-log`: a build's log, mirroring `get-actor-log` (`lines` 0–50, default 10; 0 returns the whole log).

`get-actor-log` is unchanged.

## Why
Over MCP there is no way to see whether a build finished or why it failed. Closes #1352. Part of Epic #1351. Relates to #1217.

## How
- `client.build(id).get({ waitForFinish })` and `client.build(id).log().get()`; same not-found shape as the run tools.
- `build_helpers.ts` holds the allowlisted build result, the shared `waitSecs` field (0–45, default 30 like the run tools), the shared build response (JSON, summary with one next step, Console link) that `build-actor` and `push-actor` reuse, and the next-step text gated on the session's tools with fallbacks. The Console link follows the existing UI-token policy and uses the build number, which is what Console resolves (a build ID in the path 404s).
- Registered under `builds`: not on by default, not auto-injected. README and `src/tools/AGENTS.md` updated. Categories are plural nouns for the resource their tools touch; `push-actor` (#1333) gets its own `versions` category.
- `ProgressTracker` gained `startActorBuildUpdates`, sharing the polling loop of the run variant (extracted in its own commit).

## Testing
Unit tests cover the waits, log slicing, not-found, gating, and schema conformance. `type-check`, `lint`, `format`, `test:unit`, `check:agents` pass. Integration and mcpc runs are not done yet (need a token).

## Notes
- Adds `structuredContent` shapes the hosted server consumes: `{ build }` and `{ log }`.
- Both tools are `paymentRequired: true` like the run tools and listed in `SKYFIRE_ENABLED_TOOLS`.
- Open: `HELPER_TOOLS` key `ACTOR_BUILD_LOG` (singular) vs `ACTOR_BUILDS_LOG`.
- The run-side counterpart was renamed to `get-actor-run-log` (#1375, merged), so the pairs read `get-actor-run` / `get-actor-run-log` and `get-actor-build` / `get-actor-build-log`.

### DIFF
--- a/src/tools/deploy/get_actor_build.ts
+++ b/src/tools/deploy/get_actor_build.ts
@@ -21,7 +21,7 @@ const getActorBuildArgs = z.object({
         .min(0)
         .max(50)
         .default(20)
-        .describe('Number of trailing log lines to include; 0 returns no log'),
+        .describe('Output the last N lines of the build log; pass 0 to return the entire log'),
 });
 
 /**
@@ -32,6 +32,7 @@ async function fetchLogTail(client: ApifyClient, buildId: string, lines: number)
     const log = await client.build(buildId).log().get();
     if (!log) return [];
     // Logs from the API end with a newline; drop it so the tail slice counts only content lines.
+    // slice(-0) is slice(0), so lines 0 returns the whole log, the same as get-actor-log.
     return log.replace(/\n$/, '').split('\n').slice(-lines);
 }
 
@@ -85,7 +86,7 @@ USAGE EXAMPLES:
         if (!build) {
             return respondUserError(`Build with ID '${parsed.buildId}' not found.`);
         }
-        const logTail = parsed.lines > 0 ? await fetchLogTail(client, parsed.buildId, parsed.lines) : [];
+        const logTail = await fetchLogTail(client, parsed.buildId, parsed.lines);
         const linkContext = await getConsoleLinkContext(apifyToken, client);
         const structuredContent = { build: toBuildResult(build, linkContext), logTail };
         const summary = `Build ${build.buildNumber} of Actor ${build.actId} is ${build.status}.`;
--- a/src/tools/deploy/get_actor_build.ts
+++ b/src/tools/deploy/get_actor_build.ts
@@ -1,0 +1,102 @@
+import type { Build } from 'apify-client';
+import { z } from 'zod';
+
+import type { ApifyClient } from '../../apify_client.js';
+import { HELPER_TOOLS } from '../../const.js';
+import type { InternalToolArgs, ToolEntry, ToolInputSchema } from '../../types.js';
+import { TOOL_TYPE } from '../../types.js';
+import { compileSchema, fixZodSchemaRequired } from '../../utils/ajv.js';
+import { getConsoleLinkContext } from '../../utils/console_link.js';
+import { respondOk, respondUserError } from '../../utils/mcp.js';
+import { TERMINAL_RUN_STATUSES } from '../../utils/progress.js';
+import { apifyConsoleLinkText } from '../storage/storage_helpers.js';
+import { getActorBuildToolOutputSchema } from '../structured_output_schemas.js';
+import { toBuildResult } from './build_helpers.js';
+
+const getActorBuildArgs = z.object({
+    buildId: z.string().min(1).describe('Build ID, as returned when a build is started'),
+    lines: z
+        .number()
+        .int()
+        .min(0)
+        .max(50)
+        .default(20)
+        .describe('Number of trailing log lines to include; 0 returns no log'),
+});
+
+/**
+ * https://docs.apify.com/api/v2/actor-build-log-get
+ *  /v2/actor-builds/{buildId}/log
+ */
+async function fetchLogTail(client: ApifyClient, buildId: string, lines: number): Promise<string[]> {
+    const log = await client.build(buildId).log().get();
+    if (!log) return [];
+    // Logs from the API end with a newline; drop it so the tail slice counts only content lines.
+    return log.replace(/\n$/, '').split('\n').slice(-lines);
+}
+
+function buildNextStep(build: Build, loadedToolNames: readonly string[]): string {
+    if (build.status === 'SUCCEEDED') {
+        return loadedToolNames.includes(HELPER_TOOLS.ACTOR_CALL)
+            ? `Run the Actor with ${HELPER_TOOLS.ACTOR_CALL} and set callOptions.build to ${build.buildNumber}.`
+            : 'The build is ready to run.';
+    }
+    if (TERMINAL_RUN_STATUSES.has(build.status)) {
+        return 'Read the log tail for the error, fix the source, and build again.';
+    }
+    return 'Call this tool again in about 10 seconds.';
+}
+
+/**
+ * https://docs.apify.com/api/v2/actor-build-get
+ *  /v2/actor-builds/{buildId}
+ */
+export const getActorBuild: ToolEntry = Object.freeze({
+    type: TOOL_TYPE.INTERNAL,
+    name: HELPER_TOOLS.ACTOR_BUILD_GET,
+    title: 'Get Actor build',
+    description: `Get the status of an Actor build and the last lines of its build log.
+Read-only. Returns the build (id, actorId, buildNumber, status, startedAt, finishedAt), the log tail,
+and a summary with one next step.
+
+USAGE:
+- Use to check whether a build has finished.
+- Use when a build FAILED to read the error from the log tail before fixing the source.
+
+USAGE EXAMPLES:
+- user_input: Did build 7aB3xYz9Kq finish?
+- user_input: Why did build 7aB3xYz9Kq fail?`,
+    // `fixZodSchemaRequired` strips `lines` from `required` because it has a default.
+    inputSchema: fixZodSchemaRequired(z.toJSONSchema(getActorBuildArgs)) as ToolInputSchema,
+    outputSchema: getActorBuildToolOutputSchema,
+    ajvValidate: compileSchema(z.toJSONSchema(getActorBuildArgs)),
+    paymentRequired: true,
+    annotations: {
+        title: 'Get Actor build',
+        readOnlyHint: true,
+        destructiveHint: false,
+        idempotentHint: true,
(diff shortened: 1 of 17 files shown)


src/utils/console_link.ts
```diff
@@ -58,6 +58,15 @@ export function buildConsoleRunUrl(context: ConsoleLinkContext | undefined, runI
     return buildConsoleUrl(context, `/actors/runs/${runId}`);
 }
 
+/** Builds the Console build detail URL: `<consoleBaseUrl>[/organization/<orgId>]/actors/<actorId>/builds/<buildId>`. */
+export function buildConsoleBuildUrl(
+    context: ConsoleLinkContext | undefined,
+    actorId: string,
+    buildId: string,
+): string | undefined {
+    return buildConsoleUrl(context, `/actors/${actorId}/builds/${buildId}`);
```
l2ysho, line 67: I am not sure about `buildId` because in cli we use `buildNumber`, probably worth to doublecheck

https://github.com/apify/apify-cli/blob/109db5598bc2379242a47395afb7f3194adcdaee/src/commands/builds/info.ts#L107

l2ysho (approved):
GJ 👍 
approved, one comment left to verify

src/utils/console_link.ts
jirispilka: Yes, correct, this link does not exist: https://console.apify.com/actors/BEwUoGEsRUQC2iUPa/builds/IQ3GsGDwcre6IMg7K

Only works with `buildNumber`. Isn't this actually console fix, to accept `.../builds/<build-id>` or we have `buildNumber` available there so the fix is cheap

src/tools/registry.ts
```diff
@@ -61,6 +63,7 @@ export const toolCategories = {
         getKeyValueStoreList,
     ],
     tasks: [createActorTask, getActorTask, updateActorTask, publishActorTask, unpublishActorTask],
+    deploy: [getActorBuild, getActorBuildLog],
```
jirispilka, line 71: I'm thinking aloud here. I'm not really happy with the deploy category as we don't have it in the API at all.

But at the same time I don't have better option, it could be separated to `builds` and `versions` (API vocabulary for builds and `actor push`). 

`deploy` is generic enough. But we should keep the convention of nouns within the categories while `deploy` is a verb. Should it be `deployment`?

src/tools/deploy/get_actor_build.ts
```diff
@@ -0,0 +1,86 @@
...
+        });
+        const consoleLinkText = apifyConsoleLinkText(structuredContent.build.apifyConsoleUrl);
+        return respondOk(
+            [
+                JSON.stringify(structuredContent),
+                `${summary}\n${nextStep}`,
+                ...(consoleLinkText ? [consoleLinkText] : []),
+            ],
+            { structuredContent },
+        );
```
jirispilka, line 84: Consider this comment from Claude: 

`get_actor_build.ts` currently formats the build response inline:
```ts
  const consoleLinkText = apifyConsoleLinkText(structuredContent.build.apifyConsoleUrl);

  return respondOk(
      [
          JSON.stringify(structuredContent),
          `${summary}\n${nextStep}`,
          ...(consoleLinkText ? [consoleLinkText] : []),
      ],
      { structuredContent },
  );
```
  This same three-step pattern is being duplicated in #1332 and #1333, and #1380 would add another copy. That makes it easy for the build tools to drift in content ordering or Console-link handling.

  Please extract this into build_helpers.ts, alongside toBuildResult and buildNextStepForBuild:
```ts
  export function respondWithBuild(params: {
      structuredContent: Record<string, unknown>;
      summary: string;
      nextStep: string;
  }) {
      const consoleLinkText = apifyConsoleLinkText(
          params.structuredContent.build.apifyConsoleUrl,
      );

      return respondOk(
          [
              JSON.stringify(params.structuredContent),
              `${params.summary}\n${params.nextStep}`,
              ...(consoleLinkText ? [consoleLinkText] : []),
          ],
          { structuredContent: params.structuredContent },
      );
  }
```
  Then get_actor_build.ts only needs:
```ts
  return respondWithBuild({
      structuredContent,
      summary,
      nextStep,
  });
```
  This keeps the response shape and Console-link behavior consistent across all build-related tools.

src/tools/deploy/get_actor_build.ts
```diff
@@ -0,0 +1,86 @@
...
+import { compileSchema, fixZodSchemaRequired } from '../../utils/ajv.js';
+import { getConsoleLinkContext } from '../../utils/console_link.js';
+import { respondAborted, respondOk, respondUserError } from '../../utils/mcp.js';
+import { ABORT, raceAbort, WAIT_SECS_MAX } from '../actors/actor_run_response.js';
+import { apifyConsoleLinkText } from '../storage/storage_helpers.js';
+import { getActorBuildToolOutputSchema } from '../structured_output_schemas.js';
+import { buildNextStepForBuild, toBuildResult } from './build_helpers.js';
+
+/** Default `waitSecs` for `get-actor-build`. Intentionally non-zero so polling callers wait briefly by default. */
+const WAIT_SECS_DEFAULT = 30;
```
jirispilka, line 16: Consider this comment from Claude: 

`waitSecs` is declared three times across the stack with two different defaults — 30 here,
`.default(WAIT_SECS_MAX)` in `build-actor` and `push-actor` — and only this one documents the cap and default in its `.describe()`. An agent that calls `build-actor` then `get-actor-build` gets 45 then 30 for no stated reason.

Worth a `waitSecsField(defaultSecs)` in `build_helpers.ts` and one decision on the family default, made here since this is the first tool to set one.

jirispilka (approved):
Nice work! 💪🏻 

🙏🏻 And I'm really sorry for a delayed review that caused you some merge conflicts to solve.

There are some suggestions in the code: pre-approving.

Comments below are related to this and mostly to the following PRs. I haven't reviewed the next PRs yet so please take the comments with grain of salt and ignore false ones.

Two things I would open as issues rather than pile onto this PR: 

1 - build discovery (there's no way to find a failed build's ID. The run list only carries builds that ran)

2 - And we should do proper evals for all the new tools

ok, one more thing from Claude:

`get-actor-run` drives `progressTracker.startActorRunUpdates` through its wait; this doesn't use `progressTracker` at all, and neither do `build-actor`/`push-actor`. Since #1352 specifies these as "mirroring `get-actor-run`", is emitting progress during the wait in scope

It matters most in the two later PRs, where `waitSecs` defaults to the 45s cap and platform builds routinely exceed it — so the common outcome is a full silent wait ending in "still running". Progress notifications also reset MCP client timeouts. A `startBuildUpdates` alongside `startActorRunUpdates` would cover all four tools; it
naturally lives in `build_helpers.ts`, which is why I'm raising it here rather than on #1332.

@DaveHanns can u please create the issues or should I do it?

jirispilka:
One more thing when I tested the follow up PR.

Please fix the response text: `The build is ready to run.`. This sounds awkward. Could it be: `The Actor with build 0.2.1 is ready to run?`

```
Build 0.2.1 of Actor qGXMy0NAkWsIIb9LZ is SUCCEEDED.
The build is ready to run.
```

src/utils/console_link.ts
DaveHanns: Fixed in 8eeba19: the link uses the build number now; Console does not resolve build IDs there.

src/tools/registry.ts
DaveHanns: Renamed in a838b8b: `builds` for the three build tools, and on #1333 `versions` for `push-actor`, with the folders to match.

src/tools/deploy/get_actor_build.ts
DaveHanns: Extracted in 8eeba19 as `respondWithBuild` in `build_helpers.ts`; `build-actor` and `push-actor` use it too.

src/tools/deploy/get_actor_build.ts
DaveHanns: Done in 8eeba19: `buildWaitSecsField` in `build_helpers.ts`, default 30 s like `get-actor-run` and `call-actor`, cap documented in one place.

DaveHanns:
Thanks for the review. Follow-ups:

- Response text: fixed in 8eeba19, the next step now reads "The Actor is ready to run with build 0.2.1."
- The two issues exist: #1407 (list an Actor's builds, so a failed build's ID can be found) and #1408 (evals for the build and push tools), both under #1351.
- Progress notifications: implemented in 696be70 here and in the follow-up PRs, mirroring `get-actor-run`. An unfinished build is waited for while its status changes go out as `notifications/progress`; `get-actor-build`, `build-actor` and `push-actor` joined the engine's opt-in set. `ProgressTracker` got `startActorBuildUpdates`, sharing the polling loop of the run variant.
- Category: renamed to `builds`, with `versions` for `push-actor` on #1333, folders to match.

PR #1332, opened 2026-09-04
feat: Add build-actor tool

### LINKED ISSUE #1353: Add build-actor tool
Part of #1351.

## Problem

After pushing source, an agent has no way to start a platform build over MCP.

## Proposal

`build-actor` in the `builds` category: start a build of an Actor version (`actor` as ID or `username/name`; optional `versionNumber`, `tag`, `useCache`), wait up to 45 seconds, and return the build in the same shape as `get-actor-build` with a next step (run it, keep polling, or read the build log). When `versionNumber` is omitted and the Actor has exactly one version, use it; with several, return the list.

## Acceptance

- Registered under `builds`, opt-in.
- Uses `client.actor(id).build(version, { tag, useCache, waitForFinish })`.
- Unit tests for version resolution, option forwarding, and next-step gating.

Draft PR: #1332

DaveHanns:
## What
`build-actor`: start a platform build of an Actor version and wait up to `waitSecs` for it (default 30 s, cap 45, the run tools' values). Returns the build in the same shape as `get-actor-build`, with a next step (run it, keep polling, or read the build log).

## Why
Nothing over MCP can start a build; `callOptions.build` on `call-actor` only picks an existing one. Needed for Git-sourced Actors, rebuilds, and retagging. Closes #1353. Part of Epic #1351.

## How
- `client.actor(id).build(versionNumber, { tag, useCache })`, then `client.build(id).get({ waitForFinish })`, through the shared `startBuild` helper. The wait is raced against the request abort signal, and a cancelled request aborts the build it started, the way `call-actor` aborts its run, so nothing keeps building for nobody.
- `destructiveHint: true`: `tag` re-points an existing tag such as `latest` to the new build, which changes what runs by default. `paymentRequired: true` and listed in `SKYFIRE_ENABLED_TOOLS`.
- `versionNumber` optional: one version means use it; several means an error listing them.
- Registered under `builds` after the build inspection tools. Next-step text comes from the shared helper, gated on the session's tools with fallbacks.
- While it waits, the build's status changes go out as progress notifications through the shared `waitForBuild`, the same as `call-actor` does for its run; `build-actor` is in the engine's progress opt-in set.

## Testing
Unit tests cover version resolution, option forwarding, the aborted paths (before the wait, during it, and with a failing abort call), `waitSecs: 0`, and every next-step branch. `type-check`, `lint`, `format`, `test:unit`, `check:agents` pass. Integration and mcpc runs are not done yet (need a token).

## Notes
- Stacked on #1331 (base `feat/get-actor-build`); only this PR's diff shows.
- Adds a `structuredContent` shape the hosted server consumes.
- Open: `HELPER_TOOLS` key `ACTOR_BUILD` vs a longer name; a finished build's hint names the build number, not the tag, because build tags update asynchronously.

### DIFF
--- a/src/tools/deploy/build_actor.ts
+++ b/src/tools/deploy/build_actor.ts
@@ -1,0 +1,130 @@
+import type { Build } from 'apify-client';
+import { z } from 'zod';
+
+import { HELPER_TOOLS } from '../../const.js';
+import type { InternalToolArgs, ToolDescriptionContext, ToolEntry, ToolInputSchema } from '../../types.js';
+import { ALL_TOOLS_PRESENT, TOOL_TYPE } from '../../types.js';
+import { compileSchema, fixZodSchemaRequired } from '../../utils/ajv.js';
+import { getConsoleLinkContext } from '../../utils/console_link.js';
+import { respondOk, respondUserError } from '../../utils/mcp.js';
+import { TERMINAL_RUN_STATUSES } from '../../utils/progress.js';
+import { WAIT_SECS_MAX } from '../actors/actor_run_response.js';
+import { apifyConsoleLinkText } from '../storage/storage_helpers.js';
+import { buildActorToolOutputSchema } from '../structured_output_schemas.js';
+import { startBuild, toBuildResult } from './build_helpers.js';
+
+const buildActorArgs = z.object({
+    actor: z.string().min(1).describe('Actor ID or username/name'),
+    // Format is not enforced by a regex: AJV here drops `pattern`, and the lookup against the Actor's
+    // versions below rejects anything that is not an existing MAJOR.MINOR version with a soft fail.
+    versionNumber: z
+        .string()
+        .optional()
+        .describe('Version to build in MAJOR.MINOR form; defaults to the only version when the Actor has exactly one'),
+    tag: z.string().optional().describe('Build tag to assign, for example latest'),
+    useCache: z.boolean().default(true).describe('Reuse the Docker layer cache from the previous build'),
+    waitSecs: z
+        .number()
+        .int()
+        .min(0)
+        .max(WAIT_SECS_MAX)
+        .default(WAIT_SECS_MAX)
+        .describe('How long to wait for the build to finish before returning its current status'),
+});
+
+function buildDescription({ hasTool }: ToolDescriptionContext): string {
+    return `Build an Actor version on the Apify platform and wait a bounded time for the build to finish.
+Returns the build (id, actorId, buildNumber, status, startedAt, finishedAt) and a summary with one next step.
+Use after the Actor's source changed so runs pick up the new code.${
+        hasTool(HELPER_TOOLS.ACTOR_BUILD_GET)
+            ? ` If the build is still running when the wait ends, check it with ${HELPER_TOOLS.ACTOR_BUILD_GET}.`
+            : ''
+    }
+
+USAGE:
+- Use to rebuild an Actor after changing its source.
+- Use to build a specific version and tag it, for example as latest.
+
+USAGE EXAMPLES:
+- user_input: Rebuild my-actor
+- user_input: Build version 0.2 of john/my-actor and tag it latest`;
+}
+
+function buildNextStep(build: Build, tag: string | undefined, loadedToolNames: readonly string[]): string {
+    if (build.status === 'SUCCEEDED') {
+        return loadedToolNames.includes(HELPER_TOOLS.ACTOR_CALL)
+            ? `Run the Actor with ${HELPER_TOOLS.ACTOR_CALL} and set callOptions.build to ${tag ?? build.buildNumber}.`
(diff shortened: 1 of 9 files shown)


src/tools/deploy/build_helpers.ts
```diff
@@ -24,17 +30,39 @@ export function toBuildResult(build: Build, linkContext: ConsoleLinkContext | un
...
+    const { tag, useCache, waitSecs, signal } = options;
+    return await raceAbort(
+        client.actor(actorId).build(versionNumber, {
+            ...(tag !== undefined && { tag }),
+            useCache,
+            waitForFinish: waitSecs,
+        } satisfies ActorBuildOptions),
+        signal,
+    );
+}
```
l2ysho, line 52: Q: Do we need to manually cancel build when aport signal arrives?

src/tools/deploy/build_actor.ts
```diff
@@ -0,0 +1,116 @@
...
+    type: TOOL_TYPE.INTERNAL,
+    name: HELPER_TOOLS.ACTOR_BUILD,
+    title: 'Build Actor',
+    description: buildDescription(ALL_TOOLS_PRESENT),
+    buildDescription,
+    // `fixZodSchemaRequired` strips `useCache` and `waitSecs` from `required` because they have defaults.
+    inputSchema: fixZodSchemaRequired(z.toJSONSchema(buildActorArgs)) as ToolInputSchema,
+    outputSchema: buildActorToolOutputSchema,
+    ajvValidate: compileSchema(z.toJSONSchema(buildActorArgs)),
+    paymentRequired: true,
```
l2ysho, line 65: Q: should this be paid? I am comparing with `SKYFIRE_ENABLED_TOOLS`

l2ysho (commented):
Seems fine, left 2 comments when I am not sure, but at least raceAbort one looks valid to me.

src/tools/deploy/build_helpers.ts
jirispilka: Good question, I think the answer is a proper abort-actor-build tool.

src/tools/deploy/build_actor.ts
jirispilka: Keep paymentRequired: true both skyfire.ts and x402.ts gate on that flag, and for a tool that burns
some money it is the right thing to do.

@DaveHanns you need to add the tools into SKYFIRE_ENABLED_TOOLS

src/tools/deploy/build_actor.ts
```diff
@@ -0,0 +1,116 @@
...
+    annotations: {
+        title: 'Build Actor',
+        readOnlyHint: false,
+        destructiveHint: false,
```
jirispilka, line 69: Claude: destructiveHint: false, but apify-client says of the tag option: "Tag to be applied to the build
(e.g. 'latest', 'beta'). Existing tag with the same name will be replaced."

So build-actor {actor, versionNumber:'0.1', tag:'latest'} moves `latest` off whatever build holds it
onto this one, and call-actor without an explicit callOptions.build follows the actor's default build
tag. That's a production change from one tool call.

Annotations are static per tool so we can't make it depend on whether tag was passed — it's true or
false for the whole tool. Given tag is in the schema, I'd say true. Clients use this to decide whether
to confirm with the user first.

jirispilka (approved):
I left two comments, otherwise it looks good. Thanks!

src/tools/deploy/build_helpers.ts
DaveHanns: Went with parity with `call-actor` in 41c99d8: a cancelled request now aborts the build it started. A standalone `abort-actor-build` tool can follow separately.

src/tools/builds/build_actor.ts
```diff
@@ -0,0 +1,116 @@
...
+    type: TOOL_TYPE.INTERNAL,
+    name: HELPER_TOOLS.ACTOR_BUILD,
+    title: 'Build Actor',
+    description: buildDescription(ALL_TOOLS_PRESENT),
+    buildDescription,
+    // `fixZodSchemaRequired` strips `useCache` and `waitSecs` from `required` because they have defaults.
+    inputSchema: fixZodSchemaRequired(z.toJSONSchema(buildActorArgs)) as ToolInputSchema,
+    outputSchema: buildActorToolOutputSchema,
+    ajvValidate: compileSchema(z.toJSONSchema(buildActorArgs)),
+    paymentRequired: true,
```
DaveHanns: Added all four tools to `SKYFIRE_ENABLED_TOOLS`: 4267e05 here, b772c19 on #1331 and e0a2fe9 on #1333.

src/tools/deploy/build_actor.ts
DaveHanns: Agreed. `destructiveHint` is true since 41c99d8, with a comment on why.

l2ysho (approved):
🚀

PR #1333, opened 2026-09-04
feat: Add push-actor tool

### LINKED ISSUE #1217: Deploy Actor source code using MCP

DaveHanns:
## What
`push-actor`, the MCP counterpart of `apify push`, in its own opt-in `versions` category (folder `src/tools/versions/`): it writes Actor versions, the API's noun for what a push creates. It creates the Actor if missing, uploads or updates a version's source files (`merge` keeps existing files, `replace` makes the version exactly the pushed set), and by default starts a build and waits up to `waitSecs` for it (default 30 s like the run tools).

## Why
Agents can write an Actor over MCP but cannot ship it. Closes #1217. Part of Epic #1351.

## How
- The documented JSON `sourceType: SOURCE_FILES` + `sourceFiles` contract through apify-client (`actors().create`, `version().update`, `versions().create`), the same calls `apify push` makes.
- `actor` takes an ID or a name, like `build-actor` and `call-actor`: `my-scraper`, `username/my-scraper`, or the API's `username~my-scraper` (own account only); the response always uses `username/my-scraper`. A name is looked up first; a bare value shaped like an Actor ID that no Actor is named after is looked up once as an ID, must be the caller's Actor, and never creates one, so a legal name of that shape (`scrapemapsresult1`) is still created. One extra API call, only in that case. Name and username-prefix rules come from `@apify/consts` (`ACTOR_NAME`, `USERNAME`), the same definitions the platform and CLI use; `versionNumber` gets an explicit `MAJOR.MINOR` check. Both run before any API call. `versionNumber` is optional and resolved like `build-actor`; a new Actor gets `0.0`.
- Validation: the path, duplicate, size and base64 checks run before any API call (relative POSIX paths only, no `..` or absolute paths, strict base64, 3 MiB total measured as the API receives it, so base64 counts its encoded text; in `merge` mode the kept plus pushed files are measured together). The `.actor/actor.json` requirement is enforced inside the push step for the branch that needs it: create Actor, create version, replace, or the merge result.
- A file with a binary extension (png, jpg, pdf, zip, fonts, archives, media) and no `encoding` must carry base64 and is stored as `BASE64`; an explicit `encoding` still wins. Found in review: a `.png` pushed as a base64 string without `encoding` was saved as text. The extensions are listed rather than taken from MIME types like the CLI does, because that classifies `.ts` as video.
- The tool file is organized as small named steps: resolve target, create Actor, resolve version, create version, update version.
- The build it starts is waited for through the shared `waitForBuild`, so its status changes go out as progress notifications; `push-actor` is in the engine's progress opt-in set.
- Steps throw `UserInputError` (new in `src/errors.ts`) when the call is rejected and `BuildStartError` when the push succeeded but the build did not start; `call()` maps both to responses in one place, next to the 403 mapping.
- A 403 becomes a clear "token lacks Actor write access" error. A build that fails to start after a successful push still returns the push result (unchanged behavior).

## Testing
Unit tests cover create, merge, replace, validation, error paths, and gating. `type-check`, `lint`, `format`, `test:unit`, `check:agents` pass. Not run against the live platform yet.

## Notes
- Stacked on #1332 (base `feat/build-actor`). Renamed from `push-actor-source` in this PR; the branch keeps its old name.
- Adds a `structuredContent` shape the hosted server consumes.
- Out of scope: `envVars`, single-file deletes, tarball uploads over 3 MiB, standby and run options.
- New direct dependency `@apify/consts` (already present transitively through apify-client).
- `paymentRequired: true` and listed in `SKYFIRE_ENABLED_TOOLS`.
- Open: `destructiveHint: true` covers both modes.


src/tools/deploy/push_actor.ts
```diff
@@ -0,0 +1,570 @@
...
+function formatActorFullName(username: string, bareName: string): string {
+    return `${username}/${bareName}`;
+}
+
+/**
+ * Splits `username/name` or `username~name` (the separator the Apify API uses) into its parts; a bare
+ * name has no `ownerPrefix`.
+ */
+function parseActorName(actorName: string): ActorNameParts {
+    const separatorIndex = actorName.search(/[/~]/);
```
l2ysho, line 161: Suggestion: I would add some easy check (regex) the actorName is really name and not actorId 🤔

l2ysho (approved):
Looks good, feel free to address my comment as you like

DaveHanns:
A PR to handle large code bases the same way CLI does: https://github.com/apify/apify-mcp-server/pull/1380

src/tools/versions/push_actor.ts
```diff
@@ -0,0 +1,570 @@
...
+function formatActorFullName(username: string, bareName: string): string {
+    return `${username}/${bareName}`;
+}
+
+/**
+ * Splits `username/name` or `username~name` (the separator the Apify API uses) into its parts; a bare
+ * name has no `ownerPrefix`.
+ */
+function parseActorName(actorName: string): ActorNameParts {
+    const separatorIndex = actorName.search(/[/~]/);
```
jirispilka: But the regex won't help here, e.g. `scrapemapsresults1` could be a name or actorId.
I'm afraid the resolution will cost us one API call.

src/tools/deploy/push_actor.ts
```diff
@@ -0,0 +1,570 @@
...
+const DEFAULT_BUILD_TAG = 'latest';
+
+/** The platform's inline source-files cutoff, the same one `apify push` uses; larger projects need the Apify CLI. */
+const MULTIFILE_SOURCE_MAX_MIB = MAX_MULTIFILE_BYTES / (1024 * 1024);
+
+/** `apify push` starts a new Actor at this version. */
+const DEFAULT_VERSION_NUMBER = '0.0';
+
+const pushActorArgs = z.object({
+    actorName: z
```
jirispilka: `build-actor` tool takes: "Actor ID or username/name", why not to do it for the push-actor as well?

src/tools/versions/push_actor.ts
```diff
@@ -0,0 +1,570 @@
...
+    /** The account the Actor lives in, in the platform's spelling. */
+    username: string;
+    bareName: string;
+    /** Undefined when the Actor does not exist yet. */
+    actor: Actor | undefined;
+};
+
+/** Looks up the caller's username and the Actor; a `username/` or `username~` prefix must name the caller's own account. */
+async function resolveTargetActor(client: ApifyClient, { ownerPrefix, bareName }: ActorNameParts): Promise<TargetActor> {
+    const { username } = await client.user('me').get();
```
jirispilka, line 254: we have this call cached here (if you want to use it) 
src/utils/userid_cache.ts -> `getUserInfoCached`. I guess this will be infrequent so just FYI

jirispilka (approved):
Looks good to me.

From the testing with Claude:

1. Binary files silently corrupted when encoding is omitted (push-actor)
- Problem: files[].encoding defaults to utf8. Pushing a binary file (e.g. a .png) without explicitly setting encoding: "base64" succeeds with no warning — the raw string is stored as literal text, silently corrupting the file.
- Evidence: pushed logo.png with a base64 string, encoding omitted → tool returned success (filesPushed: 3, no error). apify pull on the resulting Actor showed logo.png as 92-byte ASCII text, not a decoded PNG.
- Affected: src/tools/deploy/push_actor.ts (schema + validation for files[].encoding/content).

src/tools/deploy/push_actor.ts
DaveHanns: Yep, was just discussing it with Claude.

src/tools/versions/push_actor.ts
DaveHanns: Ended up with the CLI's behaviour in f126161: `actor` takes an ID or a name. A name is looked up first; a bare 17-character value that no Actor is named after is looked up once as an ID, own account only, and an ID never creates an Actor, so a legal name of that shape (`scrapemapsresult1`) is still created.

src/tools/deploy/push_actor.ts
DaveHanns: Done in f126161: the parameter is `actor`, ID or name, like `build-actor`.

src/tools/versions/push_actor.ts
DaveHanns: Checked: the cache holds the user ID, plan tier and organization flag, not the username, so `push-actor` keeps the direct call. Extending the cache would be its own change.

DaveHanns:
Binary files: fixed in fa20989. A file with a binary extension (images, PDF, archives, fonts, media, binaries) and no `encoding` must carry base64 and is stored as `BASE64`; an explicit `encoding` still wins. The extensions are listed rather than taken from MIME types like the CLI does, because `mime` classifies `.ts` as video, which would turn TypeScript sources into base64.

Also since your review: `actor` takes an ID or a name (f126161), and the tool moved to a `versions` category and folder (9e0a469).

PR #1334, opened 2026-09-04
fix: Gate every cross-tool mention in server instructions behind the session's tool set

MQ37:
Closes apify/ai-team#266 — split out of apify/ai-team#232, which #1326 (stacked on top) closes.

Split out of #1326/#1327 — this is the one self-contained piece both depend on: `server-instructions.ts`'s own gating logic, generic over any tool set, needing zero wiring changes to test.

## What
Server instructions unconditionally named `call-actor`, `apify/rag-web-browser`, `apify/web-fetch`, `report-problem`, and several apps-mode widget/disambiguation tools regardless of whether the session actually loaded them.

## Why
A tool name the client never received in `tools/list` invites a hallucinated call.

- `getServerInstructions` now takes a `ToolDescriptionContext` (the same gating mechanism already used for per-tool descriptions, e.g. report-problem's own gating) instead of a bare `reportProblemAvailable` boolean, and gates every cross-tool mention through it — call-actor, rag-web-browser, web-fetch, report-problem, plus 4 more spots found during review: the apps-mode widget-workflow block, the search-actors-vs-fetch-actor-details disambiguation, the apps-mode data-vs-widget bullets, and call-actor's own tool-dependency block naming fetch-actor-details.
- The legacy adapter's real per-connection call site (`ActorsMcpServer.getServerInstructions()`) now passes the session's actual tool set instead of a report-problem-only check.
- Two other call sites needed an explicit context because the parameter's default flipped from "nothing present" to "everything present except report-problem": the legacy constructor's placeholder instructions (always overwritten once the real tool set is known) and the stateless `server/discover` probe (configuration-level, answered before any request envelope is seen — must never show report-problem or any per-session tool). Both now pass an explicit context with every tool except report-problem present, matching the new default's own exception.
- Follow-up commit: the `## Tool dependencies and disambiguation` heading is emitted only when at least one subsection under it survives gating (an Actor-only session with no widgets ended on a bare heading), blank lines around subsections restored to the pre-gating text, and the apps-mode render with `call-actor` loaded is now covered — breaking that arm previously passed the whole suite.

## Testing
Full case suite for the function, offline (fake `hasTool` contexts, no fixture, no live token): every gate present/absent, single-tool fallbacks so a tool loaded alone is never left with zero mention, apps vs default mode. `pnpm run type-check / lint / test:unit / format / check:agents` all green.

### DIFF
--- a/src/mcp/legacy_server.ts
+++ b/src/mcp/legacy_server.ts
@@ -34,7 +34,7 @@ import {
 import log from '@apify/log';
 
 import type { ApifyClient } from '../apify_client.js';
-import { FAILURE_CATEGORY, TOOL_STATUS } from '../const.js';
+import { FAILURE_CATEGORY, HELPER_TOOLS, TOOL_STATUS } from '../const.js';
 import type { createPromptService } from '../prompts/prompt_service.js';
 import type { createResourceService } from '../resources/resource_service.js';
 import { getServerInfo } from '../server_card.js';
@@ -138,7 +138,9 @@ export class LegacyMcpServer {
                 prompts: {},
                 logging: {},
             },
-            instructions: getServerInstructions(),
+            // Placeholder, always overwritten below once the real tool set is known — matches
+            // the pre-gating default (everything but report-problem) rather than every mention.
+            instructions: getServerInstructions(undefined, { hasTool: (name) => name !== HELPER_TOOLS.PROBLEM_REPORT }),
         });
         this.setupInitializeHandler();
         this.setupLoggingProxy();
--- a/src/mcp/legacy_server.ts
+++ b/src/mcp/legacy_server.ts
@@ -138,8 +138,7 @@ export class LegacyMcpServer {
                 prompts: {},
                 logging: {},
             },
-            // Placeholder, always overwritten below once the real tool set is known — matches
-            // the pre-gating default (everything but report-problem) rather than every mention.
+            // Placeholder overwritten below once the real tool set is known; matches the pre-gating default (all but report-problem).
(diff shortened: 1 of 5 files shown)


jirispilka (approved):
Verified against a real session (`?tools=compass/crawler-google-places`), nice.

Two small things worth two fix, pre-approving.

Claude comments:
**1. Missing test: apps mode with `call-actor` loaded.** That combination is untested, and it's what most hosted apps sessions get. Proof: changing `hasCall` to `hasTool(HELPER_TOOLS.ACTOR_CALL) && !isApps` strips the widget rules, the polling line, `### Tool dependencies` and "Prefer dedicated tools" from every apps session — and the suite still passed 1595/1595. This catches it and passes as-is:

```ts
it('keeps every call-actor mention in apps mode when the session has call-actor', () => {
    const instructions = getServerInstructions(SERVER_MODE.APPS, ALL_TOOLS_PRESENT);
    expect(instructions).toContain(HELPER_TOOLS.ACTOR_CALL_WIDGET);
    expect(instructions).toContain('Polling `get-actor-run` after `call-actor` is fine');
    expect(instructions).toContain('### Tool dependencies');
    expect(instructions).toContain('Prefer dedicated tools when available');
});
```

**2. Gated sections leave visible artifacts.** Two real cases:

- `?tools=compass/crawler-google-places&ui=false` — the instructions **end** with a bare   `## Tool dependencies and disambiguation` and nothing under it (all three subsections gated away).
- apps mode — that heading is immediately followed by its bullet with no blank line; the pre-change text always had one.

One fix covers both: emit the heading only when at least one subsection renders, and start each subsection block with a blank line rather than ending the `call-actor` block with one (ending it there leaves apps-mode spacing tight, since that block is empty).

PR #1338, opened 2026-09-06
test(evals): Add the evals:coverage tool and argument coverage matrix

jirispilka:
**PR 4 of 5** for apify/ai-team#236 · base `chore/evals-port-selection-cases` (#1340) · [explainer](https://claude.ai/code/artifact/afae3b80-aeda-4045-b2f3-276db48d71fe) · [stack overview](https://claude.ai/code/artifact/11ee9b13-4894-4553-ac34-c676b2e09150)

## Why

Nobody could say which tools and which arguments the evals cover. The authoring skill mandates a coverage matrix; it existed nowhere as a checkable artifact. Tool breadth was one unit-test assertion in PR 3; argument coverage was unmeasured.

## Reading this diff

`coverage_matrix.ts` is the whole feature. `coverage_matrix.md` is its generated output — read it as the product, not as a diff. The rest is the test file, a captured fixture, one script entry and prose.

## What changed

- `pnpm run evals:coverage` reads the tool registry and each input schema, diffs them against the dataset, and writes `evals/mcp_agent/coverage_matrix.md`. Per tool: `pr` selection cases, `full` agent cases reachable and exercised, argument groups covered/uncovered, status `covered | uncovered | excluded`.
- Argument groups are top-level schema properties plus one nesting level. A nested group counts only when a case pins that child key: `expectedArgs` deep-equals the parent, so `{callOptions: {memory: 256}}` asserts the siblings are *absent* — the opposite of exercising them. Static and observed paths read nesting identically.
- The two default Actor tools build their schema at runtime, so they read `dynamic` and leave the totals. Widgets are `excluded`, never `covered`.
- Counts are per tier as well as per kind, so an off-tier item can't inflate the count its column label promises.
- Coverage comes from `expectedTools` / `expectedArgs`; agent-case reachability is informational only, so the `report-problem` gap stays visible.
- `--check` exits 1 on a stale matrix. It needs Langfuse for the dataset half, so the offline unit test asserts the weaker invariant that holds without it: exactly one row per in-scope and widget tool identifier, which fails `test:unit` when a tool is added, renamed or retired without regenerating.
- `--experiment <id>` reads a full-tier experiment's `TOOL` observations to fill the "exercised" column; stdout unless `--out`, and `--check` refuses to combine with it.

## Proof

```
Tools:           24 covered, 1 uncovered, 4 excluded (widgets) — 25 in scope
Argument groups: 13 covered, 82 uncovered — 95 in scope (2 dynamic Actor tools not statically measurable)
Uncovered tools: report-problem
Source: 115 selection + 60 agent cases in mcp-server-evals
```
It flags a real gap (`call-actor` 0/8: `actor, input, waitSecs, callOptions.*`) and not a covered one (`fetch-actor-details` `actor`). `--check` exits 0 on the committed file, 1 after a deliberate edit. Two regenerations byte-identical. Unit tests 104 / 1681 / 1 (base 103 / 1631 / 1), offline with every `LANGFUSE_*` / `APIFY_TOKEN` unset. `git diff -- src/` empty.

## Caveats

- `--experiment` has only run against a fixture of denial spans, a selection-run id (warning + `n/a`, as designed) and a bad id. Its first real use is the first full-tier run, which hasn't happened.
- `report-problem` shows `uncovered` (24/25) with a footnote; tracked in #1340.
- The committed matrix is a snapshot of a live dataset, so editing a case in the Langfuse UI makes `--check` stale until someone regenerates and commits.

## Follow-ups (the gap-fill waves)

Runs family and docs tools have no agent-kind coverage (a run-lifecycle chain, a docs chain). 82 of 95 argument groups uncovered, most visibly `call-actor` run options and storage pagination. Error cases via `expectedErrors`. Recapture the `--experiment` fixture from the first full-tier run; consider gating `--check` in CI.

AI assistance: Claude Code implemented this and applied fixes from an independent external review; a human reviews before merge.

Part of apify/ai-team#265 — delivers the matrix box only; the four gap-fill boxes stay open, so this does not close the issue.
Part of apify/ai-team#236

https://claude.ai/code/session_015G1k2s4ZYv1LhkvhvUNvxR


jirispilka:
Closing without merge. The stack does not depend on this PR: #1339 is rebased directly onto #1340. The one new finding here, the argument-group coverage list (13/95 covered), is posted on apify/ai-team#265 to drive the gap-fill waves. A matrix is worth revisiting only after the first real full-tier run exists, so the "exercised" column has data; if a tool is still wanted then it should be a plain read-only report with no committed snapshot and no freshness test.

PR #1344, opened 2026-09-07
feat: Track ?client= URL param in Segment telemetry

MQ37:
Part of apify/apify-mcp-server-internal#736.

Stacked on #1327, which is stacked on #1326, which is stacked on #1334 — base branch is #1327's, not master; diff here is just this PR's own commit.

## What
Adds an optional `?client=` URL query param, threaded through both protocol eras into Segment tool-call telemetry as `mcp_url_client` (also carried into the `MCP Reported Problem` event).

## Why
Distributed connector setup URLs (Cursor, ChatGPT, Claude, ...) could each carry their own `?client=` tag for Segment attribution — more reliable than the MCP handshake's self-reported `clientInfo.name`, which varies across proxies (`mcp-remote`) and client SDK versions. Follows the same per-connection URL-param pattern already used for `?ui=`/`?payment=`/`?telemetry-enabled=`.

Flagging per this repo's own rule: this touches what the hosted server consumes (`ActorsMcpServerOptions`) — `apify-mcp-server-internal` needs a matching change (its own `src/server/shared.ts` URL-param extraction) before `?client=` does anything on `mcp.apify.com`; this PR alone is inert there.

## Testing
`pnpm run type-check / lint / test:unit / format / check:agents` all green (1617 tests + 1 pre-existing unrelated skip, on top of #1327's own additions). New tests: `?client=` value reaches `mcp_url_client` on a sync tool call, and defaults to `''` when absent; `BASE_TELEMETRY_KEYS`'s exact-key-set pin updated for the new field.

### DIFF
--- a/src/dev_server.ts
+++ b/src/dev_server.ts
@@ -163,6 +163,7 @@ async function serveStatelessRequest(req: Request, res: Response, taskStore: InM
     const uiParam = urlParams.get('ui');
     const serverMode = uiParam !== null ? parseServerMode(uiParam) : parseServerMode(process.env.UI_MODE);
     const paymentProvider = await resolvePaymentProvider(urlParams.get('payment'));
+    const clientParam = urlParams.get('client') ?? undefined;
 
     // No token required in payment mode, mirroring `resolveRequestAuth`.
     const apifyToken = paymentProvider ? undefined : extractApiTokenFromRequest(req);
@@ -191,6 +192,7 @@ async function serveStatelessRequest(req: Request, res: Response, taskStore: InM
                 serverMode,
                 paymentProvider,
                 token: apifyToken,
+                clientParam,
             });
             // Client identity arrives per request in the `_meta` envelope (there is no initialize
             // handshake), so this fetch carries no request-origin tag.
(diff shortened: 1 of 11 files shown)


jirispilka:
@MQ37 this is not adhoc, issue was created: https://github.com/apify/ai-team/issues/230

jirispilka (approved):
Looks good.

I made a couple of changes and pushed 😊 

- **Renamed** `mcp_url_client` → `mcp_client_tag`, `clientParam` → `clientTag`. Old name sorted away from `mcp_client_name`/`_version` in Mixpanel and named the mechanism, not the meaning. Free to rename: production never set it, so no events carry the old name.
- **Added `parseClientTag`** (`src/utils/client_tag.ts`, modelled on `parseServerMode`). The param reached Segment raw, so `Claude Connect` / `claude_connect` / `claude-connect`  were three values and any caller could mint unbounded more. Now kebab-case, `''` when absent, `unknown` when malformed or over 32 chars. Also settles `+`: `claude+connect` decodes to a space and lands on `claude-connect`.
- **Runs in the `ActorsMcpServer` constructor**, beside `serverMode`, not in
  `prepareTelemetryData` — the ingestion boundary the class already declares. Hosted   server keeps passing the raw value.
- **Tests:** new `utils.client_tag.test.ts` owns the vocabulary; your two contract tests   became one `it.each` on wiring.

PR #1357, opened 2026-09-08
fix: Stop pairing call-actor-widget and get-actor-run-widget

MQ37:
Stacked on #1344, which is stacked on #1327, which is stacked on #1326, which is stacked on #1334 — base branch is #1344's, not master; diff here is just this PR's own commits.

## What

`WIDGET_BY_BASE_TOOL` no longer auto-pairs `call-actor`/`get-actor-run` with `call-actor-widget`/`get-actor-run-widget`. Only `search-actors`/`fetch-actor-details` keep auto-pairing.

Both dropped widgets stay fully selectable standalone via `?tools=call-actor-widget` / `?tools=get-actor-run-widget` — widget tool implementations, the shared React bundle, and the `ui://widget/actor-run.html` resource are untouched; only automatic pairing goes away.

- New `ALL_WIDGET_TOOLS` (`tools/registry.ts`) splits the map's two overloaded jobs apart: it now sources `?tools=` name classification and direct selection (all 4 widgets), while `WIDGET_BY_BASE_TOOL` (2 entries) stays sole source for auto-pairing. Trimming the shared map to 2 entries earlier had silently broken classification too — `?tools=call-actor-widget` was misclassified as an Actor ID and triggered a live 404 API lookup.
- Auto-injected run-workflow helpers (`get-dataset-items`, `get-key-value-store-record`, `abort-actor-run`) now also trigger from the widget names, not just their base tools. `get-actor-run` itself is excluded whenever its own widget sibling (`get-actor-run-widget`) is present and no non-widget trigger (`call-actor`, an Actor tool, or an explicit `get-actor-run` selection) also fired — covers both the widget alone and both widgets combined with no base tools.
- `call-actor`'s "WIDGET ALTERNATIVE" description addendum and `getServerInstructions`'s "Widget workflow" / "Data vs widget Actor tools" bullets are now gated on the exact tool name(s) they reference (base, widget, or requiring both for the still-auto-paired search/details case), with standalone wording when only the widget is present. The addendum's own `search-actors-widget` mention is gated on the widget's presence too, not just `search-actors`'.
- Adds end-to-end coverage: a real MCP client against the built server confirms standalone and combined `?tools=` selections actually serve the right widget/base tools and instruction/description text, not just a gating check that happens to render nothing.

## Why

Low usage on `call-actor-widget`/`get-actor-run-widget` justified dropping automatic pairing. Full removal (deleting the widget files, shared bundle, resource) was considered and discarded as too invasive for the actual ask.

An earlier version of this PR fixed a real `?tools=` classification regression by adding both widget names to `RETIRED_SELECTOR_NAMES`, which made them permanently unselectable as a side effect — that broke a supported case: a widget-only session (client selects only the widget, no base tool) is expected, not an edge case. This PR replaces that fix: splitting the overloaded map (`ALL_WIDGET_TOOLS` vs `WIDGET_BY_BASE_TOOL`) fixes the same classification bug without killing selectability.

## Testing

`type-check`, `lint`, `format`, `check:agents`, full `test:unit` all green (1632/1633 tests + 1 pre-existing unrelated skip). Integration cases run against the real built server with a live Apify token, covering every widget-selection shape (each widget alone, both widgets together, widget + base together) for tool loading and instruction/description text. `registration.cases.ts`'s Claude-connector fixture is single-sourced with the offline unit test that pins the same surface, instead of two hand-maintained copies.

## Follow-up (not in this PR)

A read-only scout found 6 more instances of the same "ungated cross-tool nudge" bug shape elsewhere in the codebase (pre-existing, unrelated to this change) — tracked in apify/apify-mcp-server#1358.

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -72,11 +72,19 @@ export type HelperToolName = (typeof HELPER_TOOLS)[keyof typeof HELPER_TOOLS];
 
 /**
  * Retired tool selectors: `add-actor` and `experimental` (add-actor was deleted in the stateless
- * migration) and the deprecated `preview` pseudo-category. They name neither a registry category
- * nor a real tool anymore, so they resolve to nothing — never loaded, never treated as an Actor ID,
- * never requiring a token by themselves.
+ * migration), the deprecated `preview` pseudo-category, and `call-actor-widget`/`get-actor-run-widget`
+ * (widget tools that stopped being servable when `WIDGET_BY_BASE_TOOL` dropped their pairing — without
+ * this, `resolveActorsToLoad` would misclassify either name as an Actor ID and fetch it as one).
+ * None of these name a registry category or a real tool anymore, so they resolve to nothing — never
+ * loaded, never treated as an Actor ID, never requiring a token by themselves.
  */
-export const RETIRED_SELECTOR_NAMES: ReadonlySet<string> = new Set(['add-actor', 'experimental', 'preview']);
+export const RETIRED_SELECTOR_NAMES: ReadonlySet<string> = new Set([
+    'add-actor',
+    'experimental',
+    'preview',
+    'call-actor-widget',
+    'get-actor-run-widget',
+]);
 
 /**
  * Client-name substrings (lowercased, matched against `clientInfo.name`) that `report-problem` is
--- a/src/const.ts
+++ b/src/const.ts
@@ -72,19 +72,11 @@ export type HelperToolName = (typeof HELPER_TOOLS)[keyof typeof HELPER_TOOLS];
 
 /**
  * Retired tool selectors: `add-actor` and `experimental` (add-actor was deleted in the stateless
- * migration), the deprecated `preview` pseudo-category, and `call-actor-widget`/`get-actor-run-widget`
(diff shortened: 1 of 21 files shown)


MQ37:
Known wart: \`CLAUDE_CONNECTOR_TOOLS\` is duplicated (\`registration.cases.ts\` + \`tests/unit/helpers/claude_connector_tools.ts\`). Don't like it — it's punishment for a not-optimal architectural choice: \`tests/test_kit\` is its own \`tsc -b\` project (can't import outside its root) and also ships as the published \`./test-kit\` export, while \`tests/unit\` runs with no build step at all. Every "just import the other one" attempt broke one of those constraints.

There is a real fix: put the fixture inside \`test_kit\` using its normal package self-reference, and add a vitest resolver alias mirroring \`tsconfig.json\`'s own path alias so unit tests can load it without building first. Verified working, but it's test-infra work touching all unit + integration module resolution — out of scope for a widget-unpairing PR. Filed as #1359.

jirispilka (approved):
I've chedked it and polished it a bit.

Tested with mcpc end to end without no issues

PR #1363, opened 2026-09-10
feat(actors): surface run tip from key-value store

### LINKED ISSUE #1361: Surface RAG Web Browser TIP field to MCP client
RAG Web Browser writes actor tips into a TIP key in its KVS. Currently only advice about URLs to fetch (instagram.com, facebook.com, etc). More advice types planned later.

Surface this TIP content to the MCP client so it can act on it (e.g. as tool-result metadata or a warning), instead of leaving it unread in the KVS.

MQ37:
## What
Surface an Actor's advisory `TIP` key-value-store record (e.g. rag-web-browser's specialized-Actor recommendation) in `call-actor` / `get-actor-run` responses: `structuredContent.tip` (`{message, level}`) plus a `Tip: <message>` line appended to the text summary.

## Why
Closes #1361 — the TIP record was written by the Actor but never surfaced to the MCP client. Fetched only when the key is actually listed in the run's default key-value store, so no extra round trip on the common case (no tip). Record shape isn't trusted since any Actor can write the reserved key — a malformed record is discarded, not surfaced.

## Testing
`pnpm run type-check`, `lint`, `test:unit` (1678 passed), `format`, `check:agents` all clean. Added unit tests: tip surfaced in structuredContent + text when present, no fetch and no field when the key isn't listed, malformed record discarded without throwing.

E2E against the live `apify/rag-web-browser` Actor via `mcpc` (this branch's built `dist/stdio.js`), both call sites that share `fetchActorRunData`/`buildGetActorRunResponse`:

**`call-actor`** (waitSecs=40, single call to terminal, run `CneLWC7Nmmk4rA5MF`):
```
SUCCEEDED in 16.528s. 1 item; 31 fields available. Key-value store has 4 keys.
Use get-dataset-items with datasetId=PIauqKzQ3xtLuVDKe and limit (for example 20) to fetch items (1 total). Available fields (dot notation): query, markdown, ... — pass via fields="..." to project.
Tip: For scraping booking.com, we recommend using [Booking Scraper](https://apify.com/voyager/booking-scraper)
```
`structuredContent.tip = {"message": "For scraping booking.com, we recommend using [Booking Scraper](https://apify.com/voyager/booking-scraper)", "level": "info"}`

**`get-actor-run`** (started via call-actor waitSecs=0, then polled, run `HBaQEFBBW1AormCR1`):
```
SUCCEEDED in 12.685s. 1 item; 13 fields available. Key-value store has 4 keys.
Use get-dataset-items with datasetId=RDz6SySuSyuk8MNJd and limit (for example 20) to fetch items (1 total). Full output is ~899 bytes. Available fields (dot notation): text, crawl.httpStatusCode, ... — pass via fields="..." to project.
Tip: For scraping zillow.com, we recommend using [Zillow Detail Scraper](https://apify.com/maxcopell/zillow-detail-scraper)
```

Both confirm: the key-value store's `keys` listed `TIP` alongside `INPUT`/SDK state keys, the record was fetched once, and `Tip: ...` landed at the end of the text after the console-link suffix.

### DIFF
--- a/src/tools/actors/actor_run_response.ts
+++ b/src/tools/actors/actor_run_response.ts
@@ -21,6 +21,9 @@ import { DEFAULT_DATASET_ITEMS_LIMIT } from '../storage/get_dataset_items.js';
 /** Cap on `storages.keyValueStores.default.keys` array length. */
 const KV_KEYS_LIMIT = 50;
 
+/** Reserved key-value store key some Actors use to advertise advisory guidance about the run. */
+const TIP_KVS_KEY = 'TIP';
+
 /** nextStep text for widget-rendered responses: suppresses LLM polling. */
 export const WIDGET_NO_POLL_NEXT_STEP =
     'Widget is rendering live progress. Do NOT poll — the widget self-updates until completion.';
@@ -144,6 +147,8 @@ export type RunResponse = {
         memMaxBytes?: number;
     };
     storages: RunStorages;
+    /** Advisory guidance an Actor wrote under the reserved {@link TIP_KVS_KEY}, if any. */
+    tip?: { message: string; level: 'info' | 'warning' };
     summary: string;
     nextStep: string;
 };
@@ -307,6 +312,29 @@ async function fetchKvKeys(
             errMessage: errMessage(error),
         });
         return null;
+    }
+}
+
+/** Defensive parse of a TIP record's value: any Actor can write it, so shape isn't trusted. */
+function parseActorTip(value: unknown): RunResponse['tip'] {
+    if (!value || typeof value !== 'object') return undefined;
+    const { message, level } = value as { message?: unknown; level?: unknown };
+    if (typeof message !== 'string') return undefined;
+    return { message, level: level === 'warning' ? 'warning' : 'info' };
+}
+
+/** Fetch the run's advisory tip; a transient failure or malformed record logs and yields undefined. */
+async function fetchActorTip(
+    client: ApifyClient,
+    keyValueStoreId: string,
+    mcpSessionId?: string,
+): Promise<RunResponse['tip']> {
+    try {
+        const record = await client.keyValueStore(keyValueStoreId).getRecord(TIP_KVS_KEY);
+        return parseActorTip(record?.value);
+    } catch (error) {
+        log.warning('Failed to fetch Actor tip', { keyValueStoreId, mcpSessionId, errMessage: errMessage(error) });
+        return undefined;
     }
 }
 
@@ -894,6 +922,11 @@ export async function fetchActorRunData(params: {
         keyValueStore: keyValueStores?.default,
     });
 
+    // Only fetch the record when the key was actually listed — avoids a round trip on every run.
+    const tip = keyValueStores?.default?.keys?.includes(TIP_KVS_KEY)
+        ? await fetchActorTip(client, keyValueStores.default.id, mcpSessionId)
+        : undefined;
+
     const structuredContent: RunResponse = {
         runId: run.id,
         actorId: run.actId,
@@ -908,6 +941,7 @@ export async function fetchActorRunData(params: {
             ...(datasets && { datasets }),
             ...(keyValueStores && { keyValueStores }),
         },
+        tip,
         summary,
         nextStep,
     };
--- a/src/tools/runs/get_actor_run.ts
+++ b/src/tools/runs/get_actor_run.ts
@@ -70,10 +70,11 @@ export function buildGetActorRunResponse(
 
     // Mints the `apifyConsoleUrl` fields onto structuredContent and returns the narrative suffix in one pass.
     const consoleLinks = applyConsoleLinks(structuredContent, linkContext);
+    const tipText = structuredContent.tip ? `\nTip: ${structuredContent.tip.message}` : '';
     return respondOk(
         [
             JSON.stringify(structuredContent),
-            `${structuredContent.summary}\n${structuredContent.nextStep}${consoleLinks}`,
+            `${structuredContent.summary}\n${structuredContent.nextStep}${consoleLinks}${tipText}`,
         ],
         { structuredContent, meta: buildUsageMeta(run) },
     );
--- a/src/tools/structured_output_schemas.ts
+++ b/src/tools/structured_output_schemas.ts
@@ -508,6 +508,14 @@ export const actorRunOutputSchema = {
                 },
             },
         },
+        tip: {
+            type: 'object' as const,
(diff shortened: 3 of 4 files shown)


jirispilka (commented):
I'm afraid I don't like it. This is really a hack, adding hundred lines for a single Actor. No one else is using it.

@nicklamonov A better solution would be to add this tip to the Actor's `status_message`. This would be sent to LLM in the response. Of course it would work on for a single URL, or say 3 URLs? The clear benefit is that it will work not only in MCP but also in API and CLI!

This would be good enough solution for all the interfaces IMO.

nicklamonov:
Thanks for the feedback @jirispilka !

My initial idea was to actually expose this kind of tip via API, so agents can **consume this info together with results**. 
When trying to find out which parts of the platform would need to be changed it appeared that many places would need to be affected.

One of the key problems: agents mainly fetch run results and that's it. (after all, this is all they need) So if we put this info in other objects (like run details), there is much less chance that agents fill find it. If we attach this tip to outputs, we would need to update the object structure of dataset items endpoint and introduce something else in addition to the list of items. Both don't sound like a good enough option.

**Run status** usually communicates something different - success, failure, number of processed results, specific error message. So it's really a status of how the run ended up.
The **tip** I'd like to expose communicates a recommendation how to optimize actor input to get better/different results - use more suitable actor, update inputs to avoid conflicts or mismatches, try an adjacent actor to enrich results. Pushing this info to status doesn't seem to be proper. It should complement the status, not replace it.

KVS is probably not the perfect place for it either, but for a pilot it could possibly be ok. The idea is - our MCP will know that it exists and can be a tool to communicate this information to the agent (or even human) together with providing results of the run. Once we see it working for RAG WB, we can communicate it to other Store teams to implement it in a similar way. 
(They are actually waiting for some solution to expose recommendations to their users, so we're trying to find a way to do that for them.)

If you have any suggestions where such a tip could fit better, I'm happy to discuss! ;)

RobertCrupa:
I like @jirispilka 's idea - as the implementation will change I will hold off on reviewing

MQ37:
guys, let's either ship it or kill it - this is not how you build :unicorn: 

cc @jirispilka @nicklamonov

jirispilka (changes requested):
On process: @MQ37 fair, I was slow, sorry for that. Let me make my remaining concerns and merge criteria explicit. The PR shows that the tip reaches the client; we haven't yet shown that it improves agent.

You both said "this is a pilot", "let's try and see" (discussing with MQ in person). 

If this is a pilot, let's run it as one. Two things are missing:

1. A success metric, agreed before merge. What number, over what window, says the tip works, and what number kills it? "Once we see it working for RAG WB" is not measurable. The PR already shows it working in the technical sense, and that was never the question.

2. Evals. This PR includes no behavioral eval evidence. Let's compare it with and without the tip, checking whether it leads to appropriate follow-up.

I'm fine merging once 1 and 2 exist and the code issues below are addressed. Without them, I don't think we have a sufficient data for evaluating the pilot or expanding it.

Code, if it goes ahead:

- `parseActorTip` doesn't cap `message`, and the message is repeated in `content[0]`, `content[1]`. Please bound it and test oversized records. We already bound other response content.
- The `keys.includes('TIP')` guard misses a real tip when it falls outside the first 50 keys. `getRecord` returns `undefined` for a missing key. Fetching it in the existing `Promise.all`, when the run is terminal and the default store exists, would fix this and remove the sequential dependency. The tradeoff is an extra request for runs without a tip.
- `Tip: <message>` presents Actor-authored advice without attribution. Use `Tip from Actor: "..."` so its source is clear. This is an attribution improvement, not a complete prompt-injection defense. The similar `statusMessage` concern can stay separate.
- Unknown `level` values, including `WARNING`, silently become `info`. I'd prefer omitting an unrecognized level rather than assigning a severity the Actor didn't provide. The output schema allows that; the TypeScript type would need updating. Please add coverage for missing and unknown levels.
- `src/tools/AGENTS.md` and the test file docstring say run responses never inline KV bodies. This introduces an exception; update both to describe it.

nicklamonov:
Thanks a lot for getting back to it!

Regarding success metric we may need to design something together. But I see it in the following way:

**Current flow** (before the change):
1. Prompt to crawl data from one of the configured platforms (insta, tiktok, facebook, and so on) -> 
2. MCP calls RAG WB with a link or query and delivers result to the user -> 
3. User/agent makes next prompt to run another actor, which is more fitting the source. (here we need to measure also if that would be the same actor we would have recommended or not)

We need to measure this behavior as baseline. If this already happens with some percentage of prompts, we would compare the follow-up with this level.

Can we collect this stats already now or is there anything missing?

**After the change**:
1. Prompt to crawl data from one of the configured platforms (insta, tiktok, facebook, and so on) -> 
2. MCP calls RAG WB with a link or query and delivers result to the user **along with a tip to use another specific actor** -> 
3. User/agent makes next prompt or call to run **that specific recommended actor**.

The metric can be called something like "**Follow-up rate based on recommendation**".
- If the rate is more than 50%, it's definitely a success.
- 30-50% - questionable.
- less than 30% (or maybe 20%) - likely not sufficient.


Does this sound reasonable enough or do we need to dig deeper?

jirispilka:
Good framing, thresholds make sense. Pulled a rough baseline already — tracked in apify-ai-team#328 (internal) along with the Redash queries. Short version: ~8.6% of rag-web-browser sessions already follow up with a different Actor, but that's an upper bound, not "the recommended one" — we don't have that signal yet. Getting the real number needs a `recommendedActorId` field on the TIP record instead of just free text; small change on our side once that's there.

@MQ37 @nicklamonov let's continue discussion here as there are some implications for this PR and RAG Web Browser: https://github.com/apify/ai-team/issues/328

MQ37:
Pushed the recommendedActorId plumbing this thread/ai-team#328 was waiting on:

- `RunResponse['tip']` gains `recommendedActorId?: string`, parsed defensively same as `message`/`level`.
- Output schema documents it as usable directly as `call-actor`/`fetch-actor-details`'s `actor` arg (verified live: `apify/rag-web-browser`'s TIP record on a real run resolves `recommendedActorId` to the real Actor ID, e.g. `shu8hvrXbJbY3Eb9W` → `apify/instagram-scraper`).
- `deriveResourceIds()` now emits `tip_recommended_actor_id` for the MCP Tool Call telemetry event, so the follow-up-rate metric from ai-team#328 is computable.

Confirmed production `apify/rag-web-browser` (default build, today) already writes `recommendedActorId` — this isn't waiting on a further Actor-side deploy.

type-check/lint/test:unit/format/check:agents all clean.

PR #1368, opened 2026-09-11
fix: Let explicit report-problem selection bypass the client blocklist

MQ37:
## What

`report-problem` is hidden from Anthropic-named clients (`claude`, `anthropic`, `local-agent-mode-apify` substring match on `clientInfo.name`) regardless of whether the tool was explicitly requested via `?tools=report-problem` (or `?tools=dev`, its sole category). A new `isReportProblemExplicitlySelected` check bypasses the client-name blocklist specifically when the tool was explicitly named — telemetry-off and unknown-client-identity remain unconditional guards either way. Session recovery (`loadToolsByName`) counts as explicit too, since it can only restore tools the session was already legitimately serving.

## Why

A Claude connector configuration that deliberately lists `report-problem` in its pinned `?tools=` URL still silently lost the tool for any real Claude client, since the blocklist applied unconditionally. `?client=claude+connector` (a Segment attribution tag) was never the gate — the real gate is `clientInfo.name` from the MCP handshake, which the URL tag doesn't affect.

## Testing

`type-check`, `lint`, `format`, `check:agents`, full `test:unit` all green (1682/1683, 1 pre-existing unrelated skip). Existing gating tests that seeded `report-problem` via `loadToolsByName` now correctly expect the bypass (recovery = explicit); one test re-seeded through the default-injection path so the blocklist itself stays covered. New unit coverage for `isReportProblemExplicitlySelected` (string/array/`dev`/comma-joined-string/absent). The Claude-connector integration case now sets a real `clientName: 'claude-ai'` (previously it used a generic client name and never actually exercised the blocklist) — confirmed this gated path specifically: ran it against the live built server with the fix reverted (fails, tool missing) and restored (passes).

### DIFF
--- a/tests/unit/mcp.server.report_problem_gating.test.ts
+++ b/tests/unit/mcp.server.report_problem_gating.test.ts
@@ -6,15 +6,14 @@ import { HELPER_TOOLS } from '../../src/const.js';
 import { ActorsMcpServer } from '../../src/mcp/server.js';
 import { SERVER_MODE } from '../../src/types.js';
 import type * as ToolsLoaderModule from '../../src/utils/tools_loader.js';
-import { getActors } from '../../src/utils/tools_loader.js';
 import { getLegacyServer } from './helpers/mcp_server.js';
 
 // Stub getActors so default-injection seeding needs no network.
+// Default-resolves to [] so tests that never call loadReportProblemByDefault still get a valid array.
 vi.mock('../../src/utils/tools_loader.js', async (importOriginal) => {
     const actual = await importOriginal<typeof ToolsLoaderModule>();
-    return { ...actual, getActors: vi.fn() };
+    return { ...actual, getActors: vi.fn().mockResolvedValue([]) };
 });
-const getActorsMock = vi.mocked(getActors);
 
 type InitHandler = (req: InitializeRequest, ctx: unknown) => Promise<unknown>;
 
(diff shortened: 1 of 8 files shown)


tests/unit/mcp.server.report_problem_gating.test.ts
```diff
@@ -1,12 +1,21 @@
...
+import { getActors } from '../../src/utils/tools_loader.js';
 import { getLegacyServer } from './helpers/mcp_server.js';
 
+// Stub getActors so default-injection seeding needs no network.
+vi.mock('../../src/utils/tools_loader.js', async (importOriginal) => {
+    const actual = await importOriginal<typeof ToolsLoaderModule>();
+    return { ...actual, getActors: vi.fn() };
+});
+const getActorsMock = vi.mocked(getActors);
+
```
RobertCrupa, line 17: The new vi.mock of tools_loader.js stubs getActors as a bare vi.fn() with no default resolved value. Only the first test (loadReportProblemByDefault) calls mockResolvedValue([]); every later test relies on that leaked implementation.

Verified - running any other test alone fails:

```
pnpm exec vitest run tests/unit/mcp.server.report_problem_gating.test.ts \
  -t "serves report-problem to a non-Anthropic client when composed before initialize"

TypeError: Cannot read properties of undefined (reading 'length')
 ❯ getToolsForServerMode src/utils/tools_loader.ts:154
 ❯ ActorsMcpServer.composeToolsForClient src/mcp/server.ts:358
```

7 of 8 tests break under -t, .only, or any reordering. Fix: getActors: vi.fn().mockResolvedValue([]) in the mock factory, or seed it in a beforeEach.

RobertCrupa (approved):
Looking good - it feels strange to unblock this tool on the connector, but not on claude code 🤔

PR #1372, opened 2026-09-13
feat: Add schedule tools for Actors and tasks

### LINKED ISSUE #1341: [spec] Add schedule functionality to MCP server
### Problem

### Proposed solution

### Plan

### Alternatives considered

_No response_

Janjiran:
Adds a schedules tool category (not enabled by default) with four tools:
- `create-schedule`
- `get-schedule`
- `update-schedule`
- `delete-schedule`

An agent can now run an Actor or a saved task on a cron cadence.

Notes for reviewers:

- Actions share one flat shape on input and output ({ actorId | taskId, input, build, timeoutSecs, memoryMbytes, restartOnError }), so a get-schedule result pastes straight into update-schedule. The API replaces all actions on update, so this is what makes "add one action" work.
- Actor and task names are resolved to IDs first. The schedules API accepts IDs only in actions.
- `create-schedule` defaults `isEnabled` to true. I guess this is up to discussion but since scheduled jobs are one of the most "user keeping" tool, it makes sense to me to enable them by default - there could be downsides I guess tho, so let me know @jbartadev 
- @jirispilka @Valentin-B117 Not sure about the evals right now - should I keep or drop them? 

Tested: unit suite (1735 tests), plus a live mcpc run against a real account covering create, get by name/ID, pause, action round-trip, error cases, and delete.

Follow-up: configurator copy in apify-mcp-server-internal (branch feature/schedules-tool-copy)

part of #1341


jirispilka:
@Janjiran thanks. Robert will look into this and will fix the evals in the langfuse.

@RobertCrupa I'm at the conference today, would you please review this PR.

.mcp.json
RobertCrupa, line 1: Nit - not sure if this needed to be changed

src/tools/schedules/schedule_helpers.ts
```diff
@@ -0,0 +1,264 @@
...
+ * same cached, ownership-gated lookup as call-actor, so an Actor the model just inspected costs no
+ * extra request; an Actor without a build resolves to nothing, and could not be scheduled anyway.
+ */
+async function resolveActorId(client: ApifyClient, idOrName: string): Promise<string | undefined> {
+    const actor = await getResourceByIdOrName(
+        idOrName,
+        async (safeId) => (await getActorDefinitionCached(safeId, client)) ?? undefined,
+    );
+    return actor?.info.id;
+}
```
RobertCrupa, line 137: cross-tenant cache key

resolveActorId keys the process-wide actorDefinitionCache by toSafeResourceId()'s output, so a bare Actor name becomes ~name. That key is caller-relative but the cache is not, and callerMaySeeCachedActor returns true for anything isPublic. 

Reproduced in-process with the real cache and the real gate (the schedule tests vi.mock getActorDefinitionCached, which is why they can't see it). Tenant A schedules its public foo; tenant B then schedules foo and gets ACTOR000000000001 — tenant A's Actor — after zero API calls.

Two variants, and the second is the worse one:
- B owns its own foo → silently schedules A's Actor instead of B's.
- B owns nothing named foo → the create succeeds against A's Actor where the correct answer is "not found".

The gate itself is fine: a cached private Actor is correctly withheld and re-fetched, and task actions are unaffected (no cache on that path). The defect is scoped to public Actors reached by bare name. Keying the cache by the resolved ID or full name, rather than the caller-relative ~name, would close it.

Not reproduced live — that needs two tokens from different accounts.

src/tools/schedules/schedule_helpers.ts
```diff
@@ -0,0 +1,264 @@
...
+
+/** Fetches a schedule by ID, name, or `username/name`. */
+export async function getScheduleByIdOrName(client: ApifyClient, idOrName: string): Promise<Schedule | undefined> {
+    return getResourceByIdOrName(idOrName, async (safeId) => client.schedule(safeId).get());
+}
+
+/**
+ * Resolves the Actor an action runs to its ID, which is all the schedule API accepts. Goes through the
+ * same cached, ownership-gated lookup as call-actor, so an Actor the model just inspected costs no
+ * extra request; an Actor without a build resolves to nothing, and could not be scheduled anyway.
```
RobertCrupa, line 129: The docstring claim is false. "An Actor without a build … could not be scheduled anyway" —the API schedules it fine. Verified against a never-built Actor: schedules().create() returned 200.

What actually breaks is a never-built Actor, and not via null. defaultBuild() throws 403 unknown-build-tag, and actor_definition.ts:56 only maps 404/400 to null — so it rethrows and the user gets:

```
Error calling tool "create-schedule": Build with tag "latest" was not found.
Has the Actor been built already? (API error type: unknown-build-tag).
The resource may be private or your token may lack access.
```

client.actor(id).get() resolves the ID without touching a build and avoids both.

src/tools/schedules/schedule_helpers.ts
```diff
@@ -0,0 +1,264 @@
...
+
+/** The DNS-safe rule the API applies to schedule names, same as task names; the message names the right resource. */
+const scheduleNameSchema = z
+    .string()
+    .min(3)
+    .max(63)
+    .regex(
+        TASK_NAME_REGEX,
+        'Schedule name may contain only letters, digits and dashes, and cannot start or end with a dash',
+    );
```
RobertCrupa, line 42: Minor - The name regex never runs

`ajv.ts:10` removes the pattern keyword as a ReDoS mitigation and its comment says "No src/ schema uses them." scheduleNameSchema adds one. Length is enforced, the pattern is dropped:
```
 [AJV name="-bad-"]       true      [AJV name="ab"]      false
 [AJV name="a b"]         true      [AJV name=<64 chars> false
 [AJV name="UPPER_score"] true
```
The bad name then reaches createScheduleArgs.parse() inside call, which throws instead of returning respondUserError, so the client gets the ZodError serialised — "origin", "code", the raw pattern, and the message buried inside the JSON.

It also buys nothing. The API rejects the same name on its own with a better message: "name can only contain letters 'a' through 'z', the digits '0' through '9', and the hyphen ('-') but only in the middle of the string (e.g. 'my-value-1')".

Note taskNameSchema has the same shape, so the ajv.ts comment was already stale before this branch

RobertCrupa (approved):
Mostly LGTM - I have raised a few issues. I have verified them all apart from the cross-tenant cache issue. For this one I needed two API tokens.

Feel free to merge after verifing/addressing the issues without a re-review

evals/test_cases.json
jirispilka, line 1: This is not used anymore, let's use evals with langfuse

src/tools/schedules/create_schedule.ts
```diff
@@ -0,0 +1,115 @@
...
+        .max(20)
+        .describe('What runs at each firing: one entry per Actor or task, up to 10 Actors and 10 tasks.'),
+});
+
+function buildDescription({ hasTool }: ToolDescriptionContext): string {
+    return `Create a schedule that runs Actors and saved Actor tasks automatically on a cron cadence. Each action
+runs one Actor (with its own input and run options) or one task (its stored input, optionally with top-level
+overrides); Actor and task names are resolved to IDs before the schedule is saved.
+The schedule is created enabled unless \`isEnabled: false\` is passed, and fires in \`timezone\` (default UTC).
+Up to 10 Actor actions and 10 task actions; while a previous run from an action is still running, the next one
```
Valentin-B117, line 46: Suggestion: Is there any specific count per actor or task action ? It seems the limit to be 20 actions in total. Perhaps the description should be updated or a check added ?

Valentin-B117 (approved):
LGTM, just one small suggestion

src/tools/schedules/schedule_helpers.ts
Janjiran: Good catch! I've also added a test so it doesn't happen in the future

PR #1376, opened 2026-09-14
chore(evals): Delete the Phoenix eval runner

jirispilka:
## Why

Closes apify/ai-team#262. The Langfuse two-tier gates replaced this runner — its custom loop can't represent built-in tool competition or Claude Code's prompting, so its verdicts didn't transfer. It hasn't run in CI since the tiers landed.

## What changed

Deletes the runner, its test cases and docs. The tool-description guidance from `evals/README.md` moves into `evals/mcp_agent/README.md`, rewritten against the two-tier setup.

Three things the diff doesn't explain:

- **`evals/shared/` stays**, against the issue's text. The Langfuse runner and two unit tests import six live symbols from it, so deleting it breaks `pnpm run evals:mcp-agent`. Only the Phoenix-only parts go: `openai_tools.ts`, `loadTestCases()`, and the `ToolSelectionTestCase` / `TestData` / `McpTool` types. (#1383 has since added `schedules_fixtures.ts`, which imports `findMissingEnvVars` from it.)
- **`OPENROUTER_BASE_URL` is dropped as a knob**, not just as a repo secret. Nothing set it, so the judge endpoint is now a constant — a local `.env` pointing evals at a gateway loses the override silently.
- **The lockfile gains packages in a deletion PR.** On `master`, `@arizeai/phoenix-client` declared a peer `@anthropic-ai/sdk: ^0.35.0` that conflicted with `@anthropic-ai/claude-agent-sdk`'s required `>=0.93.0`. pnpm resolved neither, so the agent SDK ran with an unmet required peer; removing Phoenix lets pnpm install it properly. A clean `master` worktree re-resolves with no diff, so this comes from the removal rather than drift.

`.env.example` still omits `ANTHROPIC_API_KEY`, which the runner requires unless `--subscription` is passed.

## Proof it works

`grep -ri phoenix` is clean apart from the `CHANGELOG.md` release entry. The surviving runner starts and its env sanitizer now lists only the five remaining variables. `type-check`, `lint`, `test:unit` (1764 passed), `format` and `check:agents` are green after merging `master`, and `_evaluations.yaml`'s secrets contract still matches across all three of its callers.

The one merge conflict was in `_evaluations.yaml`, where #1383 added a schedule-fixture teardown step to the same block this PR removes `OPENROUTER_BASE_URL` from. Both are kept: against `master`, the only change in that file is the dropped secret.

No `mcpc` transcript: this deletes an offline eval runner and touches no server behavior. Nothing for `apify-mcp-server-internal` either — it has no references to Phoenix, `arizeai`, `evals/` or `OPENROUTER`, and `evals/` was never in the published package's `files`.

AI disclosure: written with Claude Code, which made the deletion and ran the verification above.

https://claude.ai/code/session_016rN9J5QzfFTm41xYFtRGC2


.env.example
RobertCrupa, line 1: Nit - Could we add the anthropic api key here as an optional variable?

RobertCrupa (approved):
A big clean up always puts a smile on my face :D

.env.example
jirispilka: Good point, added

Issue #1351, opened 2026-09-08
Actor development loop over MCP: deploy, build, and debug tools


DaveHanns:
## Goal

A coding agent can take an Actor through the whole development loop over the Apify MCP server alone: push source to the platform, build it, run it, read build and run logs, fix, and push again, without opening Apify Console or falling back to the Apify CLI. Part of the Creators group OKR "Make natural language the first-class interface to Apify" (Q3 2026).

## Today

The server can run and inspect Actors (`call-actor`, `get-actor-run`, `get-actor-log`, dataset and key-value store tools) but cannot deploy: there is no tool to push source, start a build, or read a build's status or log. An agent that writes an Actor in its sandbox has to shell out to the CLI to ship it.

## Scope

A new opt-in `deploy` tool category that mirrors the run tools, plus one filter on an existing tool. Sub-issues:

- [ ] #1352: `get-actor-build` and `get-actor-build-log`, mirroring `get-actor-run` and `get-actor-log`
- [ ] #1353: `build-actor`, start a build with a bounded wait
- [ ] #1217: `push-actor`, create the Actor if missing, upsert a version's source files, optional build
- [ ] #1354: `actorId` filter on `get-actor-run-list` (split from #1053)

Draft PRs exist for all four; each sub-issue links its PR.

## Design decisions

- Build tools mirror run tools one to one: a status tool without log, a separate log tool with the same `lines` semantics (0 returns the whole log), the same `waitSecs` parameter.
- `push-actor` uses the same apify-client calls as `apify push` (JSON `sourceFiles` on the version), not the documented tarball endpoint, which returns 4xx (apify/apify-core#29044).
- The `deploy` category is opt-in and self-contained: a session with `?tools=deploy` can build and debug builds without the `runs` category.

## Related

- #1260: a default connection cannot read run logs, so the full loop needs `?tools=actors,docs,runs,storage,deploy` until that lands.
- Follow-up, separate issue: rename `get-actor-log` to `get-actor-run-log` so the run and build log tools read as a pair (breaking change, needs an alias period).

PR #1380, opened 2026-09-15
feat: Store push-actor sources over 3 MiB as a zip like apify push

### LINKED ISSUE #1379: push-actor: store sources over 3 MiB as a zip, the way apify push does
## Problem

`push-actor` (draft PR #1333) refuses a set of files whose total exceeds the platform's limit for inline source files (3 MiB) and tells the agent to use the Apify CLI. An agent working purely over MCP has no CLI, so any Actor with a larger source (vendored assets, fixtures, binaries) cannot be deployed from the conversation at all.

## Proposal

Do what `apify push` does above the limit: zip the files, upload the zip to the Actor's source key-value store (`actor-<actorId>-source`, record `version-<versionNumber>.zip`, the same store and record the CLI uses), and point the version at it (`sourceType: TARBALL` with the record URL, signed when the store is restricted). Sets up to the limit keep going inline as today, so Apify Console still shows their files.

Consequences to make clear to the agent in the response: a zipped version shows no editable files in Apify Console, and `mode: merge` cannot add to it, so later pushes to that version must send all files with `mode: replace`.

Part of #1351.

Draft PR: #1380

DaveHanns:
## What
`push-actor` no longer refuses a set of files over the platform's 3 MiB inline limit. Like `apify push`, it zips the files, uploads the zip to the Actor's source key-value store, and points the version at it.

## Why
An agent working over MCP has no CLI to fall back to, so larger Actors (vendored assets, fixtures, binaries) could not be deployed from the conversation. Closes #1379. Part of Epic #1351.

## How
- Same store and record as the CLI: `actor-<actorId>-source`, key `version-<versionNumber>.zip`, `application/zip`, so CLI and MCP pushes of one Actor share them.
- The version gets `sourceType: TARBALL` and the record URL built from the client's API base URL, signed with the store's key when the store is restricted (the build worker downloads without a token). The CLI's `disableRedirect` parameter is left out: the API no longer reads it.
- The size is measured the way the platform measures it (its byte count of the content as sent, so base64 counts encoded text and an astral character counts 5 bytes), and a set of exactly the limit stays inline, which the platform accepts.
- A new Actor is created first with an empty version, the placeholder `apify push` creates too, because the store is named after the Actor ID. The version is then switched to the zip. If that upload fails, the tool answers with a soft failure that names the Actor ID and says a second push fills the version; a 403 from the store keeps its auth classification.
- Zipping uses `fflate` (new dependency, MIT, no dependencies; Node has no zip writer) at compression level 6, the level the CLI uses.
- The response gains `sourceType` and, for a zipped version, a note that Apify Console shows no files for it and that later pushes must send all files with `mode: replace`, since `merge` has nothing to merge into.

## Testing
Unit tests cover the zip contents (text and base64), the signed URL, merged sets over the limit, new version and new Actor ordering, the upload failure response, 403s from the store, the output schema of a zipped result, and the boundary on all three measures (utf8, base64, astral characters). `type-check`, `lint`, `format`, `test:unit`, `check:agents` pass. Not run against the live platform yet.

## Notes
- Stacked on #1333 (base `feat/push-actor-source`). `source_archive.ts` lives in `src/tools/versions/` next to `push_actor.ts`.
- `structuredContent` gains a required `sourceType` field, which the hosted server's consumer may need to allow.
- The tool call itself has to fit the transport: stdio has no cap, the SDK's HTTP transports cap a message at about 4 MB, and this repo's dev server uses the 100 kB default of `express.json()`. Base64 content is a third larger than the bytes it carries.
- A root file named `__proto__` is refused up front: the zip writer keys entries by path on a plain object, where that name would set the prototype.
- Out of scope: merging into a version already stored as a zip (it would need downloading and unzipping the record), clearing the stale `sourceFiles` array the platform keeps on a switched version (the CLI leaves it too), and deleting the zip record when a version is switched back to inline files.


jirispilka:
@DaveHanns sorry for the delay, will do it later today.

jirispilka (changes requested):
I found a couple of things that need fix:

1. Please correct “stdio has no cap”: the SDK defaults to a 10 MiB input buffer. The files arrive as JSON before this tool compresses them, so the archive fallback does not bypass that limit

2. The hosted server has limit 100kB, please create an issue to fix it in apify-mcp-server-internal too.

And the other one and more important:

3. This tool will be also used in the hosted server and the zip happens synchronously. I'm mostly Python guy but I believe (and Claude confirms) that it will effectively block the entire event loop for all the clients. Claude says: "flate’s async zip() offloads large-file compression but still compresses small files synchronously", which might be fine I guess. Can we run some tests to prevent any performance hit?

Let me also pull @MQ37 to the review so he is aware of this.

And a couple of issues from Claude that I did not verify yet, sorry. Posting it here anyway 😊  now so that your Claude can fix it or dismiss it.

4. A failed push can still replace the version’s source.

 For an existing zipped version, we overwrite version-0.0.zip before updating the version. If the version update fails—for example because buildTag is invalid—the tool reports failure, but the next build uses the newly uploaded files. 

Location: src/tools/versions/push_actor.ts:441.

5. Existing folder entries can produce a broken archive.

Existing versions can contain folder entries such as { name: "src", folder: true, content: "" }. When merging, the ZIP builder treats this as a regular file named src. Extracting src/main.js then fails because src is already a file. Please handle folder entries correctly and test a merge that crosses the size limit with folders present.

Location: src/tools/versions/source_archive.ts:21. I reproduced the extraction failure locally.

Two smaller additions:

 - Missing response explanation: #1379 explicitly asks the response to explain that archived sources are no longer editable in Console. That sentence is missing.
- Incomplete __proto__ guard: Incoming files are checked, but an existing __proto__ file retained during merge bypasses the guard and crashes ZIP creation. Check the final archive contents too.

MQ37:
Pushed a small fix on top of this branch to remove the hardcoded `3 MiB` string in the output schema's `sourceType` description, deriving it from the same constant used elsewhere instead.

I don't like the current `push-actor` design. `push-actor` requires full new content for any file you touch, which is expensive and error-prone even in the case it already technically supports (rewriting a 400-line file to change one line, via `mode: merge`) — and it fully breaks, not just gets expensive, the moment a version's source is stored as a zip: `mode: merge` is rejected outright once `sourceType` is `TARBALL`, so every future push must resend the entire source, forever. Every real coding agent (Claude Code, Cursor, etc.) instead reads current content and applies small, exact edits. `push-actor`'s single full-source-push model is the pattern that approach replaced.

Proposal (naming below is a placeholder to convey the idea, not final — open to whatever names reviewers prefer):
- `list-actor-files` — list files and folders in a version's source.
- `get-actor-file` — read one file's current content.
- `write-actor-file` — create or overwrite one file.
- `edit-actor-file` — apply one or more exact search-replace edits (`edits: [{oldText, newText}]`) to one existing file in a single call; each `oldText` must match exactly once or the call fails.
- `remove-actor-file`.
- Keep `push-actor` for creating a new Actor (needs `.actor/actor.json` in its first write) and for bulk replace; keep `build-actor` to trigger a build once the source is ready.

The MCP server applies each write server-side against the file's current content; the agent never sends or receives content for files it isn't touching.

**Technical details**

No API endpoint supports single-file reads or writes — only whole-version `GET`/`PUT`. Handled entirely on the MCP side: reads fetch the whole file list and return just what was asked for; writes fetch the whole list, apply the change in-process, and `PUT` the whole list back — the same call `push-actor` already makes.

- Concurrent edits need careful handling to avoid race conditions.
- A locking mechanism should prevent overlapping edits from silently overwriting each other's changes.
- Zip files (for versions stored as an archive) must be handled carefully on the server to guard against zip bombs and similar decompression abuse.

MQ37 (commented):
⏫

l2ysho (approved):
Tests broken after @MQ37 last commit, otherwise looks good 🚀

MQ37:
@l2ysho yes and no - the failing CICD is not caused by my commit - it is bc they did a new build of web fetch with internal tool and it broke our asserts - will fix.

DaveHanns:
> @l2ysho yes and no - the failing CICD is not caused by my commit - it is bc they did a new build of web fetch with internal tool and it broke our asserts - will fix.

Don't worry about it much. I will try to figure out better way of handling the source updates as suggested by you above. Possibly will drop this PR.

MQ37:
@jirispilka ahh, it is worse than I thought. They pushed change to the web fetch Actor and now the MCP server classifies that as an MCP server Actor instead of normal Actor.. So we can either do a hacky solution that I am strongly against or completely rework the too loading logic.

PR #1383, opened 2026-09-15
chore(evals): Add schedule fixtures and wire the schedules eval family

### LINKED ISSUE #1341: [spec] Add schedule functionality to MCP server

Janjiran:
Follow-up to #1372. Eval coverage for the schedule tools.

The 20 test cases are already live in Langfuse: 10 tool-call cases in `mcp-server-evals-pr`, 10 agent cases in `mcp-server-evals-merge`.

## What this adds

- `evals/mcp_agent/schedules_fixtures.ts` — deletes leftover `eval-*` schedules, creates or resets the permanent fixture `eval-nightly-sum` (disabled, `0 3 * * *`, one action running the `eval-sum-nightly` task). Same shape as `tasks_fixtures.ts`. Run it after the task fixtures, since it needs that task.
- `evals:mcp-agent:schedules-fixtures` in `package.json`.
- A seed step in `_evaluations.yaml`, merge tier only, after the task fixtures step.
- A family paragraph in `evals/mcp_agent/README.md`.

## Calibration

Live account, `--subscription --claude-judge`.

| Tier | Model | Result |
| --- | --- | --- |
| pr, 10 cases | `claude-haiku-4-5` | 10/10 |
| merge, 10 cases | `claude-opus-5` | 10/10 |
| merge, 10 cases | `claude-haiku-4-5` | 10/10 |

Three test cases needed fixing during calibration. In each one the agent behaved sensibly and my test expected the wrong thing, so the fix was to the test, not to the schedule tools.


RobertCrupa (approved):
Looks good to me.

Only issue I can think of: is there a way to tear down these cron jobs 😄?

Janjiran:
Good catch. 
Added a teardown step after the eval run, guarded by `always()` so a failed or cancelled run still cleans up.

PR #1384, opened 2026-09-16
fix(evals): Stop the add-action eval case editing the read-only fixture

### LINKED ISSUE #1341: [spec] Add schedule functionality to MCP server

Janjiran:
Fixes a case fault introduced in #1383. Needs a quick review: the Langfuse half of this change is already live, so master is briefly inconsistent until this merges.

## What was wrong

`merge/schedules/add-action-medium-1` added an Actor action to `eval-nightly-sum`, which is the fixture the pure read cases assert on. Items run 8 at a time against one account, so whether `merge/schedules/get-easy-1` saw one action or two came down to which item finished first.

It passed locally and failed on the first master run ([run 35014680476](https://github.com/apify/apify-mcp-server/actions/runs/35014680476), 60/70, above the 0.6 gate so the run was still green):

## Verification

Run locally against a live account, concurrency 8, `--subscription --claude-judge`:

| Check | Result |
| --- | --- |
| `get-easy-1`, 3 trials | 3/3 |
| Whole schedules family, both cases in one concurrent run | 10/10 |

Only two cases now name `eval-nightly-sum` and both only read it. `collision-hard-1` tries to create a duplicate of that name, and its reference already fails any update or delete of it.


jirispilka (approved):
Cool. This is ok fix for now, not the end of the story :) . 

I believe the follow up issue #1394 provides better solution. It moves the platform-contract assertions (update-schedule replaces rather than appends, collision and not-found shapes) into the integration suite, make mutating eval cases create their own schedule in-conversation so eval-sched-target goes away, and switch created resources to per-trial names with an age-based sweep. That removes the shared mutable fixture rather than defending it, and makes parallel runs from CI, cron and laptops safe.

PR #1385, opened 2026-09-16
test: meta apify token propagation

### LINKED ISSUE #754: test: Add _meta.apifyToken propagation tests
Closes a gap from the [integration test coverage audit](../blob/claude/review-mcp-integration-tests-Jh11A/res/integration_test_coverage_audit.md).

The hosted server relies on `_meta.apifyToken` arriving in `tools/call` params (`src/mcp/server.ts:638`). Currently no test exercises this path — only the bearer header / env-var paths are covered.

> **Note**: marked **optional / lower priority** in the plan. `apify-mcp-server-internal` likely already covers `_meta.apifyToken` end-to-end. Land if we want the public repo to own this guarantee directly.

## Scope

Block H in `res/integration_test_coverage_plan.md`.

## Helper change

`tests/helpers.ts` — add `omitToken?: boolean` to `McpClientOptions`:
- `createMcpStdioClient` does not put `APIFY_TOKEN` in the spawned `env`
- `createMcpStreamableClient` / `createMcpSseClient` do not send the `Authorization: Bearer ...` header

~10 lines.

## Test cases

In `tests/integration/suite.ts`, gated `runIf(stdio || streamable-http)`:

1. `should accept apifyToken via tools/call _meta and run successfully`
   - create client with `omitToken: true`
   - send `tools/call` for `apify/python-example` with `_meta.apifyToken = process.env.APIFY_TOKEN`
   - assert run completes successfully (sum produced, dataset has the expected item)

## Files touched

- `tests/helpers.ts`
- `tests/integration/suite.ts`

## Acceptance

- [ ] Test passes on stdio and streamable-http
- [ ] Calling without `omitToken` still works (no regression in existing tests)
- [ ] `npm run type-check`, `npm run lint`, `npm run test:unit` clean

## Independence

No dependency on other coverage-gap issues. Can land in any order.

## Estimate

~1 h.

daveomri:
## Why

Closes #754.

Stateful token resolution (`meta?.apifyToken || options.token`) had no tests. The stateless path already had precedence coverage.

## What changed

An integration case that `_meta.apifyToken` overrides the connection bearer token, plus unit coverage of the stateful chain (including that an empty `_meta` token does not shadow the configured one).

Three places we diverged from #754:

- **No `omitToken` helper.** `SuiteClientOptions.token: null` already drops the bearer header and omits `APIFY_TOKEN` from the stdio env.
- **No tokenless e2e case.** HTTP rejects a tokenless connection before `_meta` is parsed; stdio exits or silently reads `~/.apify/auth.json`. Fallback is a unit test, same as the stateless side.
- **Override case gated to legacy HTTP.** The behaviour itself is stateful-path-wide; stdio resolves it identically, but the stateless era prefers the validated header. The case uses `get-key-value-store-list`, not `search-actors`; that tool soft-catches its `users/me` 401 into anonymous user info, making it an unreliable failure signal.

Several pointers in #754 are stale (`McpClientOptions`, `createMcpSseClient`, `res/integration_test_coverage_plan.md`) and were not followed.

Just adding tests.

## Proof it works

Integration against the live platform, `-t "overrides the connection token"`:

```
✓ integration tests/integration/actor.server_streamable.test.ts (128 tests | 127 skipped) 503ms
✓ overrides the connection token with _meta.apifyToken 496ms

Test Files 1 passed | 3 skipped (4)
Tests 1 passed | 385 skipped (386)
```

Co-written with Claude


tests/unit/mcp.server.token_resolution.test.ts
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * End-to-end coverage stops at the `_meta` override (`tools.cases.ts`): HTTP
+ * rejects a tokenless connection before `_meta` is parsed, and stdio refuses
+ * to start without a token.
+ */
+describe('ActorsMcpServer', () => {
+    const servers: ActorsMcpServer[] = [];
+
+    function makeServer(token?: string): ActorsMcpServer {
```
jirispilka, line 14: Suggestion: Please reuse withServer in the unit file — tests/unit/helpers/mcp_server.ts:67 already does exactly what makeServer + servers[] + afterEach rebuilds

But, this is pre-existing issue that shape is copied from mcp.server.tool_sources.test.ts:34-51 and other files in tests/unit/ do the same, so this PR is following the existing convention, not breaking it. Raised repo-wide in apify/apify-mcp-server#1402 instead.

jirispilka (approved):
Looks good to me, I left one suggestion but let's ignore it for this PR.

PR #1388, opened 2026-09-17
refactor(evals): Organize the Langfuse eval harness by responsibility

jirispilka:
## Why

Closes #1378.

## What changed

Mechanical reorganization, no behavior change. `evals/mcp_agent/` and `evals/shared/` are gone; the harness is grouped by responsibility into `runner/`, `agent/`, `judge/`, `langfuse/`, `scripts/`, with `config.ts` and `environment.ts` at the root. Types and constants moved to their owning modules, the renames in the issue are applied, and unit tests were renamed to match the resulting modules with coverage unchanged.

All moves used `git mv`, so the diff reads as renames. Two exceptions, both intentional: `evals/mcp_agent/types.ts` and `evals/shared/types.ts` were folded into existing modules, so they have no 1:1 counterpart to record.

One judgement call worth a look: `JudgeLlmClient` was `Pick<LlmClient, 'callLlm'>`, which becomes a cycle once the contract moves into `judge/client.ts`. It is now a standalone `JudgeClient` type that both clients `implements`, so drift is still caught at compile time.

`pnpm run evals:mcp-agent*` command names are unchanged.

## Notes for reviewers (human-written)

Mechanical refactor only

PR #1397, opened 2026-09-17
test(evals): Cover environment and tool-name helpers

### LINKED ISSUE #1389: refactor(evals): Track eval harness simplification
## Goal

Track focused changes that make the Langfuse eval harness leaner and easier to understand after the mechanical responsibility-layout work in #1378.

This is an umbrella issue, not the scope of one PR. Every child issue is independently reviewable and must preserve eval behavior unless it explicitly states otherwise.

## Dependency

All child work depends on #1378. Do not implement against `master` before that layout lands.

## Subissues

- [ ] #1390
- [ ] #1391
- [ ] #1392
- [ ] #1393

The documentation and test work is independent. The judge and runner refactors are independent and may proceed in parallel after #1378.

## Deliberately not planned

- Splitting `adaptSdkConversation()` into per-message handlers. It is a linear stateful fold; handler extraction would require a large shared state object and add indirection.
- Generic wrappers for two-line dry-run, pagination, environment-exit, or tracing blocks.
- Renaming test helpers solely to replace `make*`; this is mechanical churn without a readability payoff.
- Replacing `isDeepEqual` with `node:util.isDeepStrictEqual`; NaN semantics differ and require a separate behavior decision.

## Done when

- #1378 and all four subissues are closed.
- Each implementation PR passes the repository verification required for the files it changes.
- Eval datasets, scoring, models, thresholds, and CI behavior remain unchanged except where a child issue explicitly approves a behavior correction.

AI disclosure: re-scoped with Codex and reviewed by a human before implementation.

### LINKED ISSUE #1393: test(evals): Cover environment and tool-name helpers
## Problem

Small helpers that affect environment readiness and MCP tool classification lack direct unit coverage. A regression can change authentication checks or classify built-in tools as server tools without a focused failure.

This work depends on #1378.

## Scope

Add focused unit coverage for:

- `isMcpToolName()`
- `stripToolPrefix()`
- `findMissingEnvVars()`
- `sanitizeProcessEnv()`, limited to stable state changes and without brittle assertions over full log output

Restore `process.env` and mocks after every test so tests remain isolated.

## Out of scope

- Refactoring the helpers.
- Testing private formatting details of redacted log messages.
- Integration tests or live provider calls.
- Changes to eval behavior.

## Done when

- Prefix matching, prefix stripping, missing or sanitized-empty values, and in-place environment sanitization have direct tests.
- Tests do not leak environment state.
- Required repository verification passes.

Parent: #1389

AI disclosure: scoped with Codex and reviewed by a human before implementation.

jirispilka:
## Why

Closes #1393. Part of #1389.

`isMcpToolName()`, `stripToolPrefix()`, `findMissingEnvVars()`, and `sanitizeProcessEnv()` had no direct unit coverage, so a regression in tool classification or the env readiness check would only surface as an opaque eval failure.

## What changed

New `tests/unit/evals.config.test.ts` covers prefix matching and stripping, including the bare prefix, empty string, another server's prefix, and the prefix mid-string. `tests/unit/evals.environment.test.ts` gains `findMissingEnvVars()` (unset, whitespace-only, quotes-only, control-chars-only, valid, input order) and `sanitizeProcessEnv()` (in-place sanitization, unset key stays undefined, clean value unchanged, raw secret never logged, short secret fully masked). Each block snapshots and restores `process.env` and restores the `console.log` spy; no assertion pins the redacted log format.

Two source changes, both from review:
- `ENV_KEYS_TO_SANITIZE` is exported so the test snapshots exactly the keys `sanitizeProcessEnv()` rewrites. A hand-copied list would silently leak a real CI value if the source list ever gained a key.
- `redact()` now fully masks values of 12 chars or fewer instead of 6. Before, a 7-char value logged as `abc***efg`, exposing 6 of 7 chars, while the docstring claimed short values were masked "to prevent reconstruction". Now a partially shown value always hides at least 7 chars. Real keys are 40+ chars, so their log lines are unchanged. Test added red first, green after.

Rebased onto `master` after #1388, #1396 and #1399 landed (squash merges), so the diff is only this PR's commits.

## Proof it works

`pnpm run type-check`, `lint`, `test:unit` (106 files, 1784 passed, 1 skipped; 20 new tests), `format`, `check:agents` all clean.

Review: a blind staff-review pass returned one nit (two missing boundary cases), included. An integrated final review of all four #1389 PRs confirmed vitest isolates test files per worker and that no other unit test touches the same env keys; its nit about the hand-copied key list is fixed. Review comments: the control-chars case used a NUL byte that Node truncates, so it tested an empty string; switched to `'\x1f\x0b'`. The redaction threshold change above answers the second comment.

AI disclosure: implemented and reviewed with Claude Code, human-triggered from the issue.

https://claude.ai/code/session_01TWm2jdLpUq8nC4JDL3DtjD


tests/unit/evals.environment.test.ts
```diff
@@ -40,3 +45,97 @@ describe('sanitizeEnvValue()', () => {
...
+describe('findMissingEnvVars()', () => {
+    const KEYS = ['TEST_UNSET_VAR', 'TEST_WHITESPACE_VAR', 'TEST_QUOTES_VAR', 'TEST_CONTROL_VAR', 'TEST_VALID_VAR'];
+    const originalValues = new Map<string, string | undefined>();
+
+    beforeEach(() => {
+        for (const key of KEYS) originalValues.set(key, process.env[key]);
+        delete process.env.TEST_UNSET_VAR;
+        process.env.TEST_WHITESPACE_VAR = '   ';
+        process.env.TEST_QUOTES_VAR = '""';
+        process.env.TEST_CONTROL_VAR = '\x00\x1f';
```
RobertCrupa: Nit - Am i right in thinking node truncates env values at the NUL byte, so the var is actually stored as ''

if so -
Fix: drop the NUL and use a control char Node preserves, e.g. process.env.TEST_CONTROL_VAR = '\x1f\x0b'; (stores as 2 chars, sanitizes to '').

evals/environment.ts
```diff
@@ -40,8 +40,8 @@ export function findMissingEnvVars(keys: readonly string[]): string[] {
 }
 
 /**
- * Redact a value for safe logging: shows first 3 and last 3 chars, masks the rest.
- * Fully masks short values (≤ 6 chars) to prevent reconstruction from the log line.
+ * Redact a value for safe logging: shows first 4 and last 4 chars, masks the rest.
+ * Fully masks short values (≤ 8 chars) to prevent reconstruction from the log line.
  * Returns '(empty)' for empty strings, '(unset)' for undefined/null.
  */
```
RobertCrupa, line 46: pre-existing but in scope: The code masks only length <= 6 and slices 3+3

Relevant here because the PR adds it('does not leak the raw secret value in a logged line'), which uses a 24-char value and so can't catch this

RobertCrupa (approved):

tests/unit/evals.environment.test.ts
jirispilka: Right, Node stores `'\x00\x1f'` as `''` (checked: length 0), so the case was a duplicate of the unset one. Switched to `'\x1f\x0b'` in af466d7.

---

jirispilka: Cool, fixed

evals/environment.ts
jirispilka: Good point: Redact a value for safe logging. Values longer than 12 chars show their first 3 and
last 3 chars, so at least 7 chars stay hidden; shorter values are masked entirely.
Returns '(empty)' for empty strings, '(unset)' for undefined/null.

PR #1406, opened 2026-09-21
docs: Say call-actor's actor parameter takes an ID or a name

DaveHanns:
## What
The `actor` parameter of `call-actor` is described as taking a `username/name`. The tool has always accepted an Actor ID as well, and its not-found message says "Actor ID or name". The description now says both.

## Why
`build-actor` (#1332) describes its `actor` parameter as "Actor ID or username/name", and `push-actor` (#1333) is about to accept both too. The default tool should not be the odd one out. Came up in the review of #1333.

## How
One line in `src/tools/actors/call_actor.ts`. No behavior change.

## Testing
`type-check`, `lint`, `format`, `test:unit`, `check:agents` pass.


jirispilka (approved):
Thanks!

